PR #72, opened 2014-01-04
Add basic task attachment functionality

livthomas:
Add a new tab into the task detail panel which allows user to add a new attachment (with plain text content only) and list the existing ones as well as delete them.


psiroky:
@Salaboy @mswiderski is this something we should still consider? The PR is old and I am not sure about the current state of the attachments support in the console.

mswiderski:
I leave that to @Salaboy to decide :)

Salaboy:
Yes, I think that there are some more work to be done in order to make this useful. This was a very good contributions from @livthomas but we need to figure out how we can make it useful for the end users and have a discussion with the UX team before adding it.

porcelli:
@Salaboy @psiroky @mswiderski  is this still makes sense?

livthomas:
I think this can be closed now since there are many conflicting files after such long period of time. If someone decides to implement this, it will probably be much easier to start from scratch.

PR #108, opened 2014-12-17
BZ 1174894 Process instance list sorting ends with an exception

bxf12315:
I find this issue reason is SimpleTable set default datastorename use caption.
So i re-set   datastorename after listGrid.addColumns(columnMetas);
Code in here
https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/SimpleTable.java#L380

This is workaround way, best way should be remove default datastorename.
here:https://github.com/uberfire/uberfire-extensions/pull/8
But, i do not know is it reasonable .
Please check it.


Salaboy:
already merged in master

PR #117, opened 2015-02-04
BZ-1185698 The variables shown on the tables are not holonomic

bxf12315:
implement WordWrapCell for  word-wrap: break-word;white-space: normal

### DIFF
--- a/jbpm-console-ng-generic/jbpm-console-ng-generic-client/src/main/java/org/jbpm/console/ng/gc/client/cell/longContentCell.java
+++ b/jbpm-console-ng-generic/jbpm-console-ng-generic-client/src/main/java/org/jbpm/console/ng/gc/client/cell/longContentCell.java
@@ -1,0 +1,54 @@
+package org.jbpm.console.ng.gc.client.cell;
+
+import com.google.gwt.cell.client.AbstractSafeHtmlCell;
+import com.google.gwt.safehtml.shared.SafeHtml;
+import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
+import com.google.gwt.text.shared.SafeHtmlRenderer;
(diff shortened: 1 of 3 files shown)


jbpm-console-ng-generic/jbpm-console-ng-generic-client/src/main/java/org/jbpm/console/ng/gc/client/cell/longContentCell.java
```diff
@@ -0,0 +1,54 @@
+package org.jbpm.console.ng.gc.client.cell;
+
+import com.google.gwt.cell.client.AbstractSafeHtmlCell;
+import com.google.gwt.safehtml.shared.SafeHtml;
+import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
+import com.google.gwt.text.shared.SafeHtmlRenderer;
+import com.google.gwt.text.shared.SimpleSafeHtmlRenderer;
+
+public class longContentCell extends AbstractSafeHtmlCell<String>{
```
manstis: If this is eventually merged (I see you're talking to salaboy on #uberfire) then this class should probably be renamed LongContentTextCell - "L" and "Text" to mark it's usage similar to "TextCell".

bxf12315: ok, i will change it.

Salaboy:
not needed anymore, we will adopt a tooltip for all the cells instead of this custom solution for just one cell

mswiderski:
just one note - wouldn't it be difficult to read only via tooltip or for example copy value of the variable?

PR #133, opened 2015-04-01
 First version of mutiple view  mode of process definitions , include co...

bxf12315:
...mmon setting page

  and refacting for baisc and advanced mode

@psiroky hi, i fixed the indentation, please check it .thanks
@Salaboy @manstis Please review it.


jbpm-console-ng-generic/jbpm-console-ng-generic-client/src/main/java/org/jbpm/console/ng/gc/client/experimental/workbench/configuration/WorkbenchConfigurationPopup.java
```diff
@@ -0,0 +1,338 @@
...
+                        }
+                    }
+                });
+    }
+
+    private void showFieldEmptyWarning() {
+        ErrorPopup.showMessage(constants.PleaseSetAName());
+    }
+
+    class Node<K, V> {
```
manstis: There is an existing Pair class in uberfire.

manstis:
My main comment is that (I assume) WorkbenchConfigurationPopup is going to be used to select generic preferences (i.e. those for the workbench as a whole - e.g. language, and those for specific bits affecting for example jbpm-console-ng). Ideally the "preferences" bit should move to kie-wb-common and re-used from jbpm-console-ng (and kie-wb-distributions' webapps). 

We'd need some way for the distributions to declare what preferences they need exposed and then the WorkbenchConfigurationPopup show them accordingly. This makes the preferences bit more complicated - as a preference may be a simple String, number etc or a dropdown, checkbox etc.

Furtheremore we have a generic "Properties Framework" in uberfire-extensions already, which should probably be re-used to populate the WorkbenchConfigurationPopup: https://github.com/uberfire/uberfire-extensions/tree/master/uberfire-widgets/uberfire-widgets-properties-editor

jbpm-console-ng-generic/jbpm-console-ng-generic-client/src/main/java/org/jbpm/console/ng/gc/client/experimental/workbench/configuration/WorkbenchConfigurationPopup.java
bxf12315: yes, but Pair can not use in jbpm-console-ng, because of some rules in here .
if this code move to uberfire, i will change it

jbpm-console-ng-generic/jbpm-console-ng-generic-api/src/main/java/org/jbpm/console/ng/ga/model/ContextualView.java
```diff
@@ -0,0 +1,19 @@
...
+	private String modeName = BASIC_MODE;
+
+	public String getModeName() {
+		return modeName;
+	}
+
+	public void setModeName(String modeName) {
+		this.modeName = modeName;
+	}
+}
```
psiroky: It seems there are still tabs used in this file?

PR #135, opened 2015-04-08
Added logging in case (un)deployment fails

selrahal:


psiroky:
@Salaboy @mswiderski this seems like a safe changed. What do you think?

mswiderski:
fine with me, but not sure it makes sense to push it to 6.0.x, are we planning o make any further releases of 6.0.x?

psiroky:
@mswiderski good catch, I overlooked the 6.0.x. I think we should merge this into master only (at least for now).

Salaboy:
@psiroky @mswiderski should I go ahead and merge this in master?

Salaboy:
merged in master

PR #142, opened 2015-04-25
Remove Sonatype OSS repository - not needed anymore

psiroky:
@Salaboy @mswiderski I believe the Sonatype repo is not needed (anymore). I did full build, starting with empty local maven repo dir to make sure nothing is cached and the build was successful. Do you know about use case that would require this repo?

We should not use any third party repos to ensure builds are reproducible, fast, etc. See https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/README.md#requirements-for-dependencies for more details.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,16 +65,6 @@
         <updatePolicy>daily</updatePolicy>
       </snapshots>
     </repository>
-    <repository>
-      <id>sonatype</id>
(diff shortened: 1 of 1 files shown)


mswiderski:
+1, I guess it was needed for some designer deps that has already been (re)moved

Salaboy:
I will merge this in then

Salaboy:
merged in master

PR #154, opened 2015-05-21
dashbuilder integration

nmirasch:


Salaboy:
@nmirasch can we close this one? or is still containing something that we don't have in master?

PR #156, opened 2015-05-29
JBPM-4601 - REST operation to provide a process (instance) image

mrietveld:
This is the pull request for JBPM-4601/BPMSPL-74. 

The REST operations in this module can not be placed in droolsjbpm-integration because they rely on guvnor code. 

After talking to Kris about the best location, he suggested jbpm-console-ng.


jbpm-console-ng-rest/src/main/java/org/jbpm/console/ng/rest/ProcessImageResourceImpl.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+                
+        // get SVG String
+        String imageSVGString = getProcesImageSVGFromDeployment(deploymentId, procDef);
+        if( imageSVGString == null ) { 
+            return Response.status(Response.Status.PRECONDITION_FAILED).build();
+        }
+       
+        // find active nodes and modify image
+        List<NodeInstanceLog> logs = getAuditLogService().findNodeInstances(procInstId);
```
mswiderski: you could use RuntimeDataService to get information about active and completed node instance:
https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/RuntimeDataService.java#L182

https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/RuntimeDataService.java#L191

jbpm-console-ng-rest/src/main/java/org/jbpm/console/ng/rest/ProcessImageResourceImpl.java
```diff
@@ -0,0 +1,183 @@
...
+    /* KIE information and processing */
+
+    @Inject
+    private GuvnorM2Repository repository;
+    
+    @Inject
+    private RuntimeDataService dataService;
+
+    /** AuditLogService **/
+    private static final String PERSISTENCE_UNIT_NAME = "org.jbpm.domain";
```
Salaboy: Do we need to have this hard-coded in this class? If the code is in the jbpm-console-ng repository it should be consuming services from jbpm-kie-services instead of consuming directly the EMF, right?

Salaboy:
is this PR still valid? Should I merge it?

mrietveld:
@Salaboy I will make the modifications you and @mswiderski suggeted and update the PR this week. If you could merge it once I'm done, that would be great. 

(I was on PTO the last 3 weeks).

jbpm-console-ng-rest/src/main/java/org/jbpm/console/ng/rest/ProcessImageResourceImpl.java
mrietveld: Wouldn't doing 2 calls be slower than just doing 1 call, given that the bottleneck is the serialization over the line and the retrieval from the database?

mswiderski: no need to do two calls, you can use one to get both active and completed nodes: https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/RuntimeDataService.java#L201

mrietveld:
@Salaboy, @mswiderski: I've updated the PR. 

I'd like to merge it today (July 3rd). If you have any comments, would you guys be able to add them today? 

Thanks,
Marco

mswiderski:
+1, looks good to me, thanks Marco!

mrietveld:
Merged: fa5807bbaef51bd539fd9b4aa872f2466ce71852

PR #161, opened 2015-06-22
JBPM-4638: UI restructuring. DatasetIntegration process Instance List

nmirasch:


Salaboy:
merged in master, added to kie-wb as well

PR #163, opened 2015-06-25
JBPM-4638: UI restructuring. Adding autorefresh configuration button

nmirasch:


Salaboy:
merged in master with some modifications for the range of values and how the options are created

PR #164, opened 2015-06-29
BPMSPL-133: Tasklist should be refreshed automatically. Store refresh…

nmirasch:
… interval in each section.


Salaboy:
merged in master

PR #167, opened 2015-07-07
Refactor TaskCommentsView/Presenter for unit testability & add unit tests

jhrcek:
Hello. This is my personal proof of concept in greater mission of migrating some QE's slow and hard-to-maintain selenium tests to faster and possibly more reliable Mockito based unit tests in community, performed at the level of UI controller. I also did some refactorings to remove some duplication and make the code a bit more readable (see comments inline). I tested the changes thoroughly. Please check it and provide some feedback.

PS.: Sorry for changing the formatting, I'll try to avoid that in future commits.

### DIFF
--- a/jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/src/main/java/org/jbpm/console/ng/ht/client/editors/taskcomments/TaskCommentsPresenter.java
+++ b/jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/src/main/java/org/jbpm/console/ng/ht/client/editors/taskcomments/TaskCommentsPresenter.java
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.jbpm.console.ng.ht.client.editors.taskcomments;
 
 import java.util.Date;
@@ -23,10 +22,6 @@ import javax.enterprise.context.Dependent;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 
-import com.github.gwtbootstrap.client.ui.Button;
-import com.github.gwtbootstrap.client.ui.DataGrid;
-import com.github.gwtbootstrap.client.ui.SimplePager;
-import com.github.gwtbootstrap.client.ui.TextArea;
 import com.google.gwt.user.client.ui.IsWidget;
 import com.google.gwt.view.client.HasData;
 import com.google.gwt.view.client.ListDataProvider;
@@ -39,24 +34,21 @@ import org.jbpm.console.ng.ht.model.CommentSummary;
 import org.jbpm.console.ng.ht.model.events.TaskRefreshedEvent;
 import org.jbpm.console.ng.ht.model.events.TaskSelectionEvent;
 import org.jbpm.console.ng.ht.service.TaskCommentsService;
+import org.uberfire.client.mvp.UberView;
 import org.uberfire.ext.widgets.common.client.common.popups.errors.ErrorPopup;
 
 @Dependent
 public class TaskCommentsPresenter {
 
-    public interface TaskCommentsView extends IsWidget {
+    public interface TaskCommentsView extends UberView<TaskCommentsPresenter> {
 
-        void init( TaskCommentsPresenter presenter );
+        void clearCommentInput();
 
-        TextArea getNewTaskCommentTextArea();
+        void redrawDataGrid();
 
-        Button addCommentButton();
+        void adjustDisplayForListOfSize(int size);
 
-        DataGrid<CommentSummary> getDataGrid();
-
-        SimplePager getPager();
-
-        void displayNotification( String text );
+        void displayNotification(String text);
     }
 
     @Inject
@@ -67,7 +59,6 @@ public class TaskCommentsPresenter {
 
     private long currentTaskId = 0;
 
-    
     @Inject
     Caller<TaskCommentsService> taskCommentsServices;
 
@@ -75,7 +66,7 @@ public class TaskCommentsPresenter {
 
     @PostConstruct
     public void init() {
-        view.init( this );
+        view.init(this);
     }
 
     public IsWidget getView() {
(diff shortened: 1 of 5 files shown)


Salaboy:
Hi @jhrcek  
this is great work. We should definitely do what you did here in all the screens :) 
I will merge the changes shortly. 

One question that I have for you is about the test coverage, and if the test really helps. Looking at the test proposed in this commit I would like to understand your vision about which kind of issues we are going to detect early because of these tests. In the test that you provided we are only testing the Presenter which contains the logic as you mention and we are also mocking the backend service, so the only scope of the test is the presenter logic, am I right? If so, I think that it is valuable but we need to in some way make sure that the backend service calls that we are doing also has tests, so everytime that we create a presenter test using mockito we can go to jBPM Services and also add a test to the services that are being invoked from all our presenters. Does it make sense? Cheers

jhrcek:
@Salaoby you're right, this test intentionaly only focuses on presenter logic.

QE's recent goal has been to replace part of our selenium tests with more lightweight tests in community.
In particular we're trying to get rid of UI tests for which selenium testing is not the most suitable approach. Using TaskComments screen as an example: using selenium to verify that comment input field is cleared after user clicks "Add comment" button is clearly an overkill and should be tested at lower level if possible.

The migration of tests to lower level has turned out not as easy as we initially thought. For example some of us didn't really understand gwt enough to be able to write unit tests / refactor gwt code, and the code itself was not written with testability in mind in the first place. We'll strive to improve test coverage in the community in the future, but most of us are still learning what's the best way to do it.

However we're still keeping the kind of complex higher level tests, so the existing (selenium) coverage is not going away.
Of course ideally some backend services tests should be created as well to increase test coverage in community. If you could point me to some examples I might be able to give a hand with that.

Salaboy:
@jhrcek that will be awesome.. I will send you the links offline.. or please contact me in the IRC later today or when you have time. Regards

jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/src/main/java/org/jbpm/console/ng/ht/client/editors/taskcomments/TaskCommentsPresenter.java
```diff
@@ -87,81 +78,79 @@ public IsWidget getView() {
...
+            view.redrawDataGrid();
+        }
+    }
+
+    private static class ShowErrorInModalCallback implements ErrorCallback<Message> {
+
+        @Override
+        public boolean error(Message message, Throwable throwable) {
+            ErrorPopup.showMessage("Unexpected error encountered : " + throwable.getMessage());
+            return true;
```
yurloc: @jhrcek Why is the `message` parameter ignored?

jhrcek:
@yurloc Don't ask me, I just extracted existing implementation into static inner class :) What do you think @Salaboy ?

jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/src/main/java/org/jbpm/console/ng/ht/client/editors/taskcomments/TaskCommentsPresenter.java
```diff
@@ -87,81 +78,79 @@ public IsWidget getView() {
...
+                        view.redrawDataGrid();
+                    }
+                },
+                new ShowErrorInModalCallback()
+        ).getAllCommentsByTaskId(currentTaskId);
+    }
 
+    public void addTaskComment(final String text) {
+        if ("".equals(text.trim())) {
+            view.displayNotification("The Comment cannot be empty!");
```
yurloc: The message should be internationalized.

Salaboy: +1

jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/src/main/java/org/jbpm/console/ng/ht/client/editors/taskcomments/TaskCommentsPresenter.java
Salaboy: +1 to @jhrcek comment below, we need a generic class to use in all the error callbacks. Probably there is already one in Uberfire Extension that we can use.

manstis: There's this (but it'll only show a "generic" error message):

https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/callbacks/DefaultErrorCallback.java

Or there's this (that you can customize to perform different actions for
different Throwables):

https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-widgets/kie-wb-common-ui/src/main/java/org/kie/workbench/common/widgets/client/callbacks/CommandDrivenErrorCallback.java

The latter extends ```HasBusyIndicatorDefaultErrorCallback``` (so your View
would need to implement HasBusyIndicator).

Salaboy:
merged in master

Salaboy:
@jhrcek do you think that we can do this for the task details and task context tabs as well?

jhrcek:
@Salaboy I think It might be possible in the future, but currently we're rather busy with planing and test development for the upcoming features. I'll take a look at it when I have a bit more time.

Salaboy:
@jhrcek I do believe that before planning for new features we should try to cover the existing features. Can you please contact me to my personal email so we can discuss about that? I would like to understand the planning a little bit better

jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/src/test/java/org/jbpm/console/ng/ht/client/editors/taskcomments/TaskCommentsPresenterTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @Test
+    public void commentInputClearedAfterCommetAdded() {
+        String comment = "Working on it, man.";
+        presenter.addTaskComment(comment);
+
+        // Comment added
+        verify(commentsServiceMock)
+                .addComment(anyLong(), eq(comment), eq(USR_ID), any(Date.class));
+        // Input cleared
+        //verify(viewMock).clearCommentInput(); //TODO - will fail, because callback actually never called
```
Rikkola: If you want a tip for mocking the service calls, and getting the callbacks called, here is what I have done. I also think there was some helper class for this in Uberfire.

https://github.com/droolsjbpm/drools-wb/blob/397cda7db60eacb8bed3900a4ddd9068c86a7f41/drools-wb-screens/drools-wb-test-scenario-editor/drools-wb-test-scenario-editor-client/src/test/java/org/drools/workbench/screens/testscenario/client/KSessionSelectorTest.java#L253-252

jhrcek: Thanks for the tip Toni! I'll review it and try to incorporate it in the test.

PR #170, opened 2015-07-20
Update TaskCommentsPresenterTest to verify callbacks

jhrcek:
Extended task comments gwt unit test so that it now also verifies app.logic in callbacks. I used CallerMock from uberfire-testing-utils for this purpose. Using this I discovered and fixed an inefficiency - 2 superfluous calls to redrawDataGrid().


Salaboy:
@jhrcek I kinda like this approach, let's get one test refined so then we can start testing all the other screens in the same way. I will test this and merge today

Salaboy:
merged in master

PR #173, opened 2015-07-23
Simplify and refactor TaskProcessContextPresenter to be unit testable

jhrcek:
To make it possible to write more UI (presenter) unit tests in the future the following design principles from MVP pattern (more verbosly described by Procelli in [1]) should be respected:

1) isolate all interaction with widgets in the View (= don't leak widgets into presenters via view methods like 
BadView { TextField getXYZInput() } -> GoodView { void setXYZInput(String)}; because it makes mocking much harder.
2) view should be as dumb as possible and contain primarily getters and setters of primitive values
3) don't rely on view to store intermediate state - store it in Presenter instead
4) don't put application logic into view, put it in presenter instead. That way more can be tested in unit tests.

[1] https://docs.google.com/document/d/11V4ebVL2F-OySZrcAbfUvJyqugoit-Or2t6T7mGPsMo/edit


Salaboy:
Jan sorry for the delay.. I'm looking into this one now

Salaboy:
@jhrcek thanks! merged in master

PR #175, opened 2015-07-31
Ace editor and UberfireDocks support added

wmedvede:
- Ace editor support was added
        Editors like the deployment descriptor editor or the datamodeler doesn't work without the Ace editor support.
    - Data modeller uberfire docks was added

### DIFF
--- a/jbpm-console-ng-showcase/pom.xml
+++ b/jbpm-console-ng-showcase/pom.xml
@@ -467,6 +467,12 @@
       <artifactId>uberfire-widgets-service-backend</artifactId>
     </dependency>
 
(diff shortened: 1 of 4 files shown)


jbpm-console-ng-showcase/pom.xml
```diff
@@ -691,6 +697,21 @@
       <scope>provided</scope>
     </dependency>
 
+    <!-- Needed for the Deployment Descriptor Editor -->
```
manstis: Really!!? I'd have thought it might have need the XML Text Editor in kie-wb-common.. but the DRL editor!?!

PR #176, opened 2015-07-31
JBPM-4625 Generation of JPA enabled models

wmedvede:
- automatic adding/update of the JPAMarshalling strategy in the deployment-descriptor.xml when the persistence.xml is created/modified

Given that the deployment-descriptor.xml editor and the persistence.xml editor are completely decoupled, and also the deployment-descriptor.xml stuff is only included in WB's that includes the jbpm runtime, etc., the following PR implements the following functionality in order to maintain the marshalling strategy configuration well.

    1) when the persistence.xml is added/modified then if the deployment-descriptor.xml exist for the given project, then the JPAMarshalling strategy will be added/modified.

    2) when the deployment-descriptor.xml is created, if the persistence.xml exists for the given project, then the JPAMarshalling strategy configuration will be added to the just created descriptor.

Finally in all cases the user can always change the deployment-descriptor.xml using the editor, and eventually remove the buy default added configuration.

### DIFF
--- a/jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-backend/pom.xml
+++ b/jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-backend/pom.xml
@@ -97,6 +97,10 @@
       <groupId>org.guvnor</groupId>
       <artifactId>guvnor-services-api</artifactId>
     </dependency>
+    <dependency>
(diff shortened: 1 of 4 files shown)


mswiderski:
looks good to me, from my point of view +1 to merge it

Salaboy:
merged in master and 6.3.x

PR #177, opened 2015-08-04
BZ-1248020 JobExecutor: Job Service Settings dialog active despite di…

mswiderski:
…sabled job executor


Salaboy:
merged in master.. more checking are coming in a separate PR

PR #178, opened 2015-08-07
BPMSPL- 4721 Task quick search feature is broken

nmirasch:


mswiderski:
@nmirasch maybe we should avoid using wild cards on both sides of the expression as it will essentially degrade performance of the queries as it will on most of data bases ignore indexes
```
"%" + textSearchStr + "%"
```
could be replaced with
```
 textSearchStr + "%"
```
still if someone would like to search with contains instead of starts with could add % explicitly in the search box. Wdyt?

nmirasch:
I removed the wild card  at code, leaving the user the responsibility  of include it.

mswiderski:
great, I'll merge it in later today, thanks

Salaboy:
merged in master and 6.3.x

PR #180, opened 2015-08-12
BPMSPL-156: Process dashboard refactor

dgutierr:
The older process dashboard iframe (based on the external dashbuilder webapp) has been entirely replaced by an Uberfire native perspective built on top of the UF dashbuilder module. The new process dashboard feeds from two non-public (hidden to end users) data sets (see taskMonitoring & processMonitoring data set registration in DashbuilderBootstrap).

This new process dashboard also includes the following feature:

https://issues.jboss.org/browse/BPMSPL-17

... which allows the user to list the related instances shown in the dashboard as well as the ability to open a details screen for every process or task instance selected.


Salaboy:
merged in master

PR #183, opened 2015-09-03
JBPM-4724: Some dashbuilder DataSet columns used by console have wron…

nmirasch:
…g data types


Salaboy:
merged in master only

PR #194, opened 2015-10-13
BZ-1270690 - Task list: Active Tab shows completed tasks

nmirasch:
Reduce the number of executed queries while the tabs are regenerated


jhrcek:
@nmirasch  The fact that this line of code had to be added to five different places is a bit disconcerting. Not sure that I missed something, but the implementation of the method restoreTabs() looks identical in all five DataSet*ViewImpl classes. Wouldn't it be good to pull it up to a superclass? What do you think?

Also it would be nice to have some tests :-)

nmirasch:
You are right the same code in every section that manages tabs. We had a cleaning task of classes that are not exposed and sure includes code optimization. 
We also, planned add the UI test after this period of BZ resolving.

jhrcek:
Neus, I built jbpm-console-ng with this PR locally and the issue seems to be gone. I'm no longer able to reproduce the issue. I also executed existing task list tests and this doesn't seem to break anything. I vote +1 for merging this PR to 6.3.x

Salaboy:
merged in 6.3.x

PR #195, opened 2015-10-14
BZ-1268219 - MySQL case sensitive table names - task list view error

nmirasch:
jbpm-console-ng showcase with the queries adaptation for case sensitive db.


krisv:
+1, similar change already applied in kie-wb-distributions:
https://github.com/droolsjbpm/kie-wb-distributions/commit/08f2b571adc2b31597c436cde48e3bc3f1c673c6

Salaboy:
merged in 6.3.x

PR #196, opened 2015-10-14
BZ-1268219 - MySQL case sensitive table names - task list view error

nmirasch:
jbpm-console-ng showcase with the queries adaptation for case sensitive db in master


krisv:
+1, similar change already applied in kie-wb-distributions:
https://github.com/droolsjbpm/kie-wb-distributions/commit/08f2b571adc2b31597c436cde48e3bc3f1c673c6

psiroky:
It seems that exactly same changes are needed in multiple places. Would it be possible to share these queries between jbpm-console-ng-showcase, kie-wb and kie-drools-wb?

Salaboy:
merged in master

PR #197, opened 2015-10-15
JBPM-4602: Process Dashboard - BS3 fine tuning

dgutierr:
Latest changes and adjustments related with the BS3 upgrade.


Salaboy:
@cristianonicolai I've assigned this to you, because you probably know more about the BS3 upgrade than myself

jbpm-console-ng-dashboard/jbpm-console-ng-dashboard-client/src/main/java/org/jbpm/dashboard/renderer/client/panel/ProcessDashboardViewImpl.ui.xml
```diff
@@ -38,20 +38,21 @@
     }
 
     .header {
-      padding-left: 10px;
-      padding-bottom: 20px;
+      padding: 15px;
       background-color: white;
     }
 
+    .breadcrumb {
```
cristianonicolai: For the breadcrumb component, PatternFly already has a component https://www.patternfly.org/widgets/#breadcrumbs.
I guess that should be in the j:ProcessBreadCrumb element, so no custom css would be needed in here.

cristianonicolai:
@dgutierr I'm wondering why both Tasks and Processes dashboards are combined into a single perspective. Is there any specific reason for that and not use separated perspective for each one?

jbpm-console-ng-dashboard/jbpm-console-ng-dashboard-client/src/main/java/org/jbpm/dashboard/renderer/client/panel/ProcessDashboardViewImpl.ui.xml
dgutierr: Good point. I'm going to see if the standard pfly breadcrumb cab be used. Thanks.

dgutierr:
Well, Processes and Tasks are two separated widgets. Having them into a single or independent perspectives it is easy to achieve. We wen for single perspective as we needed specific control over the "tab selector widget". I'm not sure if we can get the same behaviour for switching between processes and tasks using separated perspectives.

cristianonicolai:
yep, instead of switching tab, we switch in the navigation menu. I think that is more standard compared to other features. I'm thinking about for example task list and process list, they are independent. My 2 cents would be to use different perspectives and keep similar to what we have across the workbench.

cristianonicolai:
@dgutierr one more thing :) If you could squash these commits into a single one, I would really appreciate. I usually keep amending the same commit until is final.

dgutierr:
Squashed :-)

psiroky:
jenkins, retest this (looks like the failure was caused by stale SNAPSHOTs)

cristianonicolai:
@dgutierr looking good, thanks! mate, the only thing I can still say is about using the font-size: small; on breadcrumb. My concern with that is that if we start redefining fonts on different screens we start to loose consistency. I did check how it would impact, removing it, and doesnt look bad to me, just standard PatternFly breadcrumbs I guess.

One tip maybe for another PR is that I think you might want to review how you created those "cards", PatternFly did included that just recently, have a look, I believe you could reuse it: 

https://rawgit.com/patternfly/patternfly/master/tests/cards.html
https://www.patternfly.org/patterns/aggregate-status-card/

dgutierr:
@cristianonicolai Yes, you are right. I've removed the "font small" and looks goods. 

As for using the new pfly cards, that would be fantastic. I didn't know they existed. Let's see what can we do to integrate them.

Salaboy:
Is this still work in progress or is it ready to merge?

dgutierr:
@Salaboy Yes, it is ready to merge.

cristianonicolai:
Nice, well done @dgutierr !

Merged to master 39eb94aeae96010499ff8f472a030367017e9ddb

PR #203, opened 2015-10-26
DASHBUILDE-58: Remove quotes in json for numeric filter parameters

nmirasch:
Changed the JSON marshaller after dashbuilder update.

### DIFF
--- a/jbpm-console-ng-generic/jbpm-console-ng-data-filters-client/pom.xml
+++ b/jbpm-console-ng-generic/jbpm-console-ng-data-filters-client/pom.xml
@@ -132,6 +132,25 @@
       <scope>provided</scope>
     </dependency>
 
+    <!-- test -->
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.gwt.gwtmockito</groupId>
+      <artifactId>gwtmockito</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+    </dependency>
+
   </dependencies>
 
 </project>
--- a/jbpm-console-ng-generic/jbpm-console-ng-data-filters-client/src/test/java/org/jbpm/console/ng/df/client/filter/json/DataSetLookupJSONMarshallerTest.java
+++ b/jbpm-console-ng-generic/jbpm-console-ng-data-filters-client/src/test/java/org/jbpm/console/ng/df/client/filter/json/DataSetLookupJSONMarshallerTest.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2015 JBoss by Red Hat.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.console.ng.df.client.filter.json;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.gwt.json.client.JSONArray;
+import com.google.gwt.json.client.JSONObject;
+import com.google.gwt.json.client.JSONString;
+import com.google.gwtmockito.GwtMock;
+import com.google.gwtmockito.GwtMockitoTestRunner;
+import org.dashbuilder.dataset.DataSetLookup;
+import org.dashbuilder.dataset.filter.ColumnFilter;
+import org.dashbuilder.dataset.filter.DataSetFilter;
+import org.dashbuilder.dataset.filter.FilterFactory;
+import org.dashbuilder.dataset.group.AggregateFunctionType;
+import org.dashbuilder.dataset.group.ColumnGroup;
+import org.dashbuilder.dataset.group.DataSetGroup;
+import org.dashbuilder.dataset.group.GroupFunction;
+import org.dashbuilder.dataset.group.Interval;
+import org.dashbuilder.dataset.sort.ColumnSort;
+import org.dashbuilder.dataset.sort.DataSetSort;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+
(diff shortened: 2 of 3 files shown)


psiroky:
There are no tests in this PR, that does not seem correct.

The changes are also very similar to the ones done in Dashbuilder (https://github.com/dashbuilder/dashbuilder/commit/3f883a17ecef1b9e31d1cf48b20f413694337cfe). Are we really duplicating the logic here?

csadilek:
Yes, as @psiroky pointed out, the ```parseInterval```, ```formatInterval```,```parseValue```, and ```formatValue``` all seem present in dashbuilder's ```DataSetLookupJSONMarshaller```.

Can we aim making them reusable or alternatively moving them somewhere so that they are?

Tests are missing :(.

Salaboy:
@nmirasch can we work on trying to remove those classes from the jbpm-console-ng repository? So we can rely in the dashbuilder tests? 
@csadilek @psiroky are you guys going to agree if we remove duplicated classes and rely on dashbuilder tests?

csadilek:
@Salaboy @nmirasch sure, if the same is done also for ```formatDataSetGroup```, ```parseSelectedIntervals```, ```parseCoreFunctionParameters``` which seems to be duplicated as well.

There won't be a change left in this commit though as the rest are formatting changes, so no tests required then ;).

Salaboy:
@csadilek the only problem that I see, and the main reason why we have the duplicated code is the fact that the dashbuilder client library is a big chuck full of stuff. We couldn't manage to just get these JSonMarshallers alone without depending on 100% of the dash builder client side. In the long run it sounds as a really bad idea to make two modules depend on each other so tightly. IMO we should review how the Dashbuilder client side libs as bundled because at this point is all or nothing. If you take a look at the duplicated classes we are only using a very small subset of functionality and hence depending on the whole stuff doesn't make any sense to me.

csadilek:
@Salaboy well, I am sure you agree, that a refactoring that allows reusing all this code is better than duplicating it, especially without tests. It also seems the jbpm-console-ng is already depending on dashbuiler-client-all....so you won't introduce any more or less coupling, but sure a refactoring of this code to a smaller reusable module would be the best solution.

Salaboy:
@csadilek I totally agree.. I will vote for that refactoring, because having the dashbuilder-client-all is only ok for the showcase (which is a distribution) but in theory you should be able to use the task lists or the process lists as components in another distribution without depending on the entire dashbuilder-client-all. So I will vote to split the dashbuilder-client-all module in order to have the filters in a separate and small module. -> @nmirasch @dgutierr

psiroky:
Unless I am missing something, you don't need to depend on `dashbuilder-client-all`. It does not contain any code directly, it is just a dependency chain which transitively depends on all dashbuilder client deps. You can still depend only on the specific dashbuilder-*-client modules, like `dashbuilder-displayer-client`.

Salaboy:
@psiroky yes we know that, you are missing the fact that we don't want to get all the displayer-client either.. that's also a big chunk of stuff. The only thing we need are the filters.

dgutierr:
I do not see any issues depending on dashbuilder-displayer-client, as it the same case as depending on uberfire-widgets-commons or any other shared module that contains several widgets. We can not pretend having a single module for every widget we want to reuse.  Moreover, afaik in the near future we will be reusing even more stuff from dashbuilder-displayer-client in the jbpm-console, since the goal is to deliver a fully customizable task list filter editor based on displayers and the dashbuilder's built-in displayer editor.

nmirasch:
Added a very basic test structure

jbpm-console-ng-generic/jbpm-console-ng-data-filters-client/pom.xml
```diff
@@ -132,6 +132,25 @@
       <scope>provided</scope>
     </dependency>
 
+    <!-- test -->
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
```
psiroky: Nit picking here, but it would be better to use `mockito-core`. `mockito-all` is uberjar which bundles all the transitive dependencies inside the same jar.

jbpm-console-ng-generic/jbpm-console-ng-data-filters-client/src/test/java/org/jbpm/console/ng/df/client/filter/json/DataSetLookupJSONMarshallerTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @GwtMock
+    protected JSONObject jsonObject;
+
+    @GwtMock
+    protected JSONArray jsonArray;
+
+    protected DataSetSort datasetSort1;
+    protected DataSetFilter datasetFilter1;
+    protected DataSetGroup datasetGroup1;
```
psiroky: Nit picking again...is there a reason for the "1" suffix?

jbpm-console-ng-generic/jbpm-console-ng-data-filters-client/src/test/java/org/jbpm/console/ng/df/client/filter/json/DataSetLookupJSONMarshallerTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+        dataSetLookupJSONMarshaller.toJson(dataSetLookup);
+
+        verify(dataSetLookup,times(1)).getDataSetUUID();
+        verify(dataSetLookup,times(1)).getNumberOfRows();
+        verify(dataSetLookup,times(1)).getRowOffset();
+        verify(dataSetLookup,times(1)).getOperationList(DataSetFilter.class);
+        verify(dataSetLookup,times(1)).getOperationList(DataSetGroup.class);
+        verify(dataSetLookup,times(1)).getOperationList(DataSetSort.class);
+        //assertDataSetDef(filterSettings,restored);
```
psiroky: We should avoid committing commented code.

psiroky:
Thanks for the updates @nmirasch. I see there are tests now. Are you also planning to re-use the logic directly from dashbuilder (instead of the copy there is right now)?

csadilek:
Thanks @nmirasch. Could you give us a quick update on the plan? I thought with @dgutierr's refactoring you might be able to reuse his code or is that not the plan for 6.3 (only for master)?

nmirasch:
I think the plan for future is unify the code (master) but this pr is just for restore the build.

csadilek:
@nmirasch OK, I think with the limited time we have left, that makes sense. Can you squash the commits  in this PR, then I will cherry-pick to 6.3.x, and send a separate PR with the refactoring to master?

nmirasch:
it's done

nmirasch:
Now I'm going to generate it for master in another PR

csadilek:
Thanks @nmirasch. So, on master, we can get rid of the code duplication thanks to @dgutierr refactoring.

nmirasch:
@csadilek  On master we have to work to avoid the code duplication but it's not direct

csadilek:
Thanks.

6.3.x: 887cbd940772de9c826fce24cbb6c686f6e779ec

PR #209, opened 2015-11-03
DASHBUILDE-58: Remove quotes in json for numeric filter parameters

nmirasch:
This pr include the refactor to avoid duplicated classes for json marshaling, and the test related.


jbpm-console-ng-generic/jbpm-console-ng-data-filters-client/src/test/java/org/jbpm/console/ng/df/client/filter/FilterSettingsJSONMarshallerTest.java
```diff
@@ -0,0 +1,149 @@
...
+        DisplayerSettingsJSONMarshaller displayerJsonMarshaller = new DisplayerSettingsJSONMarshaller(datasetJsonMarshaller, lookupJsonMarshaller);
+        filterSettingsJSONMarshaller = new FilterSettingsJSONMarshaller(displayerJsonMarshaller);
+    }
+
+    @Test
+    public void testJSONConversion() throws Exception {
+        FilterSettings filterSettings = createTableSettings();
+        String json = filterSettingsJSONMarshaller.toJsonString(filterSettings);
+        FilterSettings fs = filterSettingsJSONMarshaller.fromJsonString(json);
+        assertTrue("Check the number of operations",
```
jhrcek: It's nice to see that duplicated code from dashbuilder to be gone!
I'm also glad that we started working on some test coverage. With several tests like these we can easily expand the coverage in the future.

Just a small nitpick about this assert. In the cases like this I would recommend using
assertEquals(message, expected, actual)
instead of
assertTrue(message, expected==actual)

This way you'll get more informative error messages in case of failing tests (the former will give you something like "expecting A to be equal to B", whereas the latter just "expecting true but was false")

porcelli: +1 for the comment (have clear asserts)... but due the emergency (master broken), I'm considering OK for merge.

porcelli:
https://github.com/droolsjbpm/jbpm-console-ng/commit/0c9c7853d7af3154f715ea83cc33c3f1b05104c6

csadilek:
@porcelli wasn't here and isn't aware of the original PR: https://github.com/droolsjbpm/jbpm-console-ng/pull/203. We held back with this on purpose (for master) to get the refactoring in and I am very glad we have it this now. This saves ~900 lines of duplicated code! So, that's great.

porcelli:
@csadilek +1M for such approach!

PR #211, opened 2015-11-06
BZ:1277466 - No tasks shown in task list

nmirasch:
There is a problem with the war generated from jenkins build and the dataset conditions with datasets and their json generated, that I can't reproduce in my local environment.
This code does the same that it did before but I want to check what is the behavior  in the next build with this code.


jhrcek:
+1 to merge. Please review and merge asap. This PR is related to investigation about what's causing the critical tasklist bug (BZ 1277466 -not tasks shown in the task list).

csadilek:
I don't think we should push upstream just for investigation purposes. In this case, the new code is cleaner than the old one though (got rid of using exceptions as control flow).

6.3.x: b9ee18604c21730050ad4af3048b00cc996c0c2c

Thanks.

PR #215, opened 2015-11-10
BZ:1275245 - Oracle DBs: Tasks is status 'Ready' not visible in task …

nmirasch:
…list
The problem with the oracle database is Oracle11 considers the value '' as null. I added another condition to the dataset query to control that situation.


csadilek:
6.3.x: 34d5bfbdd064652d05f9ea3281b0acd02d07a38a

The fact that none of this code is tested / maintainable is very concerning. This particular error we would have only found with an integration test against Oracle, but we still need to investigate how to refactor this code/project to add more unit tests.

csadilek:
Thanks for the fix @nmirasch.

csadilek:
Please don't forget to send a PR to master if the fix should be applied there as well.

PR #216, opened 2015-11-11
BZ:1277466 - No tasks shown in task list. 

nmirasch:
I introduced a bug creating a constructor for doing test. The window to create addon task doesn't close after creation.
Here is the fix, removing the @inject. This pr doesn't take test  because they are included in previous pr and this code doesn't change anything in that sense


csadilek:
Hi @nmirasch, This commit message is a duplicate again. Please change the title or amend some details so we can keep them apart more easily.

csadilek:
Thanks.

6.3.x: 338e990a53439cf5a5c306ad9a35be2da998ac8b

Yes, there are tests in the previous PR but the fact that this bug slipped through is indication enough that the coverage is not where it should be. Let's make sure we improve the quality of this project going forward!

PR #217, opened 2015-11-11
BZ:1261052 - Console: various localized perspective listings does not…

nmirasch:
… work.
This commit needs to solve the issue the one asked in uf-ext. We need generate the test for the ui at console ui but for  generate really useful test, a big refactor is needed. And We'll do it at master branch.

### DIFF
--- a/jbpm-console-ng-business-domain/jbpm-console-ng-business-domain-client/pom.xml
+++ b/jbpm-console-ng-business-domain/jbpm-console-ng-business-domain-client/pom.xml
@@ -121,6 +121,26 @@
       <artifactId>org.eclipse.sisu.inject</artifactId>
       <scope>runtime</scope>
     </dependency>
+
+    <!-- Test deps -->
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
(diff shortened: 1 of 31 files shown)


cristianonicolai:
Change looks fine to me, the only concern is the lack of tests.

cristianonicolai:
@nmirasch I created a very minimal test that I think you could use for this change. Its the fair minimum you need to check, making sure that dataStoreName is set for all columns.

Have a look in here: https://github.com/cristianonicolai/jbpm-console-ng/commit/0d8bbd5c99a7d995d0898e29b3cb10d6d36e0ab8

It didn't require major refactor and I think you be really nice if you replicate the same test for all views that were changed in here.

nmirasch:
Added the basic test for make sure the columns come with dataStoreName for every view changed

cristianonicolai:
Nice, +1 from me.

csadilek:
Nice work @nmirasch!

6.3.x: bbecb7c760c500a0ecace4916a82ee50621abedd

csadilek:
Hi @nmirasch, @cristianonicolai. I've seen a bunch of commits going only to 6.3.x in this repo. Just a reminder to also submit them to master, if needed! ;)

PR #218, opened 2015-11-12
showcases: use single GWT module and share common props

psiroky:
* showcases are used only for debugging/testing, so there
   really is no real need to have two different GWT modules
   for them (FastCompiled and Full). Having just one module
   should simplify the configuration
 * removed no longer needed log4j.xml files. Logback is used
   as logging impl

@Salaboy, @nmirasch, @cristianonicolai please take a look. Same changes were already applied to guvnor and drools-wb repos.


cristianonicolai:
+1 for the change.

@psiroky @Salaboy @nmirasch @mswiderski  I'm wondering if we still need to keep two showcase apps in this repo. Could we remove jbpm-console-ng-ht-showcase and keep only the jbpm-console-ng-showcase one? Any concerns with that?

psiroky:
@cristianonicolai thanks for bringing this is up. I wanted to suggest the same. I even tried to run `mvn gwt:run` inside ht-showcase and it does not work currently. It seems it is not being maintained.

csadilek:
Thanks.

master: 99ebab5f56377e7854d60efa33f0c0508f5993f8

PR #221, opened 2015-11-16
Remove jbpm-console-ng-ht-showcase

cristianonicolai:


csadilek:
Deleting unneeded code will always get a +1 from me :). Any objections?

porcelli:
@Salaboy you did this, any objections to remove?

Salaboy:
+1 , my only objection is that this module was created to test the modularity of the modules inside the jbpm console ng, but we will need to work on that later on anyway.. and probably using a different approach. So I'm ok with removing this now

porcelli:
https://github.com/droolsjbpm/jbpm-console-ng/commit/83ce60d38f18bcf02c8a7c9f8607f1b063bc90f6

PR #229, opened 2015-11-23
Master: BZ1284004 and BZ1283537

nmirasch:
Modified SQL queries that define the SQL dataset


csadilek:
+1. This will need an integration test in the future to verify the bootstrap process.

csadilek:
The fact that we're duplicating much of this bootstrapping logic in jbpm-console-ng and kie-wb-distributions also needs to be investigated.

csadilek:
master: e761b33bb355f74089de65f005e5a07ff75d0b8e and 957b7a659f08a5cfd9df5b9305cffd09402dc2da

PR #231, opened 2015-11-27
GUVNOR-2304 - Deployment descriptor editor text fields too small

mswiderski:
@csadilek could you please have a look?


manstis:
+1 to merge.

csadilek:
Thanks, that was quick! Looks good now.

I adjusted the column width of the label column as well since it was consuming too much horizontal space.

master: 954aad74bc5e4b7a6933d74097d62c51b561952e

PR #234, opened 2015-12-03
BZ-1276354: (BZ:1261052) Tasks with ID higher than 1000 does not being completed. …

nmirasch:
…(change locale issue BZ:1261052)

### DIFF
--- a/jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/pom.xml
+++ b/jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/pom.xml
@@ -93,6 +93,19 @@
       <artifactId>jbpm-console-ng-generic-client</artifactId>
       <type>jar</type>
     </dependency>
+
+    <!-- Test -->
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.gwt.gwtmockito</groupId>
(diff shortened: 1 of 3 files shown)


cristianonicolai:
@nmirasch I'm guessing this PR is only related to BZ-1261052, is that correct? If so, please adjust the PR title.

Also, is this ok/requested to go on the 6.2.x branch?

jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/pom.xml
```diff
@@ -93,6 +93,19 @@
       <artifactId>jbpm-console-ng-generic-client</artifactId>
       <type>jar</type>
     </dependency>
+
+    <!-- Test -->
+    <dependency>
+      <groupId>org.mockito </groupId>
+      <artifactId>mockito-all</artifactId>
```
jhrcek: Please avoid the usage of mockito-all (uberjar) - instead use mockito-core. 
See https://issues.jboss.org/browse/DROOLS-999 for details.

nmirasch:
Yes this is the fix pending for the 6.1.5 patch, then it has to go to 6.2.x, they reported an issue related with the task id greater than 1000 when the locale changes. The problem with the task id was resolved but the last comments of the bz are related with the change of the locale over the datagrid. I include the same solution as  BZ:1261052, but in this case, it helps to close BZ1276354. This is what I tried to comment in the commit because someone is expecting a fix for this last one.

csadilek:
6.2.x: f8833c955d8313c44adf188c2892d46930891ff5

PR #238, opened 2015-12-07
GUVNOR-2278: Added missing dependencies

nmirasch:


psiroky:
Merged as it fixes build error.

PR #248, opened 2015-12-23
JBPM-4851: Highlight the 'all' filter by default

dgutierr:
The fix + test of the following JIRA:

https://issues.jboss.org/browse/JBPM-4851


romartin:
Hey @dgutierr , 

I have check it, no compilation neither test issues, but i was expecting to have the "All" filter enabled by default , and the default one is "Active"... was this the fix? If so, it's not working....

jbpm-4851

romartin:
note: I used the same jbpm-console-ng-showcase to test it in SDM

jhrcek:
@romartin  It seems that you've been checking at the wrong place. The was not about Tasks perspective (where the 'Active'' filter tab activation by default is desirable), but about Dashboards > Process & Task dashboards perspective (the metric displayer 'All' for selection of all visible processes).

csadilek:
+1

csadilek:
master: d42b02a32a0d82010944b70de9090bed5da24517

PR #251, opened 2016-01-11
Fix regex to allow more types of version strings

psiroky:
* we need the version strings 6.4.0.2015011 or 6.4.0.test to
   work as well

@mrietveld could you please review this?


psiroky:
I will apply the same fix for `org.jbpm.runtime.manager.impl.deploy.DeploymentDescriptorMergerTest.changeDefaultLimitSerializationClassesValueToTrueIn7x` once this one is done.

psiroky:
@mswiderski, @cristianonicolai could you please take a quick look? Marco is on PTO and this fixed failing test.

mswiderski:
@psiroky looks ok to me as far as I understood the test - most relevant is the major version of the version string so the snapshot suffix is not relevant. If that's valid then +1 from me

psiroky:
@mswiderski yes, my understanding is the same. I will merge it then.

PR #252, opened 2016-01-11
Minor tweaks for DeploymentDescriptorViewTest

psiroky:
* use java.util.UUID instead of org.jgroups.util.UUID

 * use Collections.singletonList() instead of Arrays.asList
   as it makes the intention clearer

I believe the org.jgroups.util.UUID was used by mistake. If not, I will revert the change and add comment that it is on purpose.

@mrietveld please review.


cristianonicolai:
+1

csadilek:
master: 2f1b62889c0ec88863fdbc951d43736078714776

PR #266, opened 2016-01-21
JBPM-4850: Add warning when clicking reset default filters

nmirasch:


cristianonicolai:
+1 thanks @nmirasch

cristianonicolai:
@nmirasch just a suggestion, but what do you think if we change the reset default filters icon to this one: http://fontawesome.io/icon/filter/ ?

csadilek:
@nmirasch hi, do you have an update on this? please don't forget that we need it on the 6.4.x branch as well, when ready.

PR #272, opened 2016-01-27
JBPM-4902: Every module takes care of registering its own datasets

dgutierr:
Every functional module: human tasks, process instances, process dashboard & executor service is in charge of registering at initialization time (@Startup) the required dataset definitions. That way we avoid other downstream modules (kie-wb-distributions) having duplicating code.


cristianonicolai:
nice! +1 thanks @dgutierr

csadilek:
master: 4475f21a285dce9ffb49c5c46cbc0c629c406207

PR #274, opened 2016-01-27
JBPM-4902: Every module takes care of registering its own data sets (6.4.x)

dgutierr:
Every functional module: human tasks, process instances, process dashboard & executor service is in charge of registering at initialization time (@Startup) the required dataset definitions. That way we avoid other downstream modules (kie-wb-distributions) having duplicating code.


cristianonicolai:
+1

csadilek:
6.4.x: 1ed3bbdeebf2c898b058d6bff72c8ed49b7d3b28

PR #286, opened 2016-01-28
Dashboard updates

cristianonicolai:
Combined common code between Task and Process dashboard classes, avoiding code duplication
UI styles update, manly to allow content to be responsive


cristianonicolai:
Hi @dgutierr, please take a took at this change, if makes sense to you.
There is one thing still missing that I would like to change but couldn't find a way. That is to set a 100% width for a MetricDisplayer and a TableDisplayer. As far as I can tell, you can only set specific numbers and not percentage. Not sure if I'm missing something. That would be for the metrics and the tables on the dashboard to use the entire space available on the screen, independently from the client resolution.

cristianonicolai:
@dgutierr thanks for your feedback. I've updated my PR with your recommendations.
About the code styling, thanks for bringing that to my attention, I realized my IntelliJ project setup wasn't using the proper code style defined in here: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/tree/master/ide-configuration/intellij-configuration/code-style.

csadilek:
@dgutierr did you have time to look at this? If so, please add a plus one so it can be merged.

dgutierr:
+1 Now looks perfect!

PR #290, opened 2016-01-29
JBPM-4849: Add ability to hide asset management process definitions

dgutierr:
Fix + unit tests.

I  tested against EAP 6.4,  both with and without the kiemgmt role and, as expected, the asset management instances are only shown when the kiemgtm role is present.

@cristianonicolai Can you please verify?


cristianonicolai:
+1 Thanks @dgutierr!

csadilek:
master: 0ae69d2dd2a776f903be3512b7ce847a08430988

PR #291, opened 2016-01-29
JBPM-4849: Add ability to hide asset management process definitions (6.4.x)

dgutierr:
Fix + unit tests.

I tested against EAP 6.4, both with and without the kiemgmt role and, as expected, the asset management instances are only shown when the kiemgtm role is present.

@cristianonicolai Can you please verify?


cristianonicolai:
+1

csadilek:
Thanks.

6.4.x: c078dac56c88721e250c8a932d45c53f96b19768

PR #318, opened 2016-02-17
JBPM-4945: Postgresql: can't get metadata - jbpmHumanTasksWithVariables

nmirasch:
Modified the way of doing column aliases in the SQL query.
Using the clause AS that works in postgresql, db2, oracle, h2


cristianonicolai:
@nmirasch I believe jbpmProcessInstancesWithVariables would also need fix in this case.

Can you also check if tasksMonitoring also needs fixing: https://github.com/droolsjbpm/jbpm-console-ng/blob/master/jbpm-console-ng-dashboard/jbpm-console-ng-dashboard-backend/src/main/java/org/jbpm/dashboard/renderer/backend/DataSetDefsBootstrap.java#L66

cristianonicolai:
ok, so I'm inclined to accept this without tests because I really don't see much value of a sql string test. What we really need are tests that could confirm that the query parses and even better, if we could validate running the query against the different databases.

+1

cristianonicolai:
@nmirasch please send this into master as well, thanks.

psiroky:
Merged into 6.4.x.

PR #319, opened 2016-02-17
JBPM-4943: DB2: Task list dataset lookup failing with "column does no…

nmirasch:
…t exist in DB: oeid"
Removed colum oeid alias because there isn't another column in the dataset with the same id


cristianonicolai:
+1 thanks @nmirasch

PR #320, opened 2016-02-17
JBPM-4943: DB2: Task list dataset lookup failing with "column does no…

nmirasch:
…t exist in DB: oeid"
Removed column oeid alias because there isn't conflict with any other column in the select with the same id


cristianonicolai:
@nmirasch could you include a test, similar to what you done on https://github.com/droolsjbpm/jbpm-console-ng/pull/309, checking that the column alias isn't present in the query?

cristianonicolai:
+1 thanks @nmirasch

PR #322, opened 2016-02-19
JBPM-4948 - Fix navigating from process definition to related instances

cristianonicolai:
Using place request to pass the process definition name into the search box.


manstis:
+1 to merge.

PR #324, opened 2016-02-19
BZ1286623: Console: completed process instance log shows null values …

nmirasch:
…for Oracle databases
Oracle returns value null instead empty string like other databases. This commit ensures the empty string in that cases for showing process instance logs


cristianonicolai:
+1

psiroky:
Merged into 6.4.x

PR #327, opened 2016-02-22
DataSet database tests

cristianonicolai:
Included tests for Tasks, Process and Dashboard related DataSet SQL queries.
Running SQL queries against H2 and HSQLDB for more extensive syntax verification.
No data verification, only syntax check
Removed potential use of reserved SQL keyword "at".

### DIFF
--- a/jbpm-console-ng-dashboard/jbpm-console-ng-dashboard-backend/pom.xml
+++ b/jbpm-console-ng-dashboard/jbpm-console-ng-dashboard-backend/pom.xml
@@ -59,9 +59,44 @@
       <artifactId>dashbuilder-dataset-api</artifactId>
     </dependency>
 
+    <!-- Test dependencies -->
     <dependency>
       <groupId>org.dashbuilder</groupId>
       <artifactId>dashbuilder-dataset-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-entitymanager</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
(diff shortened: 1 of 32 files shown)


cristianonicolai:
@manstis or @dgutierr please have look

manstis:
@dgutierr I know little of Dash Board DataSets.. could you please review.

jhrcek:
@cristianonicolai Wow, this looks great! Finally we'll have something to test those SQL queries. Also I didn't know about these H2 compatibility modes.
The only concern I have is that the common test & configuration logic is duplicated in those three modules. Wouldn't it be possible to pull out the common stuff in some shared module?

cristianonicolai:
@jhrcek indeed! we need to consolidate this common code somewhere. Maybe some test utils module, not sure. Right now, due to the various errors we're getting around this, I just wanted to have this added and not cause any major change on 6.4 branch. We could look into refactoring this on master.

dgutierr:
This is a very nice step forward to avoid regressions due to the usage of non ANSI SQL syntax on the jbpm console's datasets.  However, I'm still wondering to what extend the H2 emulated mode supports others DB syntax. As stated on: http://www.h2database.com/html/features.html#compatibility

"For certain features, this database can emulate the behavior of specific databases. However, only a small subset of the differences between databases are implemented in this way."

To be honest, I'm not sure if these tests can 100% guarantee that any SQL construct used in a dataset definition will run on top of all our supported DBs. From what I've seen implementing the different dialects for the dashbuilder SQL provider, there are a lot of things to consider (see https://github.com/dashbuilder/dashbuilder/tree/master/dashbuilder-backend/dashbuilder-dataset-sql/src/main/java/org/dashbuilder/dataprovider/sql/dialect) 

@cristianonicolai  wdyt?

cristianonicolai:
@dgutierr thanks for looking into this. For sure these tests are not meant to be a replacement for testing the specifics databases. H2 compatibility mode does not guarantee much for us apart from _some_ db specific features. To me this is the closest we can do to minimize errors when running those queries. Hopefully it can prevent most problems before hand, without us having to run integration tests using the various databases. But again, this is  for sure not a replacement for testing on each database, just a way to fail fast from my perspective. Just the fact that we're running on H2 and HSQLDB already gives some level of confidence that queries are portable.

nmirasch:
+1

dgutierr:
+1 ... @cristianonicolai totally agree :-)

psiroky:
@cristianonicolai could you please rebase this and fix the merge conflicts?

cristianonicolai:
@psiroky done ;)

csadilek:
6.4.x: caabebce2763ee1d4b4e2c0bd0ccdef541d3683e

PR #340, opened 2016-02-26
BZ-1255695 - kie-deployment-descriptor.xml Source tab doesn't have XM…

mswiderski:
…L syntax highlighting

didn't know how to make a test for this change so any hint/advice is more than welcome


manstis:
+1 to merge.

psiroky:
Merged into 6.4.x.

PR #348, opened 2016-03-03
JBPM-4976 and JBPM-4969: Avoid load filter setting from removed filters

nmirasch:
Avoid to call load user preferences when a new filter is being created.
Identify the user defined filters by their generated key.


cristianonicolai:
+1 thanks @nmirasch

csadilek:
6.4.x: fa95d4a0d1595b559a04eacd2bda1ced02a41a2d

PR #359, opened 2016-03-08
JBPM-4964 - Load data before popup shows

cristianonicolai:
Moved data load logic into presenter.
Use loading/busy popup while data is loading and default popup for error messages.


csadilek:
6.4.x: 86f143222d216f26de30355a59609155697c665c

PR #360, opened 2016-03-08
Fix for JBPM-4960, JBPM-5004, JBPM-4945 and JBPM-4943

cristianonicolai:
Removed column alias as DB2 does not return it by default in the resultset metadata information.
 Simplified query for retrieving process instance variables
 Use sub-select for retrieving task name as in this case DB2 keeps the appropriate alias


csadilek:
master: 1b60051a2b73f8745e0206ef2393e14b49f14f2a

PR #374, opened 2016-03-15
BZ-1203561 - Kbase name and ksession name not reflected in 'Deploymen…

mswiderski:
…ts Unit' page if we deploy kjar through REST API


mrietveld:
+1 

Hahaha.. You were right, Maciej. That was a very easy fix -- although it would have taken me much longer to find it seeing as I don't know the console code that well.

csadilek:
master: 3a8ee8bb29a247ca02dee1782612fe1cc9d243e5

PR #387, opened 2016-04-01
BZ1007859 - Add 'select all' checkbox to Process instances

nmirasch:


jbpm-console-ng-process-runtime/jbpm-console-ng-process-runtime-client/src/main/java/org/jbpm/console/ng/pr/client/editors/instance/list/variables/dash/DataSetProcessInstanceWithVariablesListViewImpl.java
```diff
@@ -372,23 +373,57 @@ public String getValue( ProcessInstanceSummary object ) {
...
+                    selectAllButton.setIcon(IconType.SQUARE_O);
+                    selectAllButton.setText(constants.SelectAll());
+                    selectAllButton.setTitle(constants.SelectAllTooltip());
+                }
+                presenter.refreshGrid();
+                controlBulkOperations();
+            }
+        });
+        extendedPagedTable.getLeftToolbar().clear();
+        extendedPagedTable.getLeftToolbar().add(selectAllButton);
```
cristianonicolai: I think it would actually be handy if this button is placed in the right, just before the bulk actions. I'm imagining something like select all and notify, that way both actions are close together. What do you think @nmirasch?

jbpm-console-ng-process-runtime/jbpm-console-ng-process-runtime-client/src/main/java/org/jbpm/console/ng/pr/client/editors/instance/list/variables/dash/DataSetProcessInstanceWithVariablesListViewImpl.java
```diff
@@ -372,23 +373,57 @@ public String getValue( ProcessInstanceSummary object ) {
...
+        selectAllButton.setIcon(IconType.SQUARE_O);
+        selectAllButton.setTitle(constants.SelectAllTooltip());
+        selectAllButton.setIcon(IconType.SQUARE_O);
+        selectAllButton.addClickHandler(new ClickHandler() {
+            @Override
+            public void onClick(ClickEvent event) {
+                selectedProcessInstances.clear();
+                if (IconType.SQUARE_O.equals(selectAllButton.getIcon())) {
+                    selectedProcessInstances.addAll(presenter.getDisplayedProcessInstances());
+                    presenter.refreshGrid();
```
cristianonicolai: Why exactly do we need to refresh the data? It also seems that we're doing it twice at L396.

jbpm-console-ng-process-runtime/jbpm-console-ng-process-runtime-client/src/main/java/org/jbpm/console/ng/pr/client/editors/instance/list/variables/dash/DataSetProcessInstanceWithVariablesListViewImpl.java
```diff
@@ -372,23 +373,57 @@ public String getValue( ProcessInstanceSummary object ) {
     }
 
     public void initExtraButtons( final ExtendedPagedTable<ProcessInstanceSummary> extendedPagedTable ) {
+        final Button selectAllButton = GWT.create(Button.class);
+        selectAllButton.setText(constants.SelectAll());
+        selectAllButton.setIcon(IconType.SQUARE_O);
+        selectAllButton.setTitle(constants.SelectAllTooltip());
+        selectAllButton.setIcon(IconType.SQUARE_O);
```
cristianonicolai: Duplicated setIcon

nmirasch:
@cristianonicolai  I didn't understand "select all and notify"  What's exactly your suggestion about the button group. Now we have a button that selects or unselect all the table instances.
Are you talking about change the behavior of the button, add a new one..? 
In this commit I put the button before the bulk actions button group

cristianonicolai:
@nmirasch "select all and notify" was just to exemplify the use case. My suggestion with the button group is to instead of adding two separated buttons to the table toolbar, you would create a button group containing both. It is really just a visual change.
Another point, I still not quite sure why we need to refresh the data when select/unselect all.

PR #394, opened 2016-04-05
BZ1323272 - DB: Process Instances history for process variables does …

nmirasch:
…not work for all Oracle DB


cristianonicolai:
+1 thanks @nmirasch

PR #396, opened 2016-04-06
BZ1318798 - Default Task List Columns Should Change

nmirasch:


cristianonicolai:
6.4.x branch is locked for blockers only.

PR #399, opened 2016-04-06
JBPM-5029: Make it clear in Jobs perspective that "Job name" will be …

nmirasch:
…used as a "businessKey" for that executor job


manstis:
+1 to merge.

PR #404, opened 2016-04-13
JBPM-5061: Jobs perpertive replaces all tab with queued when the user…

nmirasch:
… leave and return the section

### DIFF
--- a/jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-client/src/test/java/org/jbpm/console/ng/es/client/editors/requestlist/RequestListViewTest.java
+++ b/jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-client/src/test/java/org/jbpm/console/ng/es/client/editors/requestlist/RequestListViewTest.java
@@ -80,6 +80,14 @@ public class RequestListViewTest {
 
         verify(filterPagedTableMock, times(7)).getMultiGridPreferencesStore();
         verify(filterPagedTableMock, times(7)).saveTabSettings(anyString(), any(HashMap.class));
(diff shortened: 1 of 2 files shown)


jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-client/src/test/java/org/jbpm/console/ng/es/client/editors/requestlist/RequestListViewTest.java
```diff
@@ -80,6 +80,14 @@ public void setDefaultFilterTitleAndDescriptionTest() {
 
         verify(filterPagedTableMock, times(7)).getMultiGridPreferencesStore();
         verify(filterPagedTableMock, times(7)).saveTabSettings(anyString(), any(HashMap.class));
+        verify(filterPagedTableMock).saveTabSettings(eq(RequestListViewImpl.REQUEST_LIST_PREFIX + "_0"), any(HashMap.class));
```
cristianonicolai: @nmirasch would be useful if we also check that these methods are called in the expected order as well?
http://site.mockito.org/mockito/docs/current/org/mockito/Mockito.html#6

nmirasch: @cristianonicolai I can add the check but I think the order is not important here, just check that all calls are made for all default tabs.

cristianonicolai:
+1 thanks @nmirasch

PR #410, opened 2016-04-19
Use DefaultErrorCallback where possible

jhrcek:
Replace repeated occurrences of the following

    new ErrorCallback<Message>() {
        @Override
        public boolean error( Message message, Throwable throwable ) {
            ErrorPopup.showMessage( constants.UnexpectedError(throwable.getMessage()) );
            return true;
        }
    }

with 'DefaultErrorCallback' from Uberfire extensions which does the same, but with possibly improved error messages.

I limited the scope of code reformatting only to methods containing UF Caller calls, with the exception of the first 2 files - i reformatted these as a whole (in the beginning of PR).
Also removed unused/ commented-out methods from the first 2 files.
@cristianonicolai Could you please review?


cristianonicolai:
nice, thanks @jhrcek!

+1

PR #411, opened 2016-04-21
Remove GWT dependent classes from API modules

cristianonicolai:
Moved classes into client modules


cristianonicolai:
@nmirasch could you have a look on this, jbpm-console-showcase app fails to me without this fix.

nmirasch:
+1  @cristianonicolai This modifications are working fine, but I have to say that without this pr, jbpm-console-showcase app didn't fail for me.

PR #415, opened 2016-04-25
jbpm-console-ng is failing tu build on master. Fix import org.uberfir…

nmirasch:
…e.ext.widgets.table.client.Datagrid 
Compilation failure fix.


csadilek:
+1

PR #416, opened 2016-04-28
Initial code checks using PMD

cristianonicolai:
Included minimal set of rulesets: basic, imports and unused code.
Fixes for conforming with applied rules
Code formatting on changed files

### DIFF
--- a/jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-api/src/main/java/org/jbpm/console/ng/wi/dd/model/DeploymentDescriptorModel.java
+++ b/jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-api/src/main/java/org/jbpm/console/ng/wi/dd/model/DeploymentDescriptorModel.java
@@ -242,6 +242,7 @@ public class DeploymentDescriptorModel {
         return true;
     }
 
+    @SuppressWarnings("PMD.AvoidMultipleUnaryOperators")
     @Override
     public int hashCode() {
         int result = persistenceUnitName != null ? persistenceUnitName.hashCode() : 0;
--- a/jbpm-console-ng-bpm-home/jbpm-console-ng-bpm-home-client/src/main/java/org/jbpm/console/ng/bh/client/editors/home/HomeViewImpl.java
+++ b/jbpm-console-ng-bpm-home/jbpm-console-ng-bpm-home-client/src/main/java/org/jbpm/console/ng/bh/client/editors/home/HomeViewImpl.java
@@ -16,7 +16,6 @@
 
 package org.jbpm.console.ng.bh.client.editors.home;
 
-import java.util.Collection;
 import javax.enterprise.context.Dependent;
 import javax.enterprise.event.Event;
 import javax.inject.Inject;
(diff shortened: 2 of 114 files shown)


mswiderski:
can we postpone any bigger changes to console codebase I am in a middle of removing the jbpm runtime out of it and it affects "quite few" classes?

I hope to have something that runs today or tomorrow without jbpm runtime...

jhrcek:
Wow, this is great. Now we'll finally be able to focus on relevant changes!  :+1:

jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-api/src/main/java/org/jbpm/console/ng/wi/dd/model/DeploymentDescriptorModel.java
```diff
@@ -245,35 +245,20 @@ public boolean equals(Object o) {
     @Override
     public int hashCode() {
         int result = persistenceUnitName != null ? persistenceUnitName.hashCode() : 0;
-        result = ~~result;
```
jhrcek: Is it ok to delete these `result = ~~result;` ? According to this [stackoverflow answer](http://stackoverflow.com/questions/29736952/what-is-the-meaning-of-double-tilde-in-java) they might make a difference..

manstis: Nooooooo!

We started adding those operators as we found HashCode would behave really
strangely sometimes in GWT compiled code. On investigation GWT uses the ~~
operator extensively in its JRE emulated libraries' HashCode... so we
followed and our HashCode implementations behaved as expected.

Any class destined for the client side should, in my opinion use the ~~
operator in its HashCode implementation... or we'll return to having
"really hard to fix" bugs come back!

manstis:
-1 (use of ~~ operator in client-side HashCode implementations *should* remain).

jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-api/src/main/java/org/jbpm/console/ng/wi/dd/model/DeploymentDescriptorModel.java
cristianonicolai: No worries, I'll review this change to not change the `hashcode` method.

cristianonicolai:
@mswiderski no worries delaying this change but I was expecting we would use some sort of feature branch for myself and Neus to work on the jbpm runtime removal before merging it to master.

mswiderski:
it will be a feature branch but believe me, merging so much changes to feature branch where you'll have several commits diverged is not easy nor fun.

cristianonicolai:
@mswiderski @jhrcek @manstis @nmirasch I've revisited this PR after the latest changes on master, please have another look.

mswiderski:
shouldn't we remove the commented code in QuickNewTaskPopup?

and I agree with @manstis about the client hashCode, but that seems to be already covered. so  +1 from me

PR #423, opened 2016-05-30
JBPM-5030: Access control based on the new UF's authz subsystem

dgutierr:
Changes required to support the new UF's security APIs

This PR depends on uberfire/uberfire#321


porcelli:
+1 (after UF)

PR #436, opened 2016-07-14
JBPM-5071 - remove of jbpm runtime from workbench

mswiderski:
depends on https://github.com/droolsjbpm/jbpm-form-modeler/pull/57


jbpm-console-ng-business-domain/jbpm-console-ng-business-domain-api/src/main/java/org/jbpm/console/ng/bd/model/TaskDefSummary.java
```diff
@@ -24,14 +24,18 @@
...
-        this.name = name;
+    }
 
+    public TaskDefSummary(final String name) {
+        super();
+        this.name = name;
     }
 
     public TaskDefSummary() {
+        super();
```
jhrcek: Explicit call to super(); is not necessary, because call the accessible no-args constructor from the superclass is automatically added by compiler.

The same comment applies to
NodeInstanceSummary
ProcessInstanceSummary
ProcessInstanceVariableSummary
TaskDefSummary
TaskAssignmentSummary

but I guess it doesn't hurt.

mswiderski: interesting as it's always added when you generate constructor via IDE :) I'll leave it for now just to avoid too many updates to PR

Salaboy:
+1

PR #441, opened 2016-07-20
RHBPMS-4117 - jBPM UI - all passed nodes are marked as active

mswiderski:
(cherry picked from commit fe30eaa9c00f9bde95db5430636abc0910cb943f)

same here, @mrietveld do you mind taking a look?


csadilek:
+1

PR #456, opened 2016-08-18
JBPM-5261 Process Definitions: Left-side panel header displays proces…

nmirasch:
…s Id instead of process name


cristianonicolai:
+1 thanks @nmirasch

PR #463, opened 2016-08-25
GUVNOR-2173: Remove custom LuceneConfigProducer

csadilek:
This fixes the build after today's merge of GUVNOR-2173.

To my understanding this custom LuceneConfigProducer should not be needed and can be deleted.

@cristianonicolai can you review / merge this, please.

Here's the link to the PR that causes the build problem: https://github.com/droolsjbpm/kie-wb-common/pull/414


csadilek:
I should add that the removed LuceneConfigProducer shouldn't ever have been used as it wasn't enabled as a CDI alternative, and the DefaultLuceneConfigProducer was always on the classpath.

cristianonicolai:
+1 thanks @csadilek

PR #468, opened 2016-08-31
RHBPMS-4199, JBPM-5301 - Different behavior of task service on EAP6 a…

mswiderski:
…nd EAP7

unfortunately couldn't figure out how to write a test for this as the only change is to use the main interface of the listener this producer will return rather than actual impl as it seems CDI impl in wildfly 10 interprets it differently.


cristianonicolai:
+1, jenkins failures are unrelated to this change.

PR #481, opened 2016-09-08
JBPM-4650 - New case list view

cristianonicolai:
New perspective and related modules for handling starting and closing/destroy cases.
Interaction with kie-server new API for case management,
Only adding it to jbpm-console-ng-showcase until feature complete.

Depends on: https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/272

How to test it:

- Start both jbpm-console-ng and latest kie-server (includes Case Management remote API)
- Author Evaluation project and update evaluation process as ad-hoc=true (Or any other project, as long as there are cases deployment to the current kie-server instance).
- Deploy Evaluation project into Kie-Server
- Navigate to Cases perspective where you will be able to start a new case ( based on Evaluation definition) as well as close or destroy any existing case instance.


cristianonicolai:
@krisv @mswiderski @nmirasch, initial commit for case instance list. Not expecting it to be final but only so that we can get going with the different UI building blocks for the case details.

nmirasch:
+1 very good starting point. Thanks!

PR #483, opened 2016-09-11
Improved error handling when registering queries on kie server when i…

mswiderski:
…t connects via controller

@cristianonicolai could you please have a look?


cristianonicolai:
@mswiderski change looks good but would be nice to see some test around this new logic.

mswiderski:
@cristianonicolai I refactored and added few tests, could you please have another look?

cristianonicolai:
looks good, thanks @mswiderski

PR #485, opened 2016-09-13
GUVNOR-2673 Default ExceptionHandler that pops up any uncaught RuntimeException

mcimbora:
There's now single uncaught exception handler that removes the need to define empty DefaultErrorCallback on callers https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/entrypoint/DefaultWorkbenchEntryPoint.java


csadilek:
+1

PR #486, opened 2016-09-14
[6.4.x] RHBPMS-4188 - Unable to use LATEST in process image REST endpoint

livthomas:
I have fixed the problem with LATEST deployment version resolution and tested it in Business Central using QE's test suite. I will create pull-requests for other branches as well.


krisv:
+1
https://github.com/droolsjbpm/jbpm-console-ng/commit/66610eb0048aee61e112a7b24368f47bc5315802

PR #490, opened 2016-09-22
RHBPMS-4246: Jobs: Queued tab displays jobs in status DONE

nmirasch:
Force to set the first tab at the end of restoring filter to avoid show results of other queries in tabs selected while restoring is being done.

### DIFF
--- a/jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-client/src/main/java/org/jbpm/console/ng/es/client/editors/requestlist/RequestListViewImpl.java
+++ b/jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-client/src/main/java/org/jbpm/console/ng/es/client/editors/requestlist/RequestListViewImpl.java
@@ -431,10 +431,8 @@ public class RequestListViewImpl extends AbstractMultiGridView<RequestSummary, R
         initTabFilter( preferences, REQUEST_LIST_PREFIX + "_6", constants.Cancelled(), constants.FilterCancelled(), statuses );
 
         filterPagedTable.addAddTableButton( createTabButton );
-        presenter.setAddingDefaultFilters( false );
-        getMultiGridPreferencesStore().setSelectedGrid( REQUEST_LIST_PREFIX + "_0" );
-        filterPagedTable.setSelectedTab();
-        applyFilterOnPresenter( REQUEST_LIST_PREFIX + "_0" );
+
+        selectFirstTabAndEnableQueries(  REQUEST_LIST_PREFIX + "_0" );
 
     }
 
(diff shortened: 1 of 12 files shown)


manstis:
No tests? If you move your deferred code to a method and package protect use of the scheduler you can mock the scheduler in a test and verify your method is called. Just a thought.

nmirasch:
@manstis  let me try add some test. Thanks!

manstis:
Sorry.. they don't look like test-friendly classes but the initDefaultFilters() methods should be testable with minimal changes.

jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-client/src/main/java/org/jbpm/console/ng/es/client/editors/requestlist/RequestListViewImpl.java
```diff
@@ -432,9 +433,13 @@ public void initDefaultFilters( GridGlobalPreferences preferences,
 
         filterPagedTable.addAddTableButton( createTabButton );
         presenter.setAddingDefaultFilters( false );
-        getMultiGridPreferencesStore().setSelectedGrid( REQUEST_LIST_PREFIX + "_0" );
-        filterPagedTable.setSelectedTab();
-        applyFilterOnPresenter( REQUEST_LIST_PREFIX + "_0" );
+        Scheduler.get().scheduleDeferred(new Scheduler.ScheduledCommand() {
+            @Override
+            public void execute() {
+                getMultiGridPreferencesStore().setSelectedGrid( REQUEST_LIST_PREFIX + "_0" );
```
cristianonicolai: Apart from the gridId the code is the same on every View class, so perhaps move the common part into the parent class to avoid duplication.

jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-client/src/main/java/org/jbpm/console/ng/es/client/editors/requestlist/RequestListViewImpl.java
```diff
@@ -432,9 +433,13 @@ public void initDefaultFilters( GridGlobalPreferences preferences,
...
+                filterPagedTable.setSelectedTab();
```
cristianonicolai: No need to call `applyFilterOnPresenter`?

nmirasch: The applyFilterOnPresenter was called twice. It's enough the setSelectedTab which has the filter call on the command

jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-client/src/main/java/org/jbpm/console/ng/es/client/editors/requestlist/RequestListViewImpl.java
nmirasch: Done.

nmirasch:
@cristianonicolai  I added some tests and moved the common part to AbstractMultiGridView. Can you take a look, Thanks!

cristianonicolai:
jenkins retest this

PR #499, opened 2016-10-04
JBPM-5355 - Support case project

mswiderski:
@cristianonicolai @nmirasch could you please have a look?

### DIFF
--- a/jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-backend/src/main/resources/CaseWorkDefinitions.wid
+++ b/jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-backend/src/main/resources/CaseWorkDefinitions.wid
@@ -1,0 +1,70 @@
+import org.drools.core.process.core.datatype.impl.type.StringDataType;
+import org.drools.core.process.core.datatype.impl.type.ObjectDataType;
+
+[
+  [
+    "name" : "Email",
+    "parameters" : [
+      "From" : new StringDataType(),
+      "To" : new StringDataType(),
+      "Subject" : new StringDataType(),
+      "Body" : new StringDataType()
+    ],
+    "displayName" : "Email",
+    "icon" : "defaultemailicon.gif"
+  ],
+
+  [
+    "name" : "Log",
+    "parameters" : [
+      "Message" : new StringDataType()
+    ],
+    "displayName" : "Log",
+    "icon" : "defaultlogicon.gif"
+  ],
+
+  [
+    "name" : "WebService",
+    "parameters" : [
+        "Url" : new StringDataType(),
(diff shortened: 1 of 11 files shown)


jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/handlers/NewCaseProjectHandler.java
```diff
@@ -0,0 +1,218 @@
...
+        this.notification = notification;
+    }
+
+    private org.uberfire.client.callbacks.Callback<Project> configureCaseProjectCallback = new org.uberfire.client.callbacks.Callback<Project>() {
+        @Override
+        public void callback(Project project) {
+            caseProjectService.call(new RemoteCallback<Void>() {
+                @Override
+                public void callback(Void aVoid) {
+                    notification.fire( new NotificationEvent( "Case project " + project.getProjectName() + " successfully configured", NotificationEvent.NotificationType.SUCCESS ) );
```
cristianonicolai: i18n please (below for the error message as well) :)

cristianonicolai:
@mswiderski overall looks good just some minor comments. Apart from that I'm just wondering if we should move this into jbpm-console-ng-workbench-integration module instead as the case-mgmt modules are unlikely to be deployed as part of the workbench at this point. They will be part of the case custom app instead.

mswiderski:
thanks @cristianonicolai will move it to workbench integration instead. Will add one more thing - to add custom work definitions like Milestone when project/package is created as well.

mswiderski:
@cristianonicolai all updates and improvements are now in place and ready for review

jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-backend/src/main/resources/CaseWorkDefinitions.wid
```diff
@@ -0,0 +1,70 @@
...
+    "icon" : "defaultservicenodeicon.png"
+  ],
+
+  [
+    "name" : "Milestone",
+    "parameters" : [
+      "Condition" : new StringDataType()
+    ],
+    "displayName" : "Milestone",
+    "icon" : "defaultemailicon.gif"
```
cristianonicolai: Perhaps defaultservicenodeicon.png would be more appropriate?

mswiderski: I actually used it first but this icon looks more appropriate even the name is saying something else :)

PR #501, opened 2016-10-06
JBPM-5349: Drilldown in Process and task dashboard broken

dgutierr:
Verified on a SDM + wf10/kie-server setup.

Tests have been also adapted to the new metric component.

@cristianonicolai Can you verify please?


dgutierr:
jenkins retest this

cristianonicolai:
@dgutierr when I select something on either _Processes By Type_ or _Tasks By Process_ all displayers get updated but the one I selected, is that expected behavior?

jhrcek:
@dgutierr I reported this behavior described by @cristianonicolai  as a regression in JBPM-5349, because in previous versions ALL the charts were updated including the one you click for filtering.

dgutierr:
@jhrcek @cristianonicolai  I will check that, fix it and add the corresponding test case as well to avoid regressions. Thanks mates!

dgutierr:
@cristianonicolai I fixed the issue and now everything works as expected.  Can you verify it again please?

cristianonicolai:
Thanks @dgutierr. Unfortunately, looks like the problem still there. Also the aggregate card colors seems to have blue background instead of white.
Here is what I get when selecting a specific process type:
selection_094

dgutierr:
@cristianonicolai weird. it looks like you were not running the latest changes I pushed.  @nmirasch Can you take a look and check if the dashboard looks fine in your env? Thanks!

nmirasch:
@cristianonicolai, @dgutierr this is how I see the screen in my environment (chrome and firefox), I think the colour of the boxes is right  
jira_5349

dgutierr:
Thanks a lot @nmirasch . That's exact the expected behaviour. @cristianonicolai Can you please double check your setup?

cristianonicolai:
thanks @nmirasch.
@dgutierr sorry, I can confirm that I'm getting the same as @nmirasch.
As you can see in the last image, _Processes By Type_, still not filtered when Evaluation process is selected. All other widgets are filtered but this one, same goes for _Tasks By Process_.

dgutierr:
@cristianonicolai  The  "Processes By Type" chart is acting as a selector. It should not be updated since it reflects the selection made by the user. This is the expected behaviour :-)

cristianonicolai:
@dgutierr @jhrcek, ok so I guess there is some misunderstanding here. I checked a previous version and the current behavior seems fine, whatever graph is used to drill down the information remains unchanged. @jhrcek, regarding JBPM-5349, this fixes part 1, but part 2 is just the expected behavior. WDYT?

jhrcek:
@cristianonicolai I verified the previous behavior with 6.4 product and it turns out I made a mistake: the problem described in 2 in the jira is actually expected behavior - only the first problem needed to be fixed. If that's the case this PR can be merge. Sorry for confusion.

cristianonicolai:
@jhrcek no worries, thanks for clarifying that.

PR #505, opened 2016-10-14
[JBPM-5366] replace kie-remote-common with kie-server-common

psiroky:
@mswiderski we need this one as well :)

I am wondering though if console-ng should depend on this artifact? I know its used to catch the particular exception, but I am wondering if it we should be it using here, since the exception seems low level?


mswiderski:
+1 to merge

looking at how it's actually used I think you're right, it could then be wrapped with KieServicesException

PR #507, opened 2016-10-17
Fix process definition details for using FormLabel instead of Label.

cristianonicolai:


cristianonicolai:
@nmirasch, could you please have a look, thanks.

nmirasch:
+1

PR #509, opened 2016-10-18
RHBPMS-4346: User with given role cannot see his human tasks in the t…

nmirasch:
…ask perspective
Added the user roles to task queries conditions


cristianonicolai:
@nmirasch overall it looks good, but I believe tests could be improved. Would be good if we could test that the actual role was used as a filter.

nmirasch:
@cristianonicolai test changed

cristianonicolai:
@nmirasch nice! thanks

PR #518, opened 2016-10-31
JBPM-5386 - Errai UI improvements

cristianonicolai:
Removed dependency to gwtbootstrap widgets


cristianonicolai:
@nmirasch please have a look :)

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/roles/CaseRolesViewImpl.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<div>
+<div data-field="roles">
```
csadilek: just a reminder that id= is sufficient to map to ```@DataField```, the data-field attribute should only be used in case the dom node is repeated and id would the be not unique.

PR #520, opened 2016-11-01
Checkstyle validation fix

cristianonicolai:
Add license header inline, allowing validation to succeed independently from where Maven is executed


psiroky:
I don't really like adding big chunks of CDATA sections into the POM files. That being said, I don't think there is a better way in this case.

Minor comment about the property name, otherwise looks good (we will likely need to move this in to kie-parent eventually).

cristianonicolai:
@psiroky agreed, this is not ideal solution. When we actually move this into kie-parent, ideally we should create a Maven module that would then store both, the checkstyle config as well as the template header file.

nmirasch:
+1

PR #522, opened 2016-11-02
Review & expand RemoteCaseManagementServiceImplTest

jhrcek:
* RemoteCaseManagementServiceImplTest - added test for case sorting, getCaseInstance and getCaseDefinition (for cases when remote client returns null), reused some string constants

* Exrtacted CaseCommentsMapper and  RoleAssignmentsMapper


cristianonicolai:
@jhrcek thanks!

PR #533, opened 2016-11-08
Improved test coverage for executor service mapper classes

cristianonicolai:


jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-api/src/main/java/org/jbpm/console/ng/es/model/RequestDetails.java
```diff
@@ -18,10 +18,12 @@
 
 import java.io.Serializable;
 import java.util.List;
-import java.util.Vector;
```
mswiderski: so you removed Vector, why? ;)

mswiderski:
+1 from my end, especially since I do work on some of the same classes I'd like to merge before sending PR :)

jbpm-console-ng-executor-service/jbpm-console-ng-executor-service-api/src/main/java/org/jbpm/console/ng/es/model/RequestDetails.java
cristianonicolai: I guess the question is why it was there in the first place :)
This class should basically be immutable as to represent the data retrieved from the Kie server, not really intended to manipulate its state.

PR #539, opened 2016-11-09
case-mgmt-backend: Add new/re-factor existing unit tests

bsremac:
Added new tests for Mapper classes and RemoteCaseManagementServiceImpl class and performed small re-factoring to align all test implementations.


cristianonicolai:
@bsremac thanks!

PR #553, opened 2016-11-16
Reuse AbstractPresenter in Case*Presenters

jhrcek:
Take advantage of existing AbstractPresenter class to initialize connection between view and presenter.

The change required making `AbstractCaseInstancePresenter` extend the generic `AbstractPresenter` and renaming the `viewXYZ` fields in the subclasses to `view` so it is inherited from `AbstractPresenter`. 

Also removed some duplicated TranslationService fields in the subclasses which are alreade declared as protected in `AbstractCaseInstancePresenter`.

UPDATE: also removed `@PostConstruct` methods for initializing view which are no longer necessary.


cristianonicolai:
@jhrcek thanks!

PR #556, opened 2016-11-18
Fix pagination limit exception

cristianonicolai:
Temporary fix to load max results until pagination is in place.


manstis:
@cristianonicolai Wasn't there PRs recently changing from ```Integer.MAX_VALUE``` to ```-1```?

cristianonicolai:
@manstis exactly, but apparently I missed something in my tests :/ 
Double checking the beta3 tag revealed an exception coming from the Kie Server if -1 is sent as page size.

PR #561, opened 2016-11-22
JBPM-5247: Case Stages

nmirasch:
@cristianonicolai first approach of case stages.

### DIFF
--- a/jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/console/ng/cm/backend/server/CaseInstanceMapperTest.java
+++ b/jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/console/ng/cm/backend/server/CaseInstanceMapperTest.java
@@ -16,11 +16,15 @@
 
 package org.jbpm.console.ng.cm.backend.server;
 
+import java.util.ArrayList;
 import java.util.Date;
+import java.util.List;
 
 import org.jbpm.console.ng.cm.model.CaseInstanceSummary;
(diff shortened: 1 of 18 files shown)


jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/console/ng/cm/backend/server/CaseInstanceMapperTest.java
```diff
@@ -36,19 +40,29 @@ public static void assertCaseInstance(final CaseInstance ci, final CaseInstanceS
...
         assertEquals(ci.getCompletedAt(), cis.getCompletedAt());
         assertEquals(ci.getCaseDefinitionId(), cis.getCaseDefinitionId());
+        assertCaseStages(ci.getStages(), cis.getStages());
+    }
+
+    public static void assertCaseStages(final List<CaseStage> csl, final List<CaseStageSummary> cssl) {
+        assertNotNull(cssl);
+        int i = 0;
+        CaseStage caseStage;
+        for (CaseStageSummary caseStageSummary : cssl) {
```
jhrcek: Mixing for-each loop + explicit index looks confusing. Why not use standard for loop for comparison like:

    assertTrue(csl.size() == cssl.size());
    for (int i = 0; i < csl.size(); i++) {
     CaseStage original = csl.get(i);
     CaseStageSummary copied = cssl.get(i);
     assertEquals(...);
    }
Also I'd add assert that verifies that both lists have the same size.

nmirasch:
@cristianonicolai @jhrcek done, can you take a look again? thanks!

jhrcek:
+1 Thanks for the changes :)

PR #562, opened 2016-11-22
case-mgmt-client: Case Comments - add unit tests

bsremac:
Added a couple of more tests for Case Comments and did a minor re-factoring around it.


cristianonicolai:
thanks @bsremac!

PR #571, opened 2016-11-28
JBPM-5241: Case Comments

nmirasch:
Comments refactoring.

### DIFF
--- a/jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
+++ b/jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
@@ -17,7 +17,11 @@
 package org.jbpm.console.ng.cm.client.comments;
 
 import javax.enterprise.context.Dependent;
+import javax.enterprise.event.Event;
+import javax.enterprise.event.Observes;
 import javax.inject.Inject;
+
+import com.google.gwt.user.client.TakesValue;
 
 import org.jboss.errai.common.client.api.IsElement;
 import org.jboss.errai.common.client.dom.Div;
@@ -26,40 +30,48 @@ import org.jboss.errai.common.client.dom.KeyboardEvent;
 import org.jboss.errai.common.client.dom.Span;
 import org.jboss.errai.common.client.dom.TextInput;
 import org.jboss.errai.common.client.dom.UnorderedList;
+import org.jboss.errai.databinding.client.api.DataBinder;
+import org.jboss.errai.security.shared.api.identity.User;
 import org.jboss.errai.ui.client.local.spi.TranslationService;
+import org.jboss.errai.ui.shared.api.annotations.AutoBound;
+import org.jboss.errai.ui.shared.api.annotations.Bound;
 import org.jboss.errai.ui.shared.api.annotations.DataField;
 
 import org.jboss.errai.ui.shared.api.annotations.EventHandler;
 import org.jboss.errai.ui.shared.api.annotations.ForEvent;
 import org.jboss.errai.ui.shared.api.annotations.Templated;
+
+import org.jbpm.console.ng.cm.client.events.CaseCommentEditEvent;
+import org.jbpm.console.ng.cm.client.util.AbstractView;
+import org.jbpm.console.ng.cm.client.util.DateConverter;
 import org.jbpm.console.ng.cm.client.util.FormGroup;
 import org.jbpm.console.ng.cm.client.util.ValidationState;
+import org.jbpm.console.ng.cm.model.CaseCommentSummary;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static org.jboss.errai.common.client.dom.DOMUtil.*;
 import static org.jboss.errai.common.client.dom.Window.*;
-import static org.jbpm.console.ng.cm.client.resources.i18n.Constants.CASE_COMMENT_CANT_BE_EMPTY;
+import static org.jbpm.console.ng.cm.client.resources.i18n.Constants.*;
 
 
 @Dependent
 @Templated
-public class CaseCommentItemView implements IsElement {
+public class CaseCommentItemView extends AbstractView<CaseCommentsPresenter> implements TakesValue<CaseCommentSummary>, IsElement {
 
     @Inject
     @DataField("comment-author")
-    Span commentAuthor;
+    @Bound
+    Span author;
 
     @Inject
     @DataField("comment-text")
-    Span commentText;
+    @Bound(property = "text")
+    Span text;
 
     @Inject
     @DataField("comment-addedat")
-    Span commentAddedAt;
-
-    @Inject
-    @DataField("user-actions")
-    Div userActions;
+    @Bound(converter = DateConverter.class)
+    Span addedAt;
 
     @Inject
     @DataField("list-group-item")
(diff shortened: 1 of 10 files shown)


jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
```diff
@@ -19,43 +19,55 @@
...
+    Span text;
 
     @Inject
     @DataField("comment-addedat")
-    Span commentAddedAt;
+    @Bound(converter = DateConverter.class)
+    Span addedAt;
 
     @Inject
     @DataField("user-actions")
```
jhrcek: userActions field is not used anywhere. Does it need to be injected.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
```diff
@@ -173,4 +195,67 @@ public void updateCommentPressingEnter(@ForEvent("keyup") final KeyboardEvent e)
...
+                public String label() {
+                    if (editItem) {
+                        return translationService.format(CANCEL);
+                    }
+                    return translationService.format(EDIT);
+                }
+
+                @Override
+                public void execute() {
+                   setEditMode(!editItem);
```
bsremac: With this new implementation it is possible to start editing multiple comments simultaneously, which per se doesn't have to be perceived as buggy behavior.
But when multiple comments are being edited and then one of the comments saved (using its Save option), the rest of currently edited comments are saved with old comment text. This might be perceived by user as loose behavior in comment editing functionality.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentsViewImpl.java
```diff
@@ -55,8 +58,17 @@
...
-    Div comments;
+    private ListComponent<CaseCommentSummary, CaseCommentItemView> comments;
+
+    @Inject
+    @AutoBound
+    private DataBinder<List<CaseCommentSummary>> caseCommentList;
+
+    @Inject
+    @DataField("empty-list-item")
+    private Div emptyContainer;
```
jhrcek: There are two fields injected in this class which are not used anywhere. I think they should be deleted not to clutter the class unnecessarity: commentCreation and newCommentLabel

nmirasch:
@jhrcek @bsremac  Thanks for review, I have done the changes you commented. I have added an event on the edit action that makes all the existing displayed items react, closing their edition mode.
@cristianonicolai @jhrcek @bsremac  could you have another look?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
```diff
@@ -173,4 +201,79 @@ public void updateCommentPressingEnter(@ForEvent("keyup") final KeyboardEvent e)
...
 
+    @Override
+    public CaseCommentSummary getValue() {
+        return caseCommentSummary.getModel();
+    }
+
+    @Override
+    public void setValue(final CaseCommentSummary model) {
+        this.caseCommentSummary.setModel(model);
+        this.updateCommentText.setValue(model.getText());
```
cristianonicolai: I guess this could be replaced by using `@Bound` annotation on updateCommentText.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
```diff
@@ -173,4 +201,79 @@ public void updateCommentPressingEnter(@ForEvent("keyup") final KeyboardEvent e)
...
+        setEditMode(editMode);
```
cristianonicolai: isn't edit mode always false when starting the ui?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-showcase/src/main/java/org/jbpm/console/ng/cm/server/MockCaseManagementService.java
```diff
@@ -138,12 +138,21 @@ public void destroyCaseInstance(final String serverTemplateId, final String cont
     @Override
     public void addComment(final String serverTemplateId, final String containerId, final String caseId, final String author, final String text) {
         final List<CaseCommentSummary> commentSummaryList = caseCommentMap.getOrDefault(caseId, new ArrayList<>());
-        final String id = String.valueOf(commentSummaryList.size() + 1);
+
+        int i= commentSummaryList.size() + 1;
```
cristianonicolai: I'm not sure why we need this, any particular issue with the previous version?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
```diff
@@ -173,4 +201,79 @@ public void updateCommentPressingEnter(@ForEvent("keyup") final KeyboardEvent e)
...
+                    }
+                }
+            });
+        }
+    }
+
+    public void onCaseCreatedEvent(@Observes CaseCommentEditEvent event) {
+        GWT.log("event.getCommentId():"+event.getCommentId() +" .getId() "+this.caseCommentSummary.getModel().getId());
+        if(event.getCommentId().equals(this.caseCommentSummary.getModel().getId())){
+            setEditMode( true);
```
cristianonicolai: code formatting :)

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
```diff
@@ -173,4 +201,79 @@ public void updateCommentPressingEnter(@ForEvent("keyup") final KeyboardEvent e)
...
+        if (identity.getIdentifier().equals(author)) {
+            addAction( new CaseCommentsPresenter.CaseCommentAction() {
+                @Override
+                public String label() {
+                    return translationService.format(DELETE);
+                }
+
+                @Override
+                public void execute() {
+                    presenter.deleteCaseComment(id);
```
cristianonicolai: Could we pass the object model in here (CaseCommentSummary) instead of just the String id.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
```diff
@@ -173,4 +201,79 @@ public void updateCommentPressingEnter(@ForEvent("keyup") final KeyboardEvent e)
...
+                addUpdateCommentAction(new CaseCommentsPresenter.CaseCommentAction() {
+                    @Override
+                    public String label() {
+                        return translationService.format(SAVE);
+                    }
+
+                    @Override
+                    public void execute() {
+                        if (validateForm()) {
+                            presenter.updateCaseComment(getUpdatedComment(),id);
```
cristianonicolai: Same here, pass the CaseCommentSummary object, which should already contain the new value for the comment in the text attribute, set via `@Bound`.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-showcase/src/main/java/org/jbpm/console/ng/cm/server/MockCaseManagementService.java
jhrcek: @cristianonicolai The problem is that the previous implementation could generate the same id for some comments. Conside this sequence of CRUD actions: Add, add, delete the first one, add -> there will be 2 comments with id 2.

@nmirasch I know this mock implementation is not that important, but I'd suggest using static int counter, which would be incremented every time when comment is added like so. This would make the intent to generate non-existent id cleaner:

    private static int commentIdGenerator = 0;

    public void addComment(..) {
        final String newId = String.valueOf(commentIdGenerator++);
        ...
    }
String.valueOf(commentSummaryList.size() + 1);

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/comments/CaseCommentItemView.java
jhrcek: Yes, Ctrl+Alt+L in Idea is my good friend :-)

nmirasch:
@cristianonicolai I keeped the updateCommentText without the '@bound' to maintain the original value that is needed to the cancel option. The same with the presenter.updateComment call, I pass the newText as parameter to avoid change the model before the presenter action is completed. preventing that something goes wrong there and show inconsistent values . Do this make sense? 
@jhrcek changes done. Added id generation you suggested, for sure it was a better solution.
@cristianonicolai @jhrcek would you mind taking another look? Thanks!

cristianonicolai:
Looks good, thanks @nmirasch!

PR #574, opened 2016-11-29
Review form related modules structure

cristianonicolai:
Move all form modules under a common parent project.


cristianonicolai:
@pefernan as we discussed.

pefernan:
@cristianonicolai looks good but I think we could be more optimistic and merge some of that form related projects together.

Knowing that:
- the jbpm-console-ng-generic-forms module has the basic API's to use forms on the console, 
- the jbpm-console-ng-process-runtime-forms & jbpm-console-ng-human-tasks-forms modules, are basically separating the process & task forms rendering API's (and include the rendering of the old FTL forms).

I think that these modules can  be merged together into a jbpm-console-ng-forms-(api/backend/client)

And then there are the modules the integrate the kie-wb-common forms engine & jbm-form-modeler ones. 

Since kie-wb-common-forms will be the default form engine, IMO it can be merged too into the new jbpm-console-ng-forms-* modules.

The ones from jbpm-form-modeler probably can live into a diferent module to make easy the maintenance & removal when required.

What do you think?

cristianonicolai:
@pefernan I think these are all great suggestions. I've myself also tried to reduce the number of maven modules to only a few necessary ones, if we can do that for forms it would be great!
If you don't mind, I would just prefer to do these merge in different steps/commits to break down the number of changes. WDYT?

pefernan:
@cristianonicolai totally agree.

BTW, I think that we should also have to review the current integration between tasklist & forms. Right now every action on tasklist / action buttons on the form displayer is forcing to refresh the form. That means requests to kie-server to get task / process info, forms, data... When the jbpm runtime was embedded on the workbench that wasn't a "big issue" but since it is on the kie-server it would be a good idea to change it.

cristianonicolai:
@pefernan are you happy for us to merge this as is? Also, could you create a JIRA for the issue you mentioned regarding the task list integration and forms?

pefernan:
@cristianonicolai +1

cristianonicolai:
@pefernan thanks!

PR #583, opened 2016-12-05
Case Roles: Re-factor existing and add new tests

bsremac:
Commit consists of: 
1. Re-factored existing Case Roles tests, mainly around test data initialization (so that parent test class is used more and common data shared among tests) and more efficient verification in some tests.
2. New tests aiming to cover in more details code branching at certain places in Presenter class

### DIFF
--- a/jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/pom.xml
+++ b/jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/pom.xml
@@ -114,6 +114,12 @@
       <scope>test</scope>
     </dependency>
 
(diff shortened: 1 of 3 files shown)


cristianonicolai:
@bsremac would you mind to squash these commits into a single one?

bsremac:
jenkins please retest this

bsremac:
@cristianonicolai I squashed the commits into one. Apologies for the mess around multiple redundant commits.
@jhrcek I added AssertJ library, only to this module, and used it in two tests which require multiple subsequent assertions. The assertions indeed look concise (and assertion errors are more informative)

cristianonicolai:
thanks @bsremac!

PR #589, opened 2017-01-10
JBPM-5244: Case Actions

nmirasch:
First approach of Actions BB.
This commit contains two different options, one for filtering actions and display them in a single list, and the other option with all action lists expanded. 
We are waiting here the definitive design of them, including the option of remove the first one depending of the available time. 
@cristianonicolai  can you take a look on this. thanks!


jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/console/ng/cm/backend/server/RemoteCaseManagementServiceImplTest.java
```diff
@@ -326,4 +332,193 @@ private CaseMilestone createTestMilestone(String caseMilestoneId, String caseMil
...
+
+
+        CaseActionSearchRequest defaultSearchRequest = new CaseActionSearchRequest();
+        defaultSearchRequest.setSort(sortStr);
+        defaultSearchRequest.setSortOrder(sortAsc);
+
+        List<CaseActionSummary> actionSummaries = testedService.getCaseActionsByStatus(caseId,defaultSearchRequest,userId,"Completed","Suspended","Failed",
+                "Error","Exited","Obsolete");
+
+        assertEquals(2,actionSummaries.size());
```
bsremac: I suggest adding one more TaskSummary instance which will not be taken into account in this test (for example with the status "Ready"). Then this assertion would be more valuable.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-api/src/main/java/org/jbpm/console/ng/cm/util/CaseActionSearchRequest.java
```diff
@@ -0,0 +1,77 @@
...
+@Bindable
+public class CaseActionSearchRequest {
+
+    private Integer status = 1;
+
+    private CaseActionsFilterBy filterBy = CaseActionsFilterBy.AVAILABLE;
+
+    private String sort = "createdOn";
+
+    private boolean sortOrder = true;
```
bsremac: Maybe renaming this filed to "sortAsc" (as you named the local variable in getCaseActionsByStatusTest() which contains the same value) would increase readability. 
"sortOrder" being set to "true" implies more that sorting is enabled in general and not that it just means ascending order.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/console/ng/cm/backend/server/RemoteCaseManagementServiceImplTest.java
```diff
@@ -326,4 +332,193 @@ private CaseMilestone createTestMilestone(String caseMilestoneId, String caseMil
...
+
+        final ArgumentCaptor<List> captor = ArgumentCaptor.forClass(List.class);
+        verify(clientMock,never()).getAdHocFragments(containerId,caseId);
+        verify(clientMock).findCaseTasksAssignedAsBusinessAdministrator(eq(caseId),eq(userId),captor.capture(),anyInt(),anyInt(),eq(sortStr),eq(sortAsc));
+        assertEquals("Completed", captor.getValue().get(0));
+        assertEquals("Suspended", captor.getValue().get(1));
+        assertEquals("Failed", captor.getValue().get(2));
+        assertEquals("Error", captor.getValue().get(3));
+        assertEquals("Exited", captor.getValue().get(4));
+        assertEquals("Obsolete", captor.getValue().get(5));
```
bsremac: I see that this particular sequence of assertions is been repeating itself in this test class.
If the order of statuses in the captor list is not of importance in tests and I think it isn't, I suggest making less verbose assertion.
For example, create a list with these status values and then reduce the assertions to only one, by calling:
assertTrue(captorList.containsAll(ListOfStatusValues));

nmirasch:
I'll take this comments and I'm going to close this pr and open another pr with a simplest vesion of BB, according with UX requirements. Letting only the version that show the expanded actions lists

PR #592, opened 2017-01-11
JBPM-5530 - Review case-mgmt related packages to use workbench instea…

cristianonicolai:
…d of console-ng


csadilek:
@cristianonicolai nice! all good to go once conflicts are resolved, imo.

PR #593, opened 2017-01-11
JBPM-5530 - Review showcase structure to use workbench instead of con…

cristianonicolai:
…sole-ng


csadilek:
@cristianonicolai nice! all good to go once conflicts are resolved, imo.

PR #595, opened 2017-01-11
JBPM-5246: Case Actions

nmirasch:
The Case Actions building block that show the three kind of actions at the same time.


jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/main/java/org/jbpm/console/ng/cm/backend/server/CaseActionAdhocFragmentMapper.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public CaseActionSummary apply(CaseAdHocFragment adHocFragment) {
+        if (adHocFragment == null) {
+            return null;
+        }
+
+        return CaseActionSummary.builder()
+                .name(adHocFragment.getName())
+                .status("AdhocFragment")
```
jhrcek: Is this hardcoded string going to be displayed anywhere? I don't see any value in setting it to hardcoded string. If status doesn't make sense for ad hoc fragments I'd rather see something line "Does not apply" or "N/A"

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/main/java/org/jbpm/console/ng/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +164,39 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+    public CaseActionsLists getCaseActionsLists(String containerId, String caseId, String userId) {
+        List<CaseActionSummary> availableActions = getCaseActionsByStatus(caseId, userId, "Ready");
+        availableActions.addAll(client.getAdHocFragments(containerId, caseId)
+                .stream()
+                .map(new CaseActionAdhocFragmentMapper())
+                .sorted(comparing(CaseActionSummary::getName)).collect(toList()));
+
+        CaseActionsLists caseActionsLists = new CaseActionsLists();
+        caseActionsLists.setAvailableActionList(availableActions);
+        caseActionsLists.setInprogressActionList(getCaseActionsByStatus(caseId, userId, "InProgress", "Reserved"));
```
jhrcek: I also have the feeling that case statuses are going to be used at more than one place throughout case mgmt app. Wouldn't it be better to represent them using Enum for better type safety and protection from typos?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/console/ng/cm/client/util/CaseActionStatus.java
```diff
@@ -0,0 +1,49 @@
...
+    public static CaseActionStatus fromStatus(final String status) {
+        for (CaseActionStatus status1 : CaseActionStatus.values()) {
+            if (status1.getStatus().contains(status)) {
+                return status1;
+            }
+        }
+        return null;
+    }
+
+    public List<String> getStatus() {
```
jhrcek: I believe the purpose of this class is to "collapse" task statuses into simpler set of case statuses. If that is the case we should make it more clear when we're talking about Case status and when about task status. In particular I'd rename 1) statuses to taskStatuses and 2) getStatus to getTaskStatuses because it returns a list of values, not just 1 value.
Also it wouldn't hurt to make statuses immutable list `this.statuses = Collections.unmodifiableList(Arrays.asList(statuses))`

nmirasch:
@jhrcek I have done the changes commented, can you take another look? thanks!

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionItemView.java
```diff
@@ -0,0 +1,163 @@
...
+    Div actions;
+    @Inject
+    @DataField("actions-items")
+    UnorderedList actionsItems;
+    @Inject
+    @AutoBound
+    private DataBinder<CaseActionSummary> caseActionSummary;
+    private Event<CaseStartAdhocFragmentEvent> startAdhocFragmentEvent;
+
+    @Override
```
bsremac: I don't believe you need to override this method again explicitly. The method is implemented in the parent class AbstractView.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionItemView.java
```diff
@@ -0,0 +1,163 @@
...
+        this.presenter = presenter;
+    }
+
+    @Override
+    public HTMLElement getElement() {
+        return listGroupItem;
+    }
+
+    @Inject
+    public void setCommentEditEvent(final Event<CaseStartAdhocFragmentEvent> startAdhocFragmentEvent) {
```
bsremac: Is this method name a typo?
Shouldn't name be something along the lines of "setStartAdhocFragmentEvent"?

nmirasch:
@bsremac  you are absolutly right!  Changes commented are done

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/NewActionViewImpl.java
```diff
@@ -0,0 +1,242 @@
...
+    @DataField("action-desc-input")
+    TextInput actionDescInput;
+
+    @Inject
+    @DataField("action-desc-help")
+    Span actionDescHelp;
+
+    @Inject
+    @DataField("role-name-group")
+    FormGroup roleNameGroup;
```
cristianonicolai: Seems like roles are not in use, not sure if makes sense in this context(creating dynamic task). Should roles be removed from this UI?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionListsPresenter.java
```diff
@@ -0,0 +1,171 @@
...
+        ).collect(Collectors.toSet());
+    }
+
+    protected void addDynamicAction(final String taskName, final String taskDescription, String actors, String groups) {
+        caseService.call((r) -> {
+            refreshData(caseId);
+        }).addDynamicUserTask(containerId, caseId, taskName, taskDescription, actors, groups, null);
+    }
+
+    protected void triggerAdhocFragment(final String actionName) {
```
cristianonicolai: AdHoc

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionListsPresenter.java
```diff
@@ -0,0 +1,171 @@
...
+                    }
+                    final Integer roleInstanceCardinality = cis.getRoleAssignments().stream().filter(ra -> role.equals(ra.getName())).findFirst().map(ra -> ra.getGroups().size() + ra.getUsers().size()).orElse(0);
+                    return roleInstanceCardinality < roleCardinality;
+                }
+        ).collect(Collectors.toSet());
+    }
+
+    protected void addDynamicAction(final String taskName, final String taskDescription, String actors, String groups) {
+        caseService.call((r) -> {
+            refreshData(caseId);
```
cristianonicolai: { and ; can be suppressed for one line lambdas.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionListsPresenter.java
```diff
@@ -0,0 +1,171 @@
...
+        }).addDynamicUserTask(containerId, caseId, taskName, taskDescription, actors, groups, null);
+    }
+
+    protected void triggerAdhocFragment(final String actionName) {
+        caseService.call((r) -> {
+                    refreshData(caseId);
+                }
+        ).triggerAdHocFragmentInStage(containerId, caseId, "", actionName, null);
```
cristianonicolai: Should the case current stage be used in here?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionListsViewImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public void updateListHeaders() {
+        availableActions.updateListHeader(AVAILABLE_ACTIONS);
+        inprogressActions.updateListHeader(INPROGRESS_ACTIONS);
+        completedActions.updateListHeader(COMPLETED_ACTIONS);
+    }
+
+    public void onCaseStartAdhocFragmentEvent(@Observes CaseStartAdhocFragmentEvent event) {
```
cristianonicolai: The Observer method could be moved directly into the presenter.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseSimpleActionListViewImpl.java
```diff
@@ -0,0 +1,123 @@
...
+            addCSSClass(emptyContainer, "hidden");
+        }
+    }
+
+    @Override
+    public void removeAllTasks() {
+        caseActionList.setModel(new ArrayList<>());
+    }
+
+    public void updateListHeader(String headerType) {
```
cristianonicolai: I guess instead of creating String constants and passing as parameter, you could create a specific method for each action type to handle the changes needed.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/NewActionViewImpl.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    @Override
+    public String getActors() {
+        return userNameInput.getValue();
+    }
+
+    @Override
+    public String getGroups() {
+        return null;
```
cristianonicolai: Should we add support so that new tasks are assigned to either users and/or groups?

nmirasch:
@cristianonicolai I did the changes you commented except the one related with the case current stage. Could you take another look on that. Thanks!

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-api/src/main/java/org/jbpm/workbench/cm/util/CaseActionStatus.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static CaseActionStatus fromStatus(final String status) {
+        for (CaseActionStatus status1 : CaseActionStatus.values()) {
+            if (status1.getTaskStatuses().contains(status)) {
+                return status1;
+            }
+        }
+        return null;
+    }
+
```
bsremac: How about implementing fromStatus method in the same way it was implemented in CaseStageStatus enum?
Something along the lines of:
return Arrays.stream(CaseActionStatus.values()).filter(s -> s.getTaskStatuses().contains(status)).findFirst().get();

It's much more readable and in case the status argument turns out to be a bogus one, it's better from debugging point of view to have call to get() throw NoSuchElementException, then having switch statement in CaseActionItemView throw NullPointerException.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImplTest.java
```diff
@@ -326,4 +329,52 @@ private CaseMilestone createTestMilestone(String caseMilestoneId, String caseMil
...
+    private void assertActionEquals(TaskSummary taskSummary, CaseActionSummary caseActionSummary) {
+        assertNotNull(taskSummary);
+        assertNotNull(caseActionSummary);
+        assertEquals(taskSummary.getId(), caseActionSummary.getId());
+        assertEquals(taskSummary.getName(), caseActionSummary.getName());
+        assertEquals(taskSummary.getStatus(), caseActionSummary.getStatus());
+    }
+
+    @Test
+    public void getCaseActionsTest() {
```
bsremac: This test doesn't test the processing of return value of client.getAdHocFragments(containerId, caseId), but rather just that it was called. i.e. the following segment is not tested:
availableActions.addAll(client.getAdHocFragments(containerId, caseId)
        .stream()
        .map(new CaseActionAdhocMapper())
        .sorted(comparing(CaseActionSummary::getName)).collect(toList()));

Maybe writing one more test which will also have some sample AdHoc Fragments or test with only AdHoc Fragments?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionsViewImpl.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public void init(final CaseActionsPresenter presenter) {
+        this.presenter = presenter;
+        availableActions.init(presenter);
+        inprogressActions.init(presenter);
+        completedActions.init(presenter);
+    }
+
+    @Override
+    public void setActionAddCommand(Command taskAddCommand) {
```
bsremac: How about renaming this method to setTaskAddCommand?
The reason is that "setAction" part in the method name implies a more general method which can be applied to any case action which can be added dynamically (like Sub-process for example).
Furthermore, I see also methods on the server side like addDynamicTask, so taking that into account I would go even further and rename this method to setUserTaskAddCommand and the Command instance to userTaskAddCommand.

The same naming ambiguity I see in CaseActionsPresenter, which has a method addDynamicAction that actually only handles dynamic user task.
At this point, this naming might be irrelevant but as code base grows with new dynamic actions being added on UI side, it may make code hard-to-read and understand.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionsListViewImp.java
```diff
@@ -0,0 +1,101 @@
...
+import org.jboss.errai.ui.shared.api.annotations.DataField;
+import org.jboss.errai.ui.shared.api.annotations.Templated;
+import org.jbpm.workbench.cm.client.util.AbstractView;
+import org.jbpm.workbench.cm.model.CaseActionSummary;
+
+import static org.jboss.errai.common.client.dom.DOMUtil.*;
+
+@Dependent
+@Templated(stylesheet = "CaseActionsListViewImp.css")
+public class CaseActionsListViewImp extends AbstractView<CaseActionsPresenter> implements CaseActionsPresenter.CaseActionsListView {
```
cristianonicolai: Impl, missing 'l'

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionsViewImpl.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public void setCompletedActionsList(List<CaseActionSummary> caseActionList) {
+
+        completedActions.setCaseActionList(caseActionList);
+    }
+
+    @Override
+    public void updateListHeaders() {
+        availableActions.updateActionsHeader(translationService.format(AVAILABLE_ACTIONS),
```
cristianonicolai: Seems like heatherText and tooltipTitle have always the same value, perhaps you could combine into a single parameter to avoid calling the translation service twice.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/NewActionViewImpl.java
```diff
@@ -0,0 +1,216 @@
...
+    private Modal modal;
+
+    private Command okCommand;
+
+    @Inject
+    private TranslationService translationService;
+
+    @PostConstruct
+    public void init() {
+
```
cristianonicolai: please review formatting , there are a few empty lines that could be removed across different files.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/NewActionViewImpl.java
cristianonicolai: groupsInput.getValue?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-showcase/src/main/java/org/jbpm/workbench/cm/server/MockCaseManagementService.java
```diff
@@ -106,6 +115,11 @@ public String startCaseInstance(final String serverTemplateId, final String cont
                 .stages(caseStageList)
                 .build();
         caseInstanceList.add(ci);
+
+        List actions = new ArrayList();
+        actions.addAll(caseActionList);
+        caseActionMap.putIfAbsent(ci.getCaseId(), actions);
```
cristianonicolai: you could simplify this by doing `new ArrayList<>(caseActionList)`

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +165,46 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+                .filter(c -> statusList.contains(c.getStatus())).collect(toList());
+    }
+
+    public void addDynamicUserTask(String containerId, String caseId, String name, String description, String actors, String groups, Map<String, Object> data) {
+        client.addDynamicUserTask(containerId, caseId, name, description, actors, groups, data);
+    }
+
+    @Override
+    public void triggerAdHocActionInStage(String containerId, String caseId, String stageId, String adHocName, Map<String, Object> data) {
+        client.triggerAdHocFragmentInStage(containerId, caseId, stageId, adHocName, data);
```
bsremac: It looks like stageId argument hasn't been used anywhere in the code, so I'm wondering why has this method been chosen over triggerAdHocFragment method (which doesn't consider case stages)?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImplTest.java
```diff
@@ -326,4 +331,78 @@ private CaseMilestone createTestMilestone(String caseMilestoneId, String caseMil
...
+        assertNotNull(caseAdHocFragment);
+        assertNotNull(caseActionSummary);
+        assertEquals(caseAdHocFragment.getType(), caseActionSummary.getDescription());
+        assertEquals(caseAdHocFragment.getName(), caseActionSummary.getName());
+        assertEquals("AD_HOC", caseActionSummary.getStatus());
+    }
+
+    @Test
+    public void getCaseActionsTest() {
+        String userId = "userId";
```
bsremac: Maybe replacing this local variable with a class field with the same name, so it's aligned with other variables used in the test?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImplTest.java
```diff
@@ -326,4 +331,78 @@ private CaseMilestone createTestMilestone(String caseMilestoneId, String caseMil
 
         return milestone;
     }
+
+    private TaskSummary createTaskSummary(Long taskId, String name, String status) {
```
bsremac: Maybe it makes sense to rename the method to createTestTaskSummary, so that this method name is aligned with other helper methods in this class which serve the same purpose.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImplTest.java
```diff
@@ -326,4 +331,78 @@ private CaseMilestone createTestMilestone(String caseMilestoneId, String caseMil
...
+                .id(taskId)
+                .name(name)
+                .status(status)
+                .createdOn(new Date())
+                .build();
+
+        return taskSummary;
+    }
+
+    private CaseAdHocFragment createAdHocFragment(String name, String type) {
```
bsremac: Maybe it makes sense to rename the method to createAdHocFragment, so that this method name is aligned with other helper methods in this class which serve the same purpose.

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImplTest.java
```diff
@@ -326,4 +331,78 @@ private CaseMilestone createTestMilestone(String caseMilestoneId, String caseMil
...
+
+    private void assertActionEquals(TaskSummary taskSummary, CaseActionSummary caseActionSummary) {
+        assertNotNull(taskSummary);
+        assertNotNull(caseActionSummary);
+        assertEquals(taskSummary.getId(), caseActionSummary.getId());
+        assertEquals(taskSummary.getName(), caseActionSummary.getName());
+        assertEquals(taskSummary.getStatus(), caseActionSummary.getStatus());
+    }
+
+    private void assertActionEqualsAdHocFragment(CaseAdHocFragment caseAdHocFragment, CaseActionSummary caseActionSummary) {
```
bsremac: This assertion method is practically the same as the one which was written in CaseActionAdHocMapperTest i.e. assertCaseActionAdHocFragment.
Can that method be somehow reused and avoid writing the new one with the same purpose?

jbpm-console-ng-case-mgmt/jbpm-console-ng-case-mgmt-backend/src/test/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImplTest.java
```diff
@@ -326,4 +331,78 @@ private CaseMilestone createTestMilestone(String caseMilestoneId, String caseMil
...
+    private CaseAdHocFragment createAdHocFragment(String name, String type) {
+        CaseAdHocFragment caseAdHocFragment = CaseAdHocFragment.builder()
+                .name(name)
+                .type(type)
+                .build();
+
+        return caseAdHocFragment;
+    }
+
+    private void assertActionEquals(TaskSummary taskSummary, CaseActionSummary caseActionSummary) {
```
bsremac: This assertion method is practically the same as the one which was written in CaseActionMapperTest i.e. assertCaseAction (only with less assertions being written inside the method).
Can that method be somehow reused and avoid writing the new one with the same purpose?

nmirasch:
@cristianonicolai I added here the stages  ad-hoc fragments loading and triggering. Can you take a look?

nmirasch:
@cristianonicolai added the active nodes to inProgress actions and startDynamicTasktoStage.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +168,101 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+                        .map(s -> {
+                            s.setStatus(CaseActionStatus.IN_PROGRESS.getTaskStatuses().get(0));
+                            return s;
+                        }).collect(toList());
+
+        if (!activeNoUserTaskActionsSummaries.isEmpty()) {
+            allActionsSummaries.addAll(activeNoUserTaskActionsSummaries);
+        }
+
+        List<TaskSummary> actionHumanTaskSummaries = client.findCaseTasksAssignedAsBusinessAdministrator(caseId, userId, allStatus, 0, PAGE_SIZE_UNLIMITED, "createdOn", false);
```
cristianonicolai: If a task that is assigned to the current user, I suspect that the same task could be duplicated here. It would also be in activenodes and returned in here, possibly?

nmirasch: @cristianonicolai I added the  filter:
.filter(c -> !Arrays.asList(CaseActionNodeInstanceMapper.NODE_TYPE_HUMAN_TASK_NODE,
                                CaseActionNodeInstanceMapper.NODE_TYPE_Dynamic_HUMAN_TASK)
 over the active nodes to avoid create "inprogess" actions over the human tasks than comes in activeNodes. That are retrieved by client.findCaseTasksAssignedAsBusinessAdministrator that only check the current user is designed a bussinessAdministrator.

cristianonicolai: got it :) for in_progress and complete, all we need is the info from the activenodes correct? If you filter out human tasks we only show tasks related to this user and not an overall status of the case. As for the available tasks, these should be only the ones that the user has permission to execute.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +168,101 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+        actions.setAvailableActions(availableActions);
+        actions.setInProgressAction(getCaseActionsByStatus(allActionsSummaries, CaseActionStatus.IN_PROGRESS));
+        actions.setCompleteActions(getCaseActionsByStatus(allActionsSummaries, CaseActionStatus.COMPLETED_FAIL, CaseActionStatus.COMPLETED_SUCCESS));
+        return actions;
+    }
+
+    public List<CaseActionSummary> getAllUserTaskActionSummaries(String containerId, String caseId, String userId) {
+        List<CaseActionSummary> allActionsSummaries = new ArrayList<>();
+
+        List<String> allStatus = new ArrayList<>();
```
cristianonicolai: Suggestion to move these lists to an static field as they never change.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +168,101 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+
+        List<NodeInstance> activeNodes = client.getActiveNodes(containerId, caseId, 0, PAGE_SIZE_UNLIMITED);
+        List<CaseActionSummary> activeNoUserTaskActionsSummaries =
+                activeNodes.stream()
+                        .filter(c -> !Arrays.asList(CaseActionNodeInstanceMapper.NODE_TYPE_HUMAN_TASK_NODE,
+                                CaseActionNodeInstanceMapper.NODE_TYPE_Dynamic_HUMAN_TASK)
+                                .contains(c.getNodeType()))
+                        .map(new CaseActionNodeInstanceMapper())
+                        .map(s -> {
+                            s.setStatus(CaseActionStatus.IN_PROGRESS.getTaskStatuses().get(0));
```
cristianonicolai: Perhaps this should be done at the CaseActionNodeInstanceMapper?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +168,101 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+        }
+
+        return allActionsSummaries;
+    }
+
+    public List<CaseActionSummary> getAdHocActions(String serverTemplateId, String containerId, String caseId) {
+        List<CaseActionSummary> adHocActions = client.getAdHocFragments(containerId, caseId)
+                .stream()
+                .map(new CaseActionAdHocMapper())
+                .sorted(comparing(CaseActionSummary::getName)).collect(toList());
```
cristianonicolai: sorting should be done at the end, after all items are added.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +168,101 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+        List<CaseActionSummary> caseAdHocActions = getCaseStagesAdHocActions(serverTemplateId, containerId, caseId);
+        if (caseAdHocActions.size() > 0) {
+            adHocActions.addAll(caseAdHocActions);
+        }
+        return adHocActions;
+    }
 
+    public List<CaseActionSummary> getCaseStagesAdHocActions(String serverTemplateId, String containerId, String caseId) {
+        CaseInstanceSummary caseInstanceSummary = getCaseInstance(serverTemplateId, containerId, caseId);
+        List<CaseActionSummary> adHocActionsActions = new ArrayList<>();
```
cristianonicolai: stages is never null, you could use streams here to simplify, something like:
`return caseInstanceSummary.getStages().stream().flatMap(s -> s.getAcHocActions().stream()).collect(toList());`

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/CaseActionNodeInstanceMapper.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.function.Function;
+
+import org.jbpm.workbench.cm.model.CaseActionSummary;
+import org.kie.server.api.model.instance.NodeInstance;
+
+public class CaseActionNodeInstanceMapper implements Function<NodeInstance, CaseActionSummary> {
+    public static String NODE_TYPE_Dynamic_HUMAN_TASK = "Human Task";
+    public static String NODE_TYPE_HUMAN_TASK_NODE = "HumanTaskNode";
+    //public static String NODE_TYPE_Dynamic_SERVICE_TASK =  "Service Task";
```
cristianonicolai: do we need to keep this?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
nmirasch: I'm not sure if make sense  sorting all the ad-hoc actions by the name, because the result can be show available actions mixed (the stages ad-hoc and the general ad-hoc). For the moment, I remove the sorting and, the available list will show the actions in this sequence: the 'available' human task, the case general ad-hoc (here I removed the sorting) and The case stages ad-hoc. Every subgroup will be shown as they come, without any other sorting

nmirasch:
@cristianonicolai  I did some changes here, can you review please? Thanks!!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -46,6 +54,14 @@
 public class RemoteCaseManagementServiceImpl implements CaseManagementService {
 
     public static final int PAGE_SIZE_UNLIMITED = Integer.MAX_VALUE;
+    public static List<String> allCaseActionStatus = new ArrayList<String>() {
```
cristianonicolai: This actually create an inner class for the list, my suggestion would be to use `Arrays.asList` to build this list.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +175,86 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+    }
+
+    public List<CaseActionSummary> getAllUserTaskActionSummaries(String containerId, String caseId, String userId) {
+        List<CaseActionSummary> allActionsSummaries = new ArrayList<>();
+
+
+        List<NodeInstance> activeNodes = client.getActiveNodes(containerId, caseId, 0, PAGE_SIZE_UNLIMITED);
+        List<CaseActionSummary> activeNoUserTaskActionsSummaries =
+                activeNodes.stream()
+                        .filter(c -> !Arrays.asList(CaseActionNodeInstanceMapper.NODE_TYPE_HUMAN_TASK_NODE,
```
cristianonicolai: These could also be moved into a static list...

nmirasch:
@cristianonicolai done

nmirasch:
@cristianonicolai  added the current stage concept to the ad-hoc available actions and the dynamic task creation.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +182,86 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+                        .filter(c -> !humanTaskActiveNodesTypes.contains(c.getNodeType()))
+                        .map(new CaseActionNodeInstanceMapper())
+                        .collect(toList());
+
+        if (!activeNoUserTaskActionsSummaries.isEmpty()) {
+            allActionsSummaries.addAll(activeNoUserTaskActionsSummaries);
+        }
+
+        List<TaskSummary> actionHumanTaskSummaries = client.findCaseTasksAssignedAsBusinessAdministrator(caseId, userId,
+                allHumanTaskCaseActionStatus, 0, PAGE_SIZE_UNLIMITED, "createdOn", false);
```
cristianonicolai: I guess we could search for human tasks only with Available status now.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +182,86 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+    }
+
+    public List<CaseActionSummary> getAllUserTaskActionSummaries(String containerId, String caseId, String userId) {
+        List<CaseActionSummary> allActionsSummaries = new ArrayList<>();
+
+
+        List<NodeInstance> activeNodes = client.getActiveNodes(containerId, caseId, 0, PAGE_SIZE_UNLIMITED);
+        List<CaseActionSummary> activeNoUserTaskActionsSummaries =
+                activeNodes.stream()
+                        .filter(c -> !humanTaskActiveNodesTypes.contains(c.getNodeType()))
```
cristianonicolai: Filter can be removed as we show human tasks from activeNodes.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/client/resources/i18n/Constants.properties
```diff
@@ -75,5 +75,33 @@ CaseMilestoneListViewImpl.NoMilestonesFound=No Milestones found
 Available=Available
 Terminated=Terminated
 Canceled=Canceled
+CaseStages=Case Stages
 CaseStagesViewImpl.Stages=Stages
-CaseStagesViewImpl.NoStagesFound=No Stages found
\ No newline at end of file
+CaseStagesViewImpl.NoStagesFound=No Stages found
+CaseActions=Case Actions
+CaseActionListViewImpl.Actions=Actions
```
cristianonicolai: CaseActionsListViewImpl, after renamae these needs update.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase/src/main/java/org/jbpm/workbench/cm/server/MockCaseManagementService.java
```diff
@@ -55,8 +59,10 @@
     private static final String CASE_DEFINITIONS_JSON = "case_definitions.json";
     private static final String CASE_MILESTONES_JSON = "case_milestones.json";
     private static final String CASE_STAGES_JSON = "case_stages.json";
+    private static final String CASE_ACTIONS_JSON = "case_actions.json";
 
     private static int commentIdGenerator = 0;
+    private static long actionIdLongenerator = 8;
```
cristianonicolai: This is just a mock class so its fine, but you could use AtomicInteger and AtomicLong which are thread safe.

nmirasch:
@cristianonicolai is pending here the conpleteNodes call. About the design  remains there the pagination over the actions lists. I just added the full .less files . The general structure with the styles will come in another PR.
Can you have a look? Thanks!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/CaseStageMapper.java
```diff
@@ -32,6 +36,14 @@ public CaseStageSummary apply(CaseStage stage) {
                 .name(stage.getName())
                 .identifier(stage.getIdentifier())
                 .status(stage.getStatus())
+                .adHocFragments(
+                        ofNullable(stage.getAdHocFragments()).orElse(emptyList())
+                                .stream()
+                                .map(new CaseActionAdHocMapper())
+                                .map(s -> {
+                                    s.setStageId(stage.getIdentifier());
```
cristianonicolai: you could pass the stage id in the `CaseActionAdHocMapper` constructor, that avoids calling map twice.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -46,10 +55,14 @@
 public class RemoteCaseManagementServiceImpl implements CaseManagementService {
 
     public static final int PAGE_SIZE_UNLIMITED = Integer.MAX_VALUE;
+    public static final String[] NODE_TYPE_HUMAN_TASK = new String[]{"Human Task", "HumanTaskNode"};
```
cristianonicolai: Instead of creating an array, you could create the list directly with ` Arrays.asList("Human Task", "HumanTaskNode")`. That way you can avoid creating the list later on.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +172,81 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+        actions.setAvailableActions(getAdHocActions(serverTemplateId, container, caseId));
+        actions.setInProgressAction(getInProgressActions(container, caseId));
+        actions.setCompleteActions(getCompletedActions(container, caseId));
+        return actions;
+    }
+
+    public List<CaseActionSummary> getInProgressActions(String containerId, String caseId) {
+        List<NodeInstance> activeNodes = client.getActiveNodes(containerId, caseId, 0, PAGE_SIZE_UNLIMITED);
+
+        return activeNodes.stream()
```
cristianonicolai: as we do a service call to kie-server, perhaps we could use `parallelStream()` in here.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +172,81 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+                .map(new CaseActionNodeInstanceMapper())
+                .map(s -> {
```
cristianonicolai: I would try to avoid calling `.map` twice. My suggestion would be to pass actualOwner and actionType as constructor parameters instead. That way you can also avoid having to map the workItemId. You would have access to it inside the map fragment.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +172,81 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+    public List<NodeInstance> getCaseCompletedNodes(String containerId, String caseId) {
+        //TODO return client.getCompletedNodes(containerId, caseId, 0, PAGE_SIZE_UNLIMITED);
+        return new ArrayList<>();
+
+    }
+
+    public List<CaseActionSummary> getCompletedActions(String containerId, String caseId) {
+        List<NodeInstance> activeNodes = getCaseCompletedNodes(containerId, caseId);
+        return activeNodes.stream()
+                .map(new CaseActionNodeInstanceMapper())
```
cristianonicolai: Suggest to pass actiontype as constructor parameter in the mapper.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -159,5 +172,81 @@ public void removeComment(final String serverTemplateId, final String containerI
...
+                .map(new CaseActionAdHocMapper())
+                .collect(toList());
+    }
+
+    public List<CaseActionSummary> getAdHocActions(String serverTemplateId, String containerId, String caseId) {
+        List<CaseActionSummary> adHocActions = getAdHocFragments(containerId, caseId);
+        CaseInstanceSummary caseInstanceSummary = getCaseInstance(serverTemplateId, containerId, caseId);
+        List<CaseActionSummary> caseAdHocActions = caseInstanceSummary.getStages().stream()
+                .filter(s -> s.getStatus().equals(CaseStageStatus.ACTIVE.getStatus()))
+                .flatMap(ah -> ah.getAdHocActions().stream()).collect(toList());
```
cristianonicolai: You can add the items directly to the final list, something like `.forEach(ah -> adHocActions.addAll(ah.getAdHocActions()));` That avoids the if statement and the extra caseAdHocActions collection.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionItemView.java
```diff
@@ -0,0 +1,158 @@
...
+    protected TranslationService translationService;
+
+    @Inject
+    @DataField("action-name")
+    @Bound
+    Span name;
+
+    @Inject
+    @DataField("action-actualOwner")
+    //@Bound
```
cristianonicolai: minor formatting issue and remove comment out code.

cristianonicolai:
@nmirasch great work! looks good to me. I made a few minor comments that I leave to you to review if makes sense changing or not.

nmirasch:
@cristianonicolai Thanks!  Done the changes you commented.

cristianonicolai:
@nmirasch thanks!

PR #600, opened 2017-01-13
JBPM-5530 - Review jbpm-wb-integration module

cristianonicolai:
Remove redundant -workbench from artifact name
Adjust folders to match new artifact id
Refactor package names to use workbench instead of console.ng


cristianonicolai:
Depends on https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/359

PR #613, opened 2017-01-24
JBPM-5454 - KIE Server router - support for advanced queries and work…

mswiderski:
…bench integration

depends on https://github.com/droolsjbpm/droolsjbpm-integration/pull/795

this is in general to be able to use KIE Server router with workbench for basic operations via UI. Overall changes are to switch to KIE Server client operations that include container id so router can efficiently find the right kie server instead of collecting and aggregating data for single instance operations like getProcessInstance or getTaskEvents.

### DIFF
--- a/jbpm-wb-dashboard/jbpm-wb-dashboard-client/src/main/java/org/jbpm/dashboard/renderer/client/panel/TaskDashboard.java
+++ b/jbpm-wb-dashboard/jbpm-wb-dashboard-client/src/main/java/org/jbpm/dashboard/renderer/client/panel/TaskDashboard.java
@@ -37,16 +37,17 @@ import org.dashbuilder.displayer.client.DisplayerListener;
 import org.dashbuilder.displayer.client.DisplayerLocator;
 import org.dashbuilder.renderer.client.metric.MetricDisplayer;
 import org.dashbuilder.renderer.client.table.TableDisplayer;
+
 import org.jboss.errai.common.client.api.Caller;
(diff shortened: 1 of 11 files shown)


jbpm-wb-dashboard/jbpm-wb-dashboard-client/src/main/java/org/jbpm/dashboard/renderer/client/panel/TaskDashboard.java
```diff
@@ -344,14 +344,15 @@ public void tableCellSelected(String columnId, int rowIndex) {
         }
 
         final Long taskId = Double.valueOf(ds.getValueAt(rowIndex, COLUMN_TASK_ID).toString()).longValue();
-        final String taskName = ds.getValueAt(rowIndex, COLUMN_TASK_NAME).toString();
+        final String taskName = ds.getValueAt(rowIndex, COLUMN_PROCESS_EXTERNAL_ID).toString();
+        final String deploymentId = ds.getValueAt(rowIndex, COLUMN_TASK_NAME).toString();
```
sutaakar: Aren't here switched COLUMN_TASK_NAME and COLUMN_PROCESS_EXTERNAL_ID? From the naming of variables I would guess so.

mswiderski: argh, yes, you're right, fixing directly

PR #618, opened 2017-01-31
Fix case overview full-height

cristianonicolai:


cristianonicolai:
@nmirasch not sure if we still need this, perhaps get it merged and apply your styles updates on top?

nmirasch:
@cristianonicolai I created a PR with the first step of the overview integration. Here we have to choose what styles we want to apply. I just push the styles provides as they came and the application in our code. Depending on the selected way, maybe doesn't make sense merge something that is going to disappear. What do you think?

cristianonicolai:
@nmirasch agreed, closing this PR ;)

PR #619, opened 2017-01-31
Remove i18n translation key view name prefix

cristianonicolai:
Avoid duplicates


manstis:
Hi @cristianonicolai I can understand removal of the prefix when you use ```TranslationService``` programmatically and use [keys](https://github.com/cristianonicolai/jbpm-wb/blob/ffcfa051513a2b7d8e3c67e76d0864f0f9973563/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/resources/i18n/Constants.java#L108) however does this work similarly for when ```errai-ui``` performs the substitution in HTML [files](https://github.com/cristianonicolai/jbpm-wb/blob/ffcfa051513a2b7d8e3c67e76d0864f0f9973563/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.html#L24)?

cristianonicolai:
@manstis good catch, this is something that was improved quite recently on [Errai](http://errai-blog.blogspot.com.au/2017/01/errai-400cr1-released.html). Now the i18n service defaults to a key without prefix if it can not find one with the view name.

PR #622, opened 2017-02-07
JBPM-5410 - Allow providing case owner to start a case instance

cristianonicolai:
By default, current user is set as owner

### DIFF
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
@@ -40,6 +40,7 @@ import org.jbpm.workbench.cm.util.CaseMilestoneSearchRequest;
 import org.jbpm.workbench.cm.util.CaseStageStatus;
 import org.kie.server.api.model.cases.CaseComment;
 import org.kie.server.api.model.cases.CaseDefinition;
+import org.kie.server.api.model.cases.CaseFile;
 import org.kie.server.api.model.cases.CaseInstance;
 import org.kie.server.api.model.cases.CaseMilestone;
 import org.kie.server.api.model.instance.NodeInstance;
@@ -56,6 +57,7 @@ import static java.util.Comparator.comparing;
 public class RemoteCaseManagementServiceImpl implements CaseManagementService {
 
     public static final int PAGE_SIZE_UNLIMITED = Integer.MAX_VALUE;
+    public static final String CASE_OWNER_ROLE = "owner";
     public static final List<String> NODE_TYPE_HUMAN_TASK = Arrays.asList("Human Task", "HumanTaskNode");
 
     @Inject
@@ -96,8 +98,9 @@ public class RemoteCaseManagementServiceImpl implements CaseManagementService {
     }
 
     @Override
-    public String startCaseInstance(final String serverTemplateId, final String containerId, final String caseDefinitionId) {
-        return client.startCase(containerId, caseDefinitionId);
+    public String startCaseInstance(final String serverTemplateId, final String containerId, final String caseDefinitionId, final String owner) {
+        final CaseFile caseFile = CaseFile.builder().addUserAssignments(CASE_OWNER_ROLE, owner).build();
+        return client.startCase(containerId, caseDefinitionId, caseFile);
     }
 
     @Override
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
@@ -24,6 +24,7 @@ import javax.inject.Inject;
 import org.jboss.errai.common.client.dom.HTMLElement;
 import org.jboss.errai.common.client.dom.MouseEvent;
 import org.jboss.errai.common.client.dom.Span;
+import org.jboss.errai.common.client.dom.TextInput;
 import org.jboss.errai.ui.client.local.spi.TranslationService;
 import org.jboss.errai.ui.shared.api.annotations.DataField;
 import org.jboss.errai.ui.shared.api.annotations.EventHandler;
@@ -37,6 +38,7 @@ import org.jbpm.workbench.cm.client.util.Select;
 import org.jbpm.workbench.cm.client.util.ValidationState;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
+import static org.jbpm.workbench.cm.client.resources.i18n.Constants.PLEASE_PROVIDE_CASE_OWNER;
 import static org.jbpm.workbench.cm.client.resources.i18n.Constants.PLEASE_SELECT_CASE_DEFINITION;
 
 @Dependent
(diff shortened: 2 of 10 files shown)


jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
```diff
@@ -99,21 +118,35 @@ public void cleanForm() {
...
-        } else {
+            valid = false;
+        }
+
+        if (isNullOrEmpty(ownerNameInput.getValue())) {
+            ownerNameInput.focus();
+            ownerNameHelp.setTextContent(translationService.format(PLEASE_PROVIDE_CASE_OWNER));
+            ownerNameGroup.setValidationState(ValidationState.ERROR);
+            valid = false;
+        }
```
bsremac: Form user experience point of view, maybe making case owner field mandatory for every case instance is too strong requirement.
My understanding is that case owners in most of the cases will be case workers, who will also usually start a case instance. So I think we should allow the flexibility of skipping owner field for those cases and making the user who stared the case instance the owner by default.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -96,8 +98,9 @@ public CaseDefinitionSummary getCaseDefinition(final String serverTemplateId, fi
     }
 
     @Override
-    public String startCaseInstance(final String serverTemplateId, final String containerId, final String caseDefinitionId) {
-        return client.startCase(containerId, caseDefinitionId);
+    public String startCaseInstance(final String serverTemplateId, final String containerId, final String caseDefinitionId, final String owner) {
+        final CaseFile caseFile = CaseFile.builder().addUserAssignments(CASE_ROLE_OWNER, owner).build();
```
bsremac: I think you accidentally switched places of arguments in addUserAssignments.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -56,6 +57,7 @@
 public class RemoteCaseManagementServiceImpl implements CaseManagementService {
 
     public static final int PAGE_SIZE_UNLIMITED = Integer.MAX_VALUE;
+    public static final String CASE_ROLE_OWNER = "owner";
```
bsremac: Since "owner" is a role, CASE_OWNER_ROLE makes more sense to me :)

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
cristianonicolai: Variables names were switched in the API, fixed in here: https://github.com/droolsjbpm/droolsjbpm-integration/pull/810

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
cristianonicolai: Agreed, by default the current user should be assigned as the case owner. That is in place here, the current user is defined as the value for the owner field. That way not only it still going to assign the new case instance to the current user but also allow that if there is need to assign to a different user, that could be done by simply providing a different user at the start case popup.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstancePresenterTest.java
```diff
@@ -71,7 +77,7 @@ public void init() {
 
     @Test
     public void testCreateInvalidCaseInstance() {
-        presenter.createCaseInstance(null);
+        presenter.createCaseInstance(null, null);
```
bsremac: How about putting anyString() call as the second argument instead of null, thus making it clear that the value of the second argument is irrelevant in this test?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
bsremac: I see it know.
Was mislead by ownerNameInput.setValue("") :)

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
cristianonicolai: Done

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstancePresenterTest.java
cristianonicolai: Done

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
```diff
@@ -122,14 +155,19 @@ public HTMLElement getElement() {
     }
 
     private void createCaseInstance() {
-        presenter.createCaseInstance(caseTemplatesList.getValue());
+        presenter.createCaseInstance(caseTemplatesList.getValue(), ownerNameInput.getValue());
     }
 
     private void clearErrorMessages() {
         definitionNameHelp.setTextContent("");
```
nmirasch: @cristianonicolai I think we have to clean here previous error messages related with the owner

cristianonicolai: Done :)

nmirasch: @cristianonicolai remains clearValidationState over the ownerNameGroup ;)

cristianonicolai: thanks @nmirasch, fixed now ;)

nmirasch:
+1 @cristianonicolai thanks!

PR #623, opened 2017-02-09
Use JSON format for communicating with Kie Server

cristianonicolai:


mswiderski:
@cristianonicolai what's the reason to change the format to JSON? The reason why I started with XSTREAM is that is the least fragile when it comes to custom types as it does not need to deal with additional configuration like classes to be added to client to properly deal with marshalling. Have you tried to use forms with custom types and JSON to see if that works properly?

cristianonicolai:
@mswiderski just really a preference in terms of JSON being more lightweight in the amount of data that needs to be transferred. Haven't tried that specific use case but if it is know to be problematic, I'm happy to leave as is.

mswiderski:
it not problematic but requires additional configuration so custom types are wrapped with type information while xstream does not need it so was the easiest to keep it simple and working for most of the cases.

cristianonicolai:
@mswiderski fair enough to me, thanks for the feedback ;)

PR #624, opened 2017-02-09
JBPM-5666: fix misleading warning when registering queries on kie-sever

jhrcek:


mswiderski:
good catch @jhrcek, thanks

PR #626, opened 2017-02-10
RHBPMS-4584: User can perform process instance specific operations wh…

nmirasch:
…ich was not permitted on BPMS
Check the current user has a granted Role, related with the featureId="wb_process_instances" before enable the button to access to the process instance list form task process context details.


cristianonicolai:
@nmirasch just a reminder to also push it into master if needed.

PR #627, opened 2017-02-13
RHBPMS-4583: User can perform process instance specific operations wh…

nmirasch:
…ich was not permitted on BPMS
Added the control of the permission to access to the process instances perspective to enable the link at task context details.

@cristianonicolai  can you take a look? Thanks

### DIFF
--- a/jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskprocesscontext/TaskProcessContextPresenter.java
+++ b/jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskprocesscontext/TaskProcessContextPresenter.java
@@ -25,6 +25,7 @@ import javax.inject.Inject;
 import com.google.gwt.user.client.ui.IsWidget;
 import org.jboss.errai.common.client.api.Caller;
 import org.jboss.errai.common.client.api.RemoteCallback;
+import org.jboss.errai.security.shared.api.identity.User;
 import org.jbpm.workbench.pr.model.ProcessInstanceKey;
 import org.jbpm.workbench.pr.model.ProcessInstanceSummary;
 import org.jbpm.workbench.ht.model.TaskSummary;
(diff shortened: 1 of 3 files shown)


jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskprocesscontext/TaskProcessContextPresenterTest.java
```diff
@@ -128,7 +137,7 @@ public void testGoToProcessInstanceDetails() {
 
         presenter.goToProcessInstanceDetails();
 
-        verify(placeManager).goTo("DataSet Process Instances With Variables");
+        verify(placeManager).goTo("DataSetProcessInstancesWithVariables");
```
cristianonicolai: Perhaps reuse constant here to avoid duplication.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskprocesscontext/TaskProcessContextPresenter.java
```diff
@@ -33,16 +34,20 @@
...
 import org.uberfire.mvp.impl.DefaultPlaceRequest;
+import org.uberfire.security.ResourceRef;
+import org.uberfire.security.authz.AuthorizationManager;
+import org.uberfire.workbench.model.ActivityResourceType;
 
 @Dependent
 public class TaskProcessContextPresenter {
 
-    public static final String PROCESS_INSTANCE_DETAILS = "DataSet Process Instances With Variables";
+    public static final String PROCESS_INSTANCE_DETAILS = "DataSetProcessInstancesWithVariables";
```
cristianonicolai: @nmirasch should this string match what is defined in the perspective? https://github.com/droolsjbpm/jbpm-wb/blob/master/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/perspectives/DataSetProcessInstancesWithVariablesPerspective.java#L39

Also, I can see some other places where this is also defined using the original value, see:
https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb/kie-wb-webapp/src/main/resources/security-module-user.properties#L20
https://github.com/droolsjbpm/kie-wb-distributions/blob/master/kie-wb/kie-wb-webapp/src/test/java/org/kie/workbench/security/WorkbenchACLTest.java#L104

nmirasch:
yes, you are absolutly right, I changed the perspective ids, thinking if maybe there was any problem with the spaces in the keys, viewing that the kie-wb didn't take this in count. After change kie-wb-common and kie-wb-distribution I found that was a problem in the permission definition in kie-wb. Now I sent a PR to fix that and leave our ids as they were. 
But may be a good moment to update this ids without white spaces like the other wb perspectives. Even it implies 3 projects  
@cristianonicolai what do you think? do it worth?

cristianonicolai:
@nmirasch I definitely agree that would be a good time to update our perspective ids. My suggestion would be to use the following:

- Tasks
- ProcessDefinitions
- ProcessInstances
- Requests

I would just avoid adding _Perspective_ as suffix because it gets added into the wb url when navigating. WDYT?

PR #629, opened 2017-02-16
JBPM perspectives id renaming

nmirasch:
@cristianonicolai  can you take a look?


nmirasch:
@cristianonicolai now this commits depends on https://github.com/droolsjbpm/kie-wb-common/issues/688 too because of the renaming of the PerspectiveIds

PR #634, opened 2017-02-21
JBPM-5246: Case Actions. DynamicSubprocess and pagination added

nmirasch:
Added Actions Pagination and add Dynamic Subprocess available action
@cristianonicolai can you take a look?


jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionsPresenter.java
```diff
@@ -96,10 +154,18 @@ protected void refreshData(final boolean refreshAvailableActions) {
         caseService.call((Actions actions) -> {
             if (refreshAvailableActions) {
                 List<CaseActionSummary> availableActions = new ArrayList<>();
+                //availableActions.add(CaseActionSummary.builder()
```
cristianonicolai: Could this be removed?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/NewActionViewImpl.java
```diff
@@ -144,39 +181,75 @@ private boolean validateForm() {
...
             validForm = false;
         }
-        if (isNullOrEmpty(usersInput.getValue()) && isNullOrEmpty(groupsInput.getValue())) {
-            usersInput.focus();
-            usersGroup.setValidationState(ValidationState.ERROR);
-            groupsGroup.setValidationState(ValidationState.ERROR);
-            assignationHelp.setTextContent(translationService.format(PLEASE_INTRO_USER_OR_GROUP_TO_ASSIGN_NEW_ACTION));
-            validForm = false;
+        switch (caseActionType) {
+            case ADD_DYNAMIC_TASK: {
```
cristianonicolai: If we don't support dynamic task via the UI I guess this could be removed.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/NewActionViewImpl.html
```diff
@@ -38,6 +38,23 @@ <h4 class="modal-title" data-i18n-key="NewAction"></h4>
...
+                        <div class="form-group" data-field="action-nodetype-group">
+                            <label class="control-label col-md-3" data-i18n-key="NodeType"></label>
+                            <div class="col-md-9">
+                                <input type="text" data-field="action-nodetype-input" class="form-control"/>
+                                <span data-field="nodetype-help" class="help-block"></span>
+                            </div>
+                        </div>
+
+                        <div class="form-group" data-field="action-case-definitions-group">
+                            <label class="control-label col-md-3" data-i18n-key="ProcessId"></label>
```
cristianonicolai: Please add a required indicator to this field. Also perhaps name it as _Process Name_ instead.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionsPresenter.java
```diff
@@ -64,26 +70,78 @@ protected void clearCaseInstance() {
...
+                (List<CaseDefinitionSummary> definitions) -> {
+                    final List<String> definitionNames = new ArrayList<>();
+                    for (CaseDefinitionSummary summary : definitions) {
+                        definitionNames.add(summary.getName());
+                        caseDefinitions.put(summary.getName(), summary);
+                    }
+                    Collections.sort(definitionNames);
+                    newActionView.setCaseDefinitions(definitionNames);
+                }
+        ).getCaseDefinitions();
```
cristianonicolai: I think this call is incorrect, we should get all the processes that are not cases, which is returned by _CaseServicesClient.findProcessesByContainerId_.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionsListViewImpl.java
```diff
@@ -38,6 +42,7 @@
 @Dependent
 @Templated
 public class CaseActionsListViewImpl extends AbstractView<CaseActionsPresenter> implements CaseActionsPresenter.CaseActionsListView {
+    private int PAGE_SIZE = 4;
```
cristianonicolai: Something tricky here, the Available actions column should be taller to match In Progress and Completed columns. One way to fix it could be adding some extra info for each line in column, that could be the stage as @bsremac suggested. I can think of three status here:
- Dynamic (For user task and process task)
- Stage name (For ad-hoc tasks from a specific case stage)
- Case level (For ad-hoc tasks from the case instance level)
wdyt?

bsremac: This is definitely a good place to add clarification for the case worker about where the Ad-hoc actions are located in the model.
The same three statuses come to my mind too.
My two pennyworth when it comes to status names:
- Dynamic (would also drop "Add new dynamic" from the user/process task names)
- Available in: Stage name
- Available in: Case

nmirasch:
Removed all referred to generic Dynamic task and changed the service to fill the available subprocess in the dynamic suprocess addition.
@cristianonicolai Could you take another look? thanks!!!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/actions/CaseActionsListViewImpl.java
nmirasch: @bsremac Sorry I saw this comment after updating the pr. 
I'm going to do those changes and refresh the PR. Thanks!

nmirasch:
@cristianonicolai @bsremac changes done. Ready for review :)

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-api/src/main/java/org/jbpm/workbench/cm/service/CaseManagementService.java
```diff
@@ -69,8 +70,18 @@
 
     void addDynamicUserTaskToStage(String containerId, String caseId, String stageId, String name, String description, String actors, String groups, Map<String, Object> data);
 
+    void addDynamicTask(String containerId, String caseId, String nodeType, String name, Map<String, Object> data);
+
+    void addDynamicTaskToStage(String containerId, String caseId, String stageId, String nodeType, String name, Map<String, Object> data);
```
bsremac: Just for consistency reasons, do we want to keep methods which handle dynamic service tasks in the interface while completely removing dynamic service task from CaseActionType enum?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-api/src/main/java/org/jbpm/workbench/cm/model/ProcessDefinitionSummary.java
```diff
@@ -0,0 +1,153 @@
...
+
+import java.util.Objects;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.jboss.errai.databinding.client.api.Bindable;
+
+
+@Bindable
+@Portable
+public class ProcessDefinitionSummary implements Comparable<ProcessDefinitionSummary> {
```
bsremac: I don't see where in the code the implementation of Comparable interface pays off.

nmirasch:
@cristianonicolai @bsremac  removed the addDynamicTask and the comparator.

PR #638, opened 2017-02-23
JBPM-4647: Case Roles. Pagination added and partial UX design applied

nmirasch:
Added a generic treatment of pagination to be used by building blocs, applied to actions and Roles.
Joined the user and group role assignment in one action. Includes some UX proposed changes.
@cristianonicolai could you take a look? Thanks


cristianonicolai:
@nmirasch looks good, thanks!

PR #639, opened 2017-02-24
JBPM-5714 - Provision Case Management showcase app when workbench is …

cristianonicolai:
…deployed

### DIFF
--- a/jbpm-wb-integration/jbpm-wb-integration-backend/pom.xml
+++ b/jbpm-wb-integration/jbpm-wb-integration-backend/pom.xml
@@ -30,6 +30,10 @@
 
   <name>jBPM Workbench - Integration Backend</name>
 
+  <properties>
+    <jboss.home>${project.build.directory}/wildfly-${version.org.wildfly.gwt.sdm}</jboss.home>
+  </properties>
+
   <dependencies>
 
     <dependency>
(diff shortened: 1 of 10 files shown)


cristianonicolai:
Depends on:
- https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/396
- https://github.com/droolsjbpm/guvnor/pull/418

cristianonicolai:
Instructions for testing: 

Start a Wildfly instance and deploy kie-wb.war with the following system property set:
_org.jbpm.casemgmt.showcase.deploy=true_

By default it will try to use the following parameters:
- Widfly port: 8080 ( system property to override: org.jbpm.casemgmt.showcase.wildfly.port )
- Widfly management port: 9990 ( system property to override: org.jbpm.casemgmt.showcase.wildfly.management-port )
- Widlfly user: admin ( system property to override: org.jbpm.casemgmt.showcase.wildfly.username )
- Widlfly user password: admin ( system property to override: org.jbpm.casemgmt.showcase.wildfly.password )
- WilldFly host: localhost (  system property to override:  org.jbpm.casemgmt.showcase.wildfly.host )

Once the provisioning service is started, it will use Maven dependency resolution to resolve the jbpm-wb-case-mgmt-showcase:war:7.0.0-SNAPSHOT dependency which will get deployed into Wildfly.

If provisioning is successful, the showcase app will be available on http://localhost:8080/jbpm-cm.

nmirasch:
+1

cristianonicolai:
@bsremac as I mentioned, there is also an alternative method to get the final showcase app war file which points to a path in the file system. For checking that, you should use the following system property: _org.jbpm.casemgmt.showcase.path_. By setting this property, the maven resolution will be turned off for the case-mgmt-showcase.war file.

PR #641, opened 2017-02-26
Case Actions: Separate status from type in model

bsremac:
Case Action type can be: 
- AD_HOC_TASK,
- DYNAMIC_USER_TASK
- DYNAMIC_SUBPROCESS_TASK

Case Action status can be:
- AVAILABLE,
- IN_PROGRESS
- COMPLETED

The idea behind status AVAILABLE is quite simple.
All dynamic actions will have status AVAILABLE throughout the entire case life-cycle. 
The same goes for ALL ad-hoc tasks regardless of their scope (case level or stage level). This treatment of ad-hoc tasks is safe because when displaying available ad-hoc actions which are part of a stage, the stage status will determine whether its ad-hoc actions will be exposed to the case worker or not (i.e. only if the stage status is Available, its ad-hoc actions will be offered).

Have also changed the mapping of Case Action "id" property in CaseActionNodeInstanceMapper, so it's mapped to Node Instance "id" instead of "work-item-id". This change doesn't influence the implemented use of work-item-id in retrieving "In progress/Completed" node instances.

In addition, have done a smaller re-factoring around structuring the code, removing unused pieces and enhancing overall readability.

### DIFF
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-api/src/main/java/org/jbpm/workbench/cm/model/CaseActionSummary.java
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-api/src/main/java/org/jbpm/workbench/cm/model/CaseActionSummary.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 
 import org.jboss.errai.common.client.api.annotations.Portable;
 import org.jboss.errai.databinding.client.api.Bindable;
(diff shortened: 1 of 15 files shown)


cristianonicolai:
looks good to me. @nmirasch as you done most of this code, could you also have a look?

nmirasch:
+1
@bsremac nice changes.

PR #642, opened 2017-03-02
JBPM-5714 - Handle adding new app launcher once Case Management Showc…

cristianonicolai:
…ase is provisioned

Add provisioning lifecycle events
Add provisioning status
Rename package cases to casemgmt to match jBPM core API

### DIFF
--- a/jbpm-wb-integration/jbpm-wb-integration-backend/src/main/java/org/jbpm/workbench/wi/backend/server/casemgmt/service/CaseProvisioningExecutor.java
+++ b/jbpm-wb-integration/jbpm-wb-integration-backend/src/main/java/org/jbpm/workbench/wi/backend/server/casemgmt/service/CaseProvisioningExecutor.java
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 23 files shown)


cristianonicolai:
Depends on https://github.com/droolsjbpm/kie-wb-common/pull/703

cristianonicolai:
@manstis mind checking this one as well.

jbpm-wb-integration/jbpm-wb-integration-backend/src/main/java/org/jbpm/workbench/wi/backend/server/casemgmt/service/CaseProvisioningExecutor.java
```diff
@@ -61,6 +79,12 @@ protected int getManagementInterfaceStatus(String host, String managementPort, S
         }
     }
 
+    private interface ProvisioningCommand {
```
manstis: Do we really need another ```Command``` interface? There are quite a few around already.

cristianonicolai:
@mantis I duplicated the command interface to avoid pulling uberfire
dependencies into the integration test.

PR #646, opened 2017-03-07
GUVNOR-2916: Library: It should be possible to create other project types on the library.

paulovmr:


paulovmr:
Part of ensemble:

https://github.com/droolsjbpm/kie-wb-common/pull/712
https://github.com/droolsjbpm/jbpm-wb/pull/646

PR #647, opened 2017-03-08
JBPM-5714 - Showcase fixes

cristianonicolai:
Add security policy definition
Moved .less files to public folder
Renamed perspectives
Included Errai generated classes into final war file, via errai.server.classOutput system property.


cristianonicolai:
@bsremac the missing security policy file was the reason for the empty screen after login, should be fixed now.

bsremac:
@cristianonicolai I've built the _jbpm-wb-case-mgmt-showcase.war_ with the changes contained in this PR and used _org.jbpm.casemgmt.showcase.path_ property to point at it during provisioning - I still see the same blank page after logging in.

cristianonicolai:
Depends on https://github.com/uberfire/uberfire/pull/641

cristianonicolai:
@bsremac thanks for verifying that, after investigating it further, the real issue is that both apps workbench and case app were sharing the same .niogit folder to store its Uberfire related settings. Pushing a fix to Uberfire so that each one can have independent folders.

cristianonicolai:
@nmirasch @bsremac when you guys have a chance, please review.

bsremac:
@cristianonicolai I am able to access _CaseList_ after provisioning, but I couldn't verify access to _CaseOverview_ because I can't start a case instance through _New Case Instance_ dialog (after clicking Start button nothing happens).
+1 as I think the problem with starting a case instance is unrelated to the scope of this PR.

nmirasch:
+1

PR #650, opened 2017-03-14
JBPM-5241: Case Roles Redesign

nmirasch:
Changed the roles assignment bb UI design.
Please @cristianonicolai  could you take a look? thanks!


jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -48,42 +56,150 @@ public String getTittle() {
...
-        view.addUser(cis.getOwner(), translationService.format(OWNER));
-
-        setupRoleAssignments(cis);
-
-        setupNewRoleAssignments(cis);
-
-        view.setupPagination();
+        if (passFilter(Arrays.asList(cis.getOwner()), null)) {
+            view.addOwnerAssignment(translationService.format(OWNER), cis.getOwner());
+        }
```
bsremac: Not sure the evaluation of cis.getOwner() has a value here. The case instance can't be created without providing the owner, which means that if condition will always evaluate to _true_.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
```diff
@@ -84,35 +104,43 @@ public void removeAllRoles() {
...
     }
 
-    @Override
-    public void addGroup(final String groupName, final String roleName, final CaseRolesPresenter.CaseRoleAction... actions) {
-        addRoleView(groupName, roleName, "pficon-users", actions);
+    public String getFilterValue() {
+        if (isNullOrEmpty(filterSelect.getValue())) {
+            return "All";
+        }
+        return filterSelect.getValue();
```
bsremac: Maybe would make more sense to move the logic of processing _filterSelect_ value to presenter and leave just a simple getter method in the view.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
```diff
@@ -129,23 +157,21 @@ public void setupPagination() {
...
     }
 
-    @EventHandler("user-add")
-    public void onUserAddClick(@ForEvent("click") Event e) {
-        if (userAddCommand != null) {
-            userAddCommand.execute();
-        }
+    @EventHandler("filter-select")
+    public void onRolesAssignmentFilterChange(@ForEvent("change") Event e) {
+        filterCommand.execute();
```
bsremac: Does @ForEvent("change") means that each time the filter is changed between the options: _All_, _Unassigned_ and _Assigned_, the method is executed?
If yes, that means the every filter change makes a new request to kie server through findCaseInstance() and re-loads all the assignments all over again. Is this necessary?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -133,9 +134,39 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
     @Override
-    public void assignGroupAndUserToRole(final String serverTemplateId, final String containerId, final String caseId, final String roleName, final String user, final String group) {
-        client.assignGroupToRole(containerId, caseId, roleName, group);
-        client.assignUserToRole(containerId, caseId, roleName, user);
+    public void saveAssignment(final String serverTemplateId, final String containerId, final String caseId,
+                               final String roleName, final String users, final String groups) {
+        CaseRoleAssignmentSummary ras = getCaseInstance(serverTemplateId, containerId, caseId).getRoleAssignments()
+                .stream()
+                .filter(ra -> roleName.equals(ra.getName()))
+                .findFirst().orElse(null);
```
triceo: The null can be avoided. `Optional.ifPresent()` will let you do what you need to do.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -133,9 +134,39 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
+        usersList.forEach(u -> assignUserToRole(serverTemplateId, containerId, caseId, roleName, u));
+        groupsList.forEach(g -> assignGroupToRole(serverTemplateId, containerId, caseId, roleName, g));
+
+    }
+
+    private List<String> getListFromString(String stringList) {
+        if (stringList != null && stringList.trim().length() > 1) {
+            return Arrays.asList(stringList.split(",")).stream().map(s -> s.trim()).collect(toList());
+        } else {
+            return new ArrayList<String>();
```
triceo: Unless you later need to add items to the list, `Collections.emptyList()` will work better in terms of both memory use and performance.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
```diff
@@ -67,14 +74,27 @@
...
     public void init(final CaseRolesPresenter presenter) {
+    }
+
+    @PostConstruct
+    public void init() {
+        stream(CaseRolesAssignmentFilterBy.values())
+                .collect(toMap(s -> s.getLabel(),
+                        s -> translationService.format(s.getLabel())))
+                .entrySet()
+                .forEach(s -> filterSelect.addOption(s.getValue(), s.getKey()));
```
triceo: The intermediate map does not seem to bring value. Why not do the `forEach()` directly on the first stream, saving the cycles and memory it costs to create the map?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-api/src/main/java/org/jbpm/workbench/cm/service/CaseManagementService.java
```diff
@@ -58,7 +58,8 @@
 
     void assignGroupToRole(String serverTemplateId, String containerId, String caseId, String roleName, String group);
 
-    void assignGroupAndUserToRole(final String serverTemplateId, final String containerId, final String caseId, final String roleName, final String user, final String group);
+    void saveAssignment(final String serverTemplateId, final String containerId, final String caseId,
```
cristianonicolai: perhaps _assingToRole_?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -133,9 +134,39 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
-    public void assignGroupAndUserToRole(final String serverTemplateId, final String containerId, final String caseId, final String roleName, final String user, final String group) {
-        client.assignGroupToRole(containerId, caseId, roleName, group);
-        client.assignUserToRole(containerId, caseId, roleName, user);
+    public void saveAssignment(final String serverTemplateId, final String containerId, final String caseId,
+                               final String roleName, final String users, final String groups) {
+        CaseRoleAssignmentSummary ras = getCaseInstance(serverTemplateId, containerId, caseId).getRoleAssignments()
+                .stream()
+                .filter(ra -> roleName.equals(ra.getName()))
+                .findFirst().orElse(null);
+        List<String> usersList = getListFromString(users);
```
cristianonicolai: Removing and adding all assignments seems unnecessary, could we just add/remove what was actually changed?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/client/resources/i18n/Constants.properties
```diff
@@ -21,6 +21,10 @@ CaseDetails=Case Details
 Owner=Owner
 Roles=Roles
 RoleAssignments=Role Assignments
+DeleteAssignment=Delete assignment
```
cristianonicolai: I think this can be discussed but perhaps we should use _Remove_ instead of _Delete_.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/client/resources/i18n/Constants.properties
```diff
@@ -21,6 +21,10 @@ CaseDetails=Case Details
...
+DeleteTheAssignmentForRole=Delete the assignment for role: {0}?
```
cristianonicolai: Remove all users and groups assignments from role {0}?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/client/resources/i18n/Constants.properties
```diff
@@ -21,6 +21,10 @@ CaseDetails=Case Details
...
+DeleteUserAssignmentForRole=Delete user {0} assignment for role: {1}?
```
cristianonicolai: Remove user {0} assignment from role {1}?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/client/resources/i18n/Constants.properties
```diff
@@ -21,6 +21,10 @@ CaseDetails=Case Details
...
+DeleteGroupAssignmentForRole=Delete group {0} assignment for role: {1}?
```
cristianonicolai: Remove group {0} assignment from role {1}?

cristianonicolai:
@nmirasch a few notes in here:

- The Delete option in the kebab, should only be visible if there is at least one assignment for the role.
- The text style for users, groups and unassigned looks strange, I believe that is because this text lives inside a div with a _pficon_
- Instead of Delete, we should probably call it _Remove all assignments_?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
nmirasch: @bsremac not sure if I understand correctly,  this condition passFilter(Arrays.asList(cis.getOwner()), null)) return false when the requested assignments are the Unassigned ones.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
nmirasch: @bsremac  this event is fired when the filter select value changes, With the current implementation, yes, the list of roles assignment is requested again. Let me discuss with @cristianonicolai if we can store the values in any array at the presenter and avoid to reload it every time.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
bsremac: @nmirasch Yes, I understand now. This comment is actually related to that other comment I made about sending requests to kie server every time filtering is changed.
When I made this comment I didn't expect that loadCaseInstance() method will be used repeatedly, but rather only once i.e. when the case is loaded. In that case, this evaluation wouldn't make sense.
It would be good if it would be possible to somehow decouple loading of case roles from handling their lists in various filter tabs, as I think all the data you need can be saved are re-used without calling this method repeatedly.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
cristianonicolai: that should be fine, I agree that we shouldn't need to go back to the server to filter the data.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
```diff
@@ -84,35 +104,43 @@ public void removeAllRoles() {
...
+        if (isNullOrEmpty(filterSelect.getValue())) {
+            return "All";
+        }
+        return filterSelect.getValue();
     }
 
-    private void  addRoleView(final String name, final String roleName, final String iconType, final CaseRolesPresenter.CaseRoleAction... actions) {
+    public void addAssignment(String roleName, List<CaseRolesPresenter.CaseAssignmentItem> users, List<CaseRolesPresenter.CaseAssignmentItem> groups,
+                              CaseRolesPresenter.CaseRoleAction... actions) {
         final CaseRoleItemView roleItemView = provider.get();
```
cristianonicolai: Would it be possible to use a ListComponent in here instead of manually setting the items view?

nmirasch:
depends on https://github.com/kiegroup/jbpm-wb/pull/651 Once it is merged,I have to rebase this PR.
@cristianonicolai @bsremac could you  take a look please?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/util/AbstractView.java
```diff
@@ -31,4 +31,7 @@ protected native void tooltip(final HTMLElement e) /*-{
         $wnd.jQuery(e).tooltip();
     }-*/;
 
+    protected native void popover(final HTMLElement e) /*-{
```
cristianonicolai: I end up creating a component for the popover on this PR: https://github.com/kiegroup/jbpm-wb/pull/652/files#diff-85ccc65888795800cb45c1d405460fadR28 Perhaps you can reuse it.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/client/resources/i18n/Constants.properties
```diff
@@ -21,6 +21,14 @@ CaseDetails=Case Details
 Owner=Owner
 Roles=Roles
 RoleAssignments=Role Assignments
+AssignedUsers=Assigned users
+AssignedGroups=Assigned groups
+AssignmentHelp=Intro user or groups separated by comma
```
cristianonicolai: Added similar text in here: https://github.com/kiegroup/jbpm-wb/pull/652/files#diff-79e45c754aaa85e67034c166a87e3e1bR99

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase/src/main/resources/case_definitions.json
```diff
@@ -4,7 +4,10 @@
     "name": "Example Case for Development",
     "containerId": "example",
     "roles": {
-      "participant" : -1
```
cristianonicolai: role owner is also returned from the kie-server and needs to be excluded, added this in here: https://github.com/kiegroup/jbpm-wb/pull/652/files#diff-e70320980c3b40bbe1c0b856e9fae313R7

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/static/case-mgmt.less
```diff
@@ -258,5 +258,69 @@ dd:not(:last-child) {
...
+  margin-top: -12px;
+  border: 1px solid #bbb;
+  background-color: @color-pf-white;
+  position : absolute;
+  width: 80%;
+  z-index: 2;
+}
+
+.kie-more-info{
+  //color: #363636;
```
cristianonicolai: remove?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/static/case-mgmt.less
```diff
@@ -258,5 +258,69 @@ dd:not(:last-child) {
...
+  text-decoration: none;
+  float: right;
+  margin-top: 6px;
```
cristianonicolai: _margin-top: 0px;_ seems to not squeeze the user icon.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/static/case-mgmt.less
```diff
@@ -258,5 +258,69 @@ dd:not(:last-child) {
   padding-left: 5px;
 }
 
+.kie-roles-item-line{
+  width: 90%;
```
cristianonicolai: duplicated width declaration.

cristianonicolai:
@nmirasch I notice that the group text is getting partially hidden. It needs some more space there. See below.

selection_108

Another nice to have feature would be to change the cursor style to _cursor:pointer;_ in all actionable items. For example, cross to remove user/group, expand user/groups and close popover.

You can currently end up with both popover open, perhaps there should be only one open.
selection_109
 
I'm also wondering if the icon to show users and groups should be aligned right below the kebab. Something like the image below.
selection_110

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -180,37 +234,40 @@ protected void removeGroupFromRole(final String groupName, final String roleName
 
         void removeAllRoles();
 
-        void addUser(String userName, String roleName, CaseRoleAction... actions);
+        void addOwnerAssignment(String roleName, String user, CaseRolesPresenter.CaseRoleAction... actions);
```
bsremac: Once the case owner is assigned, no actions should be available for that assignment, right?
So what is the purpose of the last argument?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -48,120 +58,164 @@ public String getTittle() {
...
+        if (cis.getRoleAssignments() != null && cis.getRoleAssignments().size() > 0) {
+            return cis.getRoleAssignments()
+                    .stream()
+                    .filter(ra -> roleName.equals(ra.getName()))
+                    .findFirst().orElse(CaseRoleAssignmentSummary.builder().name(roleName).build());
+        }
+        return CaseRoleAssignmentSummary.builder().name(roleName).build();
+
+    }
+
```
bsremac: Just a suggestion how to avoid returning "dummy" CaseRoleAssignmentSummary instances which doesn't seem to serve any purpose where this method is called:
```java
protected Optional<CaseRoleAssignmentSummary> getRoleAssignment(final CaseInstanceSummary cis, final String roleName) {
        if (cis.getRoleAssignments() != null && cis.getRoleAssignments().size() > 0) {
            return cis.getRoleAssignments()
                    .stream()
                    .filter(ra -> roleName.equals(ra.getName()))
                    .findFirst();
        }
        return Optional.empty();
```

Then in setupExistingAssignments method you are only taking into account CaseRoleAssignmentSummary instances which have the role with the name you are searching:
```java
cds.getRoles().keySet().stream().forEach(
                            roleName -> {
                                Optional<CaseRoleAssignmentSummary> caseRoleAssignmentSummary = getRoleAssignment(cis, roleName);
                                if(caseRoleAssignmentSummary.isPresent()) {
                                    List<CaseAssignmentItem> usersList = new ArrayList<CaseAssignmentItem>();
                                    caseRoleAssignmentSummary.get().getUsers()
                                            .forEach(u -> usersList.add(new CaseAssignmentItem() {
                                                @Override
                                                public String label() {
                                                    return u;
                                                }
...
```

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -48,120 +58,164 @@ public String getTittle() {
...
-                                            newRoleAssignmentView.getRoleName())));
+                    cds.getRoles().keySet().stream().forEach(
+                            roleName -> {
+                                CaseRoleAssignmentSummary caseRoleAssignmentSummary = getRoleAssignment(cis, roleName);
+                                List<CaseAssignmentItem> usersList = new ArrayList<CaseAssignmentItem>();
+                                caseRoleAssignmentSummary.getUsers()
+                                        .forEach(u -> usersList.add(new CaseAssignmentItem() {
+                                            @Override
+                                            public String label() {
+                                                return u;
```
bsremac: Could we have "u" renamed to "user", for better readability?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -48,120 +58,164 @@ public String getTittle() {
...
+                                                        translationService.format(REMOVE_USER_ASSIGNMENT_FROM_ROLE, u, roleName),
+                                                        () -> removeUserFromRole(u, roleName));
+                                            }
+                                        }));
+                                List<CaseAssignmentItem> groupsList = new ArrayList<CaseAssignmentItem>();
+                                caseRoleAssignmentSummary.getGroups()
+                                        .forEach(g -> groupsList.add(new CaseAssignmentItem() {
+                                            @Override
+                                            public String label() {
+                                                return g;
```
bsremac: Could we have "g" renamed to "group", for better readability?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -48,120 +58,164 @@ public String getTittle() {
...
-                        }
-                    }));
-                }
-        );
+            @Override
+            public void execute() {
+                editRoleAssignmentView.show(caseRoleAssignmentSummary,
+                        () -> assignToRole(caseRoleAssignmentSummary.getName(),
+                                editRoleAssignmentView.getUsersNames(),
+                                editRoleAssignmentView.getGroupsNames()
```
bsremac: Model (caseRoleAssignmentSummary) is already present as the first argument of in the _show_ call.
Why not calling directly:
caseRoleAssignmentSummary.getUsers() and 
caseRoleAssignmentSummary.getGroups()
instead of: 
editRoleAssignmentView.getUsersNames()
editRoleAssignmentView.getGroupsNames()?

Taking this into account, getUsersNames() and getGroupsNames() are unnecessary in EditRoleAssignmentView interface.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/EditRoleAssignmentViewImpl.java
```diff
@@ -124,17 +150,11 @@ public void hide() {
...
-            roleNameList.getElement().focus();
-            roleNameHelp.setTextContent(translationService.format(PLEASE_SELECT_ROLE));
-            roleNameGroup.setValidationState(ValidationState.ERROR);
-            validForm=false;
-        }
 
-        if (isNullOrEmpty(userNameInput.getValue()) && isNullOrEmpty(groupNameInput.getValue())) {
-            userNameInput.focus();
+        if (model.getUsers().size() + model.getGroups().size() == 0 ) {
+            users.focus();
```
bsremac: Are we going to allow entering users/groups with special characters? 
_!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~_

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -133,9 +134,51 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
     @Override
-    public void assignGroupAndUserToRole(final String serverTemplateId, final String containerId, final String caseId, final String roleName, final String user, final String group) {
-        client.assignGroupToRole(containerId, caseId, roleName, group);
-        client.assignUserToRole(containerId, caseId, roleName, user);
+    public void assignToRole(final String serverTemplateId, final String containerId, final String caseId,
+                             final String roleName, final List<String> usersList, final List<String> groupsList) {
+        CaseRoleAssignmentSummary ras = getCaseInstance(serverTemplateId, containerId, caseId).getRoleAssignments()
+                .stream()
+                .filter(ra -> roleName.equals(ra.getName()))
+                .findFirst().orElse(null);
```
bsremac: This looks very similar to what **getRoleAssignment** method defined in the presenter is doing.
Can this logic be delegated to presenter class and avoid fetching the case instance again from kie server?
That _hasAssignment_ is practically the result of this calculation, no?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -133,9 +134,51 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
+
+        if (ras != null) {
+            if (ras.getUsers() != null) {
```
bsremac: I think this condition can never evaluate to _false_.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -133,9 +134,51 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
+                List<String> usersToremove = new ArrayList();
+                ras.getUsers().forEach(
+                        user -> {
+                            if (!usersList.stream().filter(newU -> user.equals(newU)).findFirst().isPresent()){
+                                usersToremove.add(user);
```
bsremac: Instead of creating a new "holder" list for the users which are going to be removed, maybe it makes sense to call immediately:
_removeUserFromRole(serverTemplateId, containerId, caseId, roleName, user)_
That way, you wouldn't need to create a separate iteration for this.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -133,9 +134,51 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
+                groupsList.forEach(
+                        newGroup -> {
+                            if (!ras.getGroups().stream().filter(group -> newGroup.equals(group)).findFirst().isPresent()){
+                                assignGroupToRole(serverTemplateId, containerId, caseId, roleName, newGroup);
+                            }
+                        });
+            }
+        } else {
+           usersList.forEach(u -> assignUserToRole(serverTemplateId, containerId, caseId, roleName, u));
+            groupsList.forEach(g -> assignGroupToRole(serverTemplateId, containerId, caseId, roleName, g));
```
bsremac: This approach allows entering duplicate users/groups, as _CommaListValuesConverter_ doesn't perform any filtering if duplicates are entered.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -48,120 +58,161 @@ public String getTittle() {
...
+                    .stream()
+                    .filter(ra -> roleName.equals(ra.getName()))
+                    .findFirst().orElse(CaseRoleAssignmentSummary.builder().name(roleName).build());
+        }
+        return CaseRoleAssignmentSummary.builder().name(roleName).build();
+
     }
 
-    protected void addGroupToRole(final String groupName, final String roleName) {
+    protected Set<String> getRolesAvailableForAssignment(final CaseInstanceSummary cis, final CaseDefinitionSummary cds) {
```
bsremac: Where is this method used?
Where are we checking during assignments that the role cardinality hasn't been reached?
I see this method only used in Presenter test class.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
nmirasch: I think that are different things but if you thing retrieve he case instance is retrieve at the moment of saving the content just to have the latest values, I can pass this information in the call of the service at presenter. It looks good to me.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
nmirasch: It's true, I changed it in the next commit. Thanks!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
nmirasch: This two iterations are created to avoid the problem of ConcurrentModificationException that is thrown when you are iterating over a list and removing items from that list.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
nmirasch: I have added a 'distinct' call in this iterations to avoid duplicates.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
bsremac: I was just thinking that since we already have CaseInstanceSummary available in loadCaseInstance method of the presenter and the same filtering is done there, maybe it can be re-used and hence make this method lighter, nothing else :)

nmirasch:
@bsremac, @cristianonicolai Could you take a look? thanks again!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -140,9 +140,23 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
 
     @Override
-    public void assignGroupAndUserToRole(final String serverTemplateId, final String containerId, final String caseId, final String roleName, final String user, final String group) {
-        client.assignGroupToRole(containerId, caseId, roleName, group);
-        client.assignUserToRole(containerId, caseId, roleName, user);
+    public void assignToRole(final String serverTemplateId, final String containerId, final String caseId,
+                             final String roleName, final List<String> usersAssignments, final List<String> groupsAssignments,
+                             final List<String> oldUserAssignments, final List<String> oldGroupsAssignments) {
+
+        List<String> usersToRemove = new ArrayList<String>(oldUserAssignments);
```
cristianonicolai: Suggestion to move this bit into the presenter and use only the remove/assign methods.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
```diff
@@ -96,4 +220,59 @@ public void setLastElementStyle(boolean last){
...
+        roleAssignmentListOpenEventEvent.fire(new CaseRoleAssignmentListOpenEvent(roleName.getTextContent() + GROUPS_LINE));
+    }
+
+    @EventHandler("less-groups")
+    @SuppressWarnings("unsued")
+    public void onLessGroups(@ForEvent("click") final org.jboss.errai.common.client.dom.Event event) {
+        showLessItems(groupsLineDiv,roleGroups,moreGroupsLinkDiv,lessGroupsLinkDiv);
+    }
+
+    public void onCaseRoleAssignmentLineOpenEvent(@Observes CaseRoleAssignmentListOpenEvent event) {
```
cristianonicolai: If the event is only fired and received in here, could we just call the method instead?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
```diff
@@ -96,4 +220,59 @@ public void setLastElementStyle(boolean last){
...
+        if (event.getAssignmentLineId().equals(roleName.getTextContent() + USERS_LINE)) {
```
cristianonicolai: I think if you break this into two methods, one for users and another one for groups, you could avoid having the USER_LINE and GROUPS_LINE control.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
```diff
@@ -65,87 +70,113 @@
     @DataField("pagination")
     private PaginationViewImpl pagination;
 
-    List allElementsList = new ArrayList();
-
-    private Command userAddCommand;
+    List <CaseRoleItemView> allFilteredElements = new ArrayList();
```
cristianonicolai: These collections seems like state, ideally this should only be kept in the presenter. Same goes for actual filtering.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
```diff
@@ -65,87 +70,113 @@
...
+    public void init(final CaseRolesPresenter presenter) {
     }
 
-    @Override
-    public void enableNewRoleAssignments() {
-        removeCSSClass(userAddButton, "hidden");
-    }
+    @PostConstruct
+    public void init() {
+        stream(CaseRolesAssignmentFilterBy.values())
```
cristianonicolai: _stream(CaseRolesAssignmentFilterBy.values()).forEach_ seems all that you need in here to iterate over all elements, or even a simple _for_ would be enough.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase/src/main/java/org/jbpm/workbench/cm/server/MockCaseManagementService.java
```diff
@@ -188,14 +188,9 @@ public void assignGroupToRole(final String serverTemplateId, final String contai
...
     @Override
-    public void assignGroupAndUserToRole(final String serverTemplateId, final String containerId, final String caseId, final String roleName, final String user, final String group) {
-        executeOnCaseRole(caseId, roleName, r -> r.getGroups().add(group));
-        executeOnCaseRole(caseId, roleName, r -> r.getUsers().add(user));
-    }
-
-    @Override
     public void removeUserFromRole(final String serverTemplateId, final String containerId, final String caseId, final String roleName, final String user) {
         executeOnCaseRole(caseId, roleName, r -> r.getUsers().remove(user));
+        return;
```
cristianonicolai: No need to add a return statement in here.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -16,152 +16,227 @@
...
+                    .findFirst().orElse(CaseRoleAssignmentSummary.builder().name(roleName).build());
         }
+        return CaseRoleAssignmentSummary.builder().name(roleName).build();
     }
 
-    protected void addUserToRole(final String userName, final String roleName) {
-        caseService.call(
-                (Void) -> findCaseInstance()
-        ).assignUserToRole(serverTemplateId, containerId, caseId, roleName, userName);
+    protected List<String> validateRolesAssignments(final CaseDefinitionSummary caseDefinition,
```
cristianonicolai: Would be nice if we could share this code between both presenter classes, NewCaseInstancePresenter and this one.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/static/case-mgmt.less
```diff
@@ -258,5 +258,69 @@ dd:not(:last-child) {
...
+.kie-full-width {
+  width: 100%;
+  display:inline-block;
+}
+
+.kie-unassigned-color {
+  color: #8b8d8f;
+}
+
+.kie-unassigned-margin{
```
cristianonicolai: This css class is causing the text to get hidden, I believe it can be removed.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/static/case-mgmt.less
```diff
@@ -258,5 +258,69 @@ dd:not(:last-child) {
...
+  z-index: 3;
+}
+
+.kie-more-info-link{
+  color: #363636;
+  text-decoration: none;
+  cursor: pointer;
+}
 
+.kie-remove-role-item {
```
cristianonicolai: Missing _cursor: pointer;_ in here.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/static/case-mgmt.less
```diff
@@ -258,5 +258,69 @@ dd:not(:last-child) {
...
+  background-color: @color-pf-white;
+  position : absolute;
+  width: 88%;
+  z-index: 2;
+}
+
+.kie-more-info{
+  text-decoration: none;
+  float: right;
+  margin-top: 4px;
```
cristianonicolai: _margin-top_ could be removed, it would actually better align with user and group icons.

cristianonicolai:
@nmirasch regarding the icon to show the popover, after talking to Liz, it should actually be right next the last user or group and not right aligned. Sorry for the confusion.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
nmirasch: @cristianonicolai  I leave it into this method because when I moved it into presenter the case instance ((Void) -> findCaseInstance()) is called with any change in the assignments. Maybe we have to find another way for refreshing the assignments after all changes are done. What do you think?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
nmirasch: @cristianonicolai  the idea was all the displayed CaseRoleItem consume the event and close themselves it they are not the one who launched the event. If I call a simple method only the one who is calling notice the change. Make it sense?

nmirasch:
@cristianonicolai I do some changes in the implementation using ListComponent and fixing some of your comments. Remains the icons alignment and verify if we want do the call to the service for any change on the role list assignments.  I think this implementation is more homogeneous  with the other blocks that we have.

nmirasch:
@cristianonicolai, @bsremac  I have updated the way of 'more items calculation' adding an event to fill the items line, creating a new component (ItemsLine) that shows a list of items and adds the 'more' items link when they don't fit in one line. I have changed the popover setting up to improve the displaying of the message and to work with trigger:focus. Could you take a look? Thanks!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
```diff
@@ -84,16 +238,40 @@ public void addAction(final CaseRolesPresenter.CaseRoleAction action) {
         a.setOnclick(e -> action.execute());
 
         final HTMLElement li = getDocument().createElement("li");
+        if (!action.isEnabled()) {
+            addCSSClass(li, "disabled");
```
bsremac: Although disabled, the action is still clickable, after which **Remove assignment** dialog gets opened.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
```diff
@@ -16,64 +16,218 @@
...
-        addCSSClass(this.iconType, iconTypeClass);
+    public void setOwnerAssignment() {
+        usersItemsLine.initWithSingleItem(name.getTextContent(), caseRoleAssignmentSummary.getModel().getUsers().get(0));
+
+        addCSSClass(usersDiv, "kie-owner-role-height");
+        addCSSClass(groupsDiv, "hidden");
+        removeCSSClass(actions, "hidden");
+        addCSSClass(actionsItems, "hidden");
+        addCSSClass(actionsIcon, "kie-unassigned-color");
+    }
```
bsremac: Mouse cursor implies that the actions dropdown is clickable when hovering over it, although actions are disabled

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
```diff
@@ -16,64 +16,218 @@
...
+                public void execute() {
+                    confirmPopup.show(translationService.format(REMOVE_ASSIGNMENT),
+                            translationService.format(REMOVE),
+                            translationService.format(REMOVE_ALL_USERS_GROUPS_FROM_ROLE, model.getName()),
+                            () -> presenter.storeRoleAssignments(model.getName(), Collections.emptyList(),
+                                    Collections.emptyList(), model.getUsers(), model.getGroups()));
+                }
+            });
+        }
+    }
```
bsremac: Scenario:
1. Role doesn't have any assignments.
2. Click Edit to add role assignment.
3. Add some user/group then cancel the action (either by clicking Cancel or closing the dialog) 
4. Open actions dropdown for that role again (screenshot below)
screenshot from 2017-03-28 18-19-31

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -16,152 +16,144 @@
...
-                    (Void) -> findCaseInstance()
-            ).assignGroupAndUserToRole(serverTemplateId, containerId, caseId, roleName, userName,groupName);
-        }else {
-            if (!isNullOrEmpty(userName)) {
-                addUserToRole(userName, roleName);
-            }
-            if(!isNullOrEmpty(groupName)){
-                addGroupToRole(groupName,roleName);
-            }
+        if (cis.getRoleAssignments() != null && cis.getRoleAssignments().size() > 0) {
```
nmirasch: sure, thanks!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
nmirasch: fixed

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
nmirasch: now, applying disable to actions button and not only the text color.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
bsremac: Thinking from user experience, maybe it makes more sense for the actions dropdown to stay open when a user clicks on disabled _Remove all assignments_ i.e. that nothing changes in the UI after clicking in disabled area.
It could happen that a user clicks that area by accident whereas they wanted to click on _Assign_ action. Then the user needs to open dropdown again just to get to the action they really wanted.

bsremac:
@nmirasch @cristianonicolai I have one general remark when it comes to this redesign in particular, but also something to have in mind in the future. 
Taking into account the decision that only a handful of Selenium integration, happy-path tests will be allowed per feature (Selenium tests are expensive - slow, fault prone and difficult to maintain), the majority of logic is expected to be covered with unit tests, in community.
This means that only logic contained in Presenter class (and View free, util classes) is testable with unit tests.
Here, a lot of logic is moved to View classes, which are quite a few with a very complex dependencies. This means that any logic (by logic I mean every if/else construct) in these classes which is not covered with Selenium tests, won't be covered at all.
In some cases, it is possible to re-factor the code at some point later on, and draw some of the logic contained in View classes back to presenter and then write tests for it. But in this case, taking into account the number of View classes and their dependencies, that won't be an easy task.
Ideally, we should aim for "dummy" Views which are purely handlers of DOM and possibly nothing else.

nmirasch:
@bsremac @cristianonicolai I think the underlying problem is when we use the ListComponent the view is in charge off  how the items are displayed., controlling the DOM objects in their setValue method. In this implementation we have two extra things that makes that presentation logic go away from the presenter: how to display the user and groups lines. (component: itemsLine, and an event for the initialization using the container width) and the pagination( Component that manage the visible items loaded).
My suggestion could be add extra unit test for this view components to ensure that their behavior is the expected. 

For other side, I'm planning to remove the call findCaseInstance() after each removal or update and, as @cristianonicolai suggested in one or these comments, doing the best of using the list component updating the model and do several calls to addusertoRole and removeUserFromRo to do what currently is done by the service assignToRole that would be removed. 
What do you think?

bsremac:
@nmirasch If use of binding mechanism between model instances and view elements can get us away from making calls to findCaseInstance() after every update, I'm all for it. My only concern is again the same, that the simplest approach to achieve this would take away even more logic from presenter (actually back-end in this case, which is again better than to have logic in views :))

As for testability concerns, I don't know what kind of unit tests could be written for view classes. I have been perceiving the presenter as the last frontier when it comes to testing the logic.

nmirasch:
@cristianonicolai @bsremac I did the changes in calls from the presenter I told you in previous comments,  About the testing, I was trying it with the itemsLine and  even we can test the basic logic, it  have a lot of controls and mock generations to interact with the DOM objects that makes the feeling that they not provide too much value, maybe is the confirmation that "the presenter as the last frontier when it comes to testing the logic." :)

nmirasch:
I have renamed the commit to match with the corresponding jira https://issues.jboss.org/browse/JBPM-4647 that references Case Management Roles.

nmirasch:
I'm going to close this pr and open another one with the same commit and the right jira reference.
I hope it don't  cause any problem

PR #653, opened 2017-03-16
JBPM-5769 - Allow showcase app context to be the same the deployed wa…

cristianonicolai:
…r file name


cristianonicolai:
@krisv as discussed.

krisv:
+1

bsremac:
@cristianonicolai I can't figure it out from the changes whether this mean that Showcase App context is now configurable on Wildfly's start-up or the context is hard-coded to to whatever is the name of the Showcase App .war file.
The latter one works as expected, just not sure about the context configuration options.

cristianonicolai:
@bsremac exactly, this change only means that now the context should be the same as the final .war file name. There is no specific configuration for it on Wildfly.

PR #655, opened 2017-03-21
JBPM-5767 and JBPM-5894 - Migrate example case project to playground

cristianonicolai:
Remove reference to upstream gihub project
Use playground module for provisioning example project into kie server instance


cristianonicolai:
Depends on https://github.com/kiegroup/kie-wb-playground/pull/12

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase/pom.xml
```diff
@@ -403,6 +407,12 @@
       <artifactId>guava</artifactId>
     </dependency>
 
+    <!-- KIE Workbench Playground Repository -->
+    <dependency>
+      <groupId>org.kie.workbench.playground</groupId>
```
jhrcek: Please make sure the example has proper &lt;description&gt; in the pom as it gets used in the UI as mentioned in the [following issue](https://github.com/kiegroup/kie-wb-playground/issues/9)

cristianonicolai: @jhrcek it should be covered in here https://github.com/kiegroup/kie-wb-playground/pull/12/files#diff-73d7cad6005591e4b2ffd6e386d7abbbR13

cristianonicolai:
@bsremac any chance you could help verifying this one?

bsremac:
@cristianonicolai Sure. What would be the verification criterion?
When Casey is running in Dev mode, IT Orders definition is available, a case instance can be started and worked through?

cristianonicolai:
@bsremac it provisions the it orders example into a running kie server, you can check using your own setup and defining _org.kie.demo=true_ or the showcase docker image. Once the provisioning is done (takes a few minutes to complete), you should be able to start an IT Order case definition.

bsremac:
@cristianonicolai It's not working for me. I didn't use the docker image, but I did use the latest available snapshots of kie-wb and kie-server.
I built Casey war from this PR but had to make the changes in _RemoteCaseManagementServiceImpl_ and _RemoteCaseManagementServiceImplTest_ in order to pass compile due to the changes in the latest kie-server-client api i.e.
```java
-        final List<CaseInstance> caseInstances = client.getCaseInstances(singletonList(request.getStatus()), 0, PAGE_SIZE_UNLIMITED);
+        final List<CaseInstance> caseInstances = client.getCaseInstances(singletonList(request.getStatus().toString()), 0, PAGE_SIZE_UNLIMITED);
```
In order to deploy the latest kie-wb snapshot, I had to use that EAP/Wildfly switch that Jan posted to bsig-all: _-Dorg.uberfire.nio.git.ssh.enabled=false_.
After setting everything up, here's the error I'm getting. EAP log with errors attached.
[error.txt](https://github.com/kiegroup/jbpm-wb/files/872769/error.txt)

cristianonicolai:
@bsremac I've rebased this PR to allow you to build. The workaround is fine, I don't think there is any relation to the error. In fact this seems to be caused by the Maven settings. Do you have the _repository.jboss.org_ repository enabled?

bsremac:
@cristianonicolai Thanks for rebasing this PR. 
You were right, my Maven settings were the problem (feel a bit stupid now :))

IT Orders deploys successfully but there's a problem with Casey. From within Casey, I can see the case definition and I can start it, but after a new case instance has been started, I don't see it anymore in the Case List. Was expecting to see the new instance listed among opened ones, of course, but I shuffled through all the statuses just to make sure it didn't land somewhere where it shouldn't.
How do I know new instances were started? - They are listed in Process Instances List in Workbench

Another thing I didn't like much is the user experience in Workbench _Deploy > Execution Servers_ perspective, once the container with id: itorders-1.0.0-SNAPSHOT is deployed to Kie Server.
When the user navigates to that page, they can't see the container immediately.
The user needs to know that this container was deployed outside of the current template and that is hidden in **Remote Servers** drop-down once they click on the kie server instance to which the container was deployed. Otherwise, they are not aware of its existence.

cristianonicolai:
@bsremac I still couldn't reproduce this error where the cases disappear from the list. Just checking, but I guess you're assigning the case owner as the current user? And that user is the same used to connect with the kie server? Are you able to reproduce this error using the complete docker image?

Regarding the Remove Servers view, I agree that seems a bit odd but I guess that view is designed to focus on the server templates and not on specific server instances. Because we're provisioning the demo into a kie server instance and not looking at adding it to the server template, we get this different view. To be honest not sure what could be changed, adding the container to a server template would mean to make the showcase depend on the workbench to work.

bsremac:
@cristianonicolai First to answer your questions regarding the previous round of testing:
- I didn't use the docker solution, only manual setup.
- I was leaving the default case owner i.e. the currently logged-in user.
- I was logging into Casey with an arbitrary user (not the one specified with _org.kie.server.user/pwd_ switches). My current understanding is, based on **Authentication options** section found in 
https://github.com/kiegroup/jbpm-wb/tree/master/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase, that regardless of which user is currently logged into Casey, Casey will, under the hood, still use credentials found in _org.kie.server.user/pwd_ switches for communication with Kie Server, that is if _login-module_ setup is omitted for Casey .war (Authentication option No.2 in the above document), which was the case in my setup.

With the latest snapshots available today, I had the following results:
1. Using the complete docker image, ITOrders example project doesn't get deployed at all. (I think I waited long enough for deploying part to occur)
2. Using my manual setup, I see now three containers deployed, compared to only one that I was seeing in the previous round of testing. By what I see in Process Definitions perspective in Workbench, I would say that all three existing variations of ITOrders have been deployed. Screenshot below:
screenshot from 2017-03-31 17-56-38

As a consequence, when trying to start a case instance through Casey, the following error gets returned:

screenshot from 2017-03-31 17-59-12

bsremac:
@cristianonicolai As for the example container visibility within Workbench, I get your point. It's just that the current example exposure within _Execution Servers_ perspective puts it on the margin for the customers to notice it, compared to, for example, other playground examples being offered first when users access new _Library_ perspective.

bsremac:
@cristianonicolai I built Casey off of PR and got ITOrders example project provisioned in docker environment by adding _org.kie.demo=true_ property to _standalone-full-kie.xml_.

Good news is that I don't see multiple example projects deployed, but only one, in both, docker setup and in my local setup. You said you rebased the PR. Could that solve the problem with multiple example projects?

In my local setup, I still can't see started case instances. I tried also with adding KieLoginModule for Casey to _standalone.xml_  - it didn't help.

In docker environment, I see the started instances, but when I access them, I get the following error: (I tried this with both users _admin/admin_ and _user/user_ and the  result is the same):

docker

After I close the error dialog, it seems that I don't have any problems interacting with the case instance.
Pay attention to Role Assignments block. Where does the bogus owner entry comes from?

cristianonicolai:
@bsremac sounds like we're getting closer results now! :) The error you got when using the docker image is the same I'm facing...still couldn't figure that one but at least we both can reproduce it. Still not sure why cases disappear for your manual setup, still need to check that.
As for the bogus owner, that is duplicated, it will be fixed by @nmirasch's Case Roles pr.

cristianonicolai:
@bsremac mind checking again, the authorization issue has been fixed in here: https://github.com/kiegroup/jbpm-wb/pull/670

bsremac:
@cristianonicolai I tried with both available users in Casey, _admin/admin_ and _user/user_, and the error is not seen anymore :)

cristianonicolai:
@bsremac cool, so I expect that the provisioning is now working for you at the least when using the docker image, correct?

bsremac:
@cristianonicolai Yes, we can say that. Docker is a go :)
I'm curious, though, what kind of setup was Maciej using when testing his fix in https://github.com/kiegroup/jbpm-wb/pull/670
If he managed to run everything without a docker, maybe there's something missing in my setup that I can't figure out.

mswiderski:
@bsremac I'll write an article with all the steps to run it on standalone servers. will drop you a note once done

bsremac:
@mswiderski Thanks a lot :)

mswiderski:
@bsremac @cristianonicolai @nmirasch  here it is http://mswiderski.blogspot.com/2017/04/case-management-application-in-workbench.html
let me know if I didn't mix up anything ...

cristianonicolai:
@mswiderski looks great, thanks a lot for that! Only minor fix is that I end up changing one of the system properties after the PR instructions I gave you. 
__org.jbpm.casemgmt.showcase.wildfly.host_
_optional host name that wildfly is running on - default localhost__
should actually be
__oorg.jbpm.casemgmt.showcase.wildfly.management-host_
_optional wildfly management host name - default localhost__

You could also include in the note for the local deployment that instead a system property could be used to point to the war file in the local file system: _org.jbpm.casemgmt.showcase.path_

mswiderski:
@cristianonicolai article updated, thanks

cristianonicolai:
@mswiderski thanks!

bsremac:
@mswiderski That user which you used in Case Management Showcase, _maciek_, which roles are assigned to this user in _application-roles.properties_ in your setup? The article doesn't mention _user_ role, which is mandatory in my setup in order to login into Case Management Showcase.

mswiderski:
@bsremac here are the roles that I have that user:
maciek=HR,PM,admin,IT,Accounting,kie-server,Administrators
not all of them are needed for the example as most of it is actually driven by case role assignment - I used HR and PM for the manager and supplier to make them as group assignments.

bsremac:
@cristianonicolai I can't figure out what the problem is in my setup, so that I can't see the case list.
In order to have a user which is authorized to login to Casey, I have to add _user_ role to that user. Having only _kie-server_ and _admin_ roles is not enough.
When I try to refresh the Case List perspective, this is what I see in browser console:

screenshot from 2017-04-10 22-33-43

cristianonicolai:
@bsremac the role _user_ is definitely mandatory for using the showcase. @mswiderski mind adding that to the blog?
So far the only situation I can reproduce for not showing the started case in the list is when using pre-defined user to connect with the Kie Server (_org.kie.server.user_ and _org.kie.server.pwd_), ie _admin_ and connecting to Casey with another user, ie _testadmin_. New case instances will be assigned to _testadmin_ as owner and not displayed because the connection is done with the _admin_ user. That still doesn't explain the exception you're getting. Apart from the UI error, is there any useful logs?

mswiderski:
@cristianonicolai updated article to state that required roles are user and kie server

bsremac:
@cristianonicolai The kind of setup with which you were able to reproduce the issue is exactly the kind of setup I'm running :)
That means, a dedicated user for communication with Kie Server (set through _org.kie.server.user_ and _org.kie.server.pwd_) and any number of arbitrary users who only have _user_ role assigned to them. I was expecting that in this kind of setup, these users can see the case list with cases they own.

If this is expected behavior, what is the alternative setup which allows multiple users to see the case list?
I tried with adding _kie-server_ role to these users along with _user_ role along with adding that Kie login module to _standalone.xml_ file i.e.
<code="org.kie.security.jaas.KieLoginModule" flag="required" module="deployment.jbpm-wb-case-mgmt-showcase-7.0.0-SNAPSHOT.war"/>
My understanding is that with this login module added, the user set by _org.kie.server.user_ and _org.kie.server.pwd_ won't be used and logged-in users will be able to see their cases, but that is not happening. The result is the same - empty case list.

cristianonicolai:
@bsremac if a specific user is used to connect with the Kie Server(via _org.kie.server.user_ and _org.kie.server.pwd_), all data will be retrieved as that user. If you create an instance and set a different owner, the case instance should not be returned. Only cases where the user used for the Kie Server connection has permission should be listed.

As for the second part, using the login module is the correct setup to allow multiple users to start and see their cases. I think the issue here is because you still have _org.kie.server.user_ and _org.kie.server.pwd_ set as system property. If you set it that will take precedence over the login module, which in this case isn't used at all. So for the login module to actually work you can't set these sytem properties.

mswiderski:
@cristianonicolai where is the implementation that takes the system property given user instead of logged in user? Is this done on purpose in case mgmt showcase?

cristianonicolai:
@mswiderski this is actually a code that is shared for both, workbench and case management showcase. Both behave the same as it is now. Take a look here: https://github.com/kiegroup/jbpm-wb/blob/master/jbpm-wb-kie-server/jbpm-wb-kie-server-api/src/main/java/org/jbpm/workbench/ks/utils/KieServerUtils.java#L48

mswiderski:
then that will be a problem. Main purpose of the system properties given user/password is for administrative purpose - to use for the AdminClient instances and for the server management to properly operate.
It might not be so common for us to set them but in most of the cases users will set it to not rely on defaults (that everyone knows :))

cristianonicolai:
@mswiderski tricky one :P Perhaps we should use different system properties for each type of connection? Trying to fallback from the login module doesn't seem trivial to me.

mswiderski:
I'd say we should never use system property based user to perform operations on behalf of logged in user. This completely bypass security, why do we need that?

cristianonicolai:
Agreed, let me review that and remove this.

cristianonicolai:
@bsremac @mswiderski ok so updated this PR removing the option to use a single user to connect to the Kie Server. I've also updated to docs to better describe the available settings.

bsremac:
@cristianonicolai I have built Casey from this PR and I'm still seeing the same behavior.
Login module was added for Casey and _org.kie.server.user_/_org.kie.server.pwd_ were set too. I still can see the cases only when logged into Casey with the user from the properties. Any other user with _user_ role can't see the list.

Which document have you updated?
This one still holds the old setup info:
https://github.com/kiegroup/jbpm-wb/tree/master/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase

cristianonicolai:
@bsremac updated the README file in the PR itself: https://github.com/cristianonicolai/jbpm-wb/blob/4c92b54ab3b3a019a33aeaeec247d6f79d10b682/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase/README.md

cristianonicolai:
@bsremac here is what I got:

user and roles:
```
user=kie-server,user
admin=kie-server,user,admin
```
system properties:
```
<system-properties>
        <property name="org.kie.demo" value="true"/>
        <property name="org.jbpm.casemgmt.showcase.url" value="/jbpm-cm"/>
        <property name="org.kie.server.user" value="admin"/>
        <property name="org.kie.server.pwd" value="admin"/>
        <property name="org.uberfire.nio.git.daemon.enabled" value="false"/>
        <property name="org.uberfire.nio.git.ssh.enabled" value="false"/>
        <property name="org.kie.server.id" value="test-kie-server"/>
        <property name="org.kie.server.controller" value="http://localhost:8080/kie-wb/rest/controller"/>
        <property name="org.kie.server.location" value="http://localhost:8080/kie-server/services/rest/server"/>
    </system-properties>
```

login module:
```
<login-module code="org.kie.security.jaas.KieLoginModule" flag="required" module="deployment.kie-wb.war"/>
```

with this setup I can get that case instances started by user _admin_ are only shown to this user and same for the user _user_, it can only see the case instances started by this user.

bsremac:
@cristianonicolai This is what I have.
users and roles:
```java
bamuser=user,kie-server
kieserver=kie-server,user
testadmin=admin,kie-server,user
```

system properties:
```java
    <system-properties>
        <property name="org.kie.demo" value="true"/>
        <property name="org.guvnor.project.gav.check.disabled" value="true"/>
        <property name="activemq.artemis.client.global.thread.pool.max.size" value="30"/>
        <property name="org.kie.server.user" value="kieserver"/>
        <property name="org.kie.server.pwd" value="kieserver123*"/>
        <property name="org.kie.server.location" value="http://${jboss.bind.address:127.0.0.1}:${jboss.http.port:8080}/kie-server/services/rest/server"/>
        <property name="org.kie.server.controller.user" value="kiecontroller"/>
        <property name="org.kie.server.controller.pwd" value="kiecontroller123*"/>
        <property name="org.kie.server.controller" value="http://${jboss.bind.address:127.0.0.1}:${jboss.http.port:8080}/business-central/rest/controller"/>
        <property name="org.kie.server.id" value="test-kie-server"/>
        <property name="org.jbpm.casemgmt.showcase.deploy" value="true"/>
    </system-properties>
```

login module:
```java
<login-module code="org.kie.security.jaas.KieLoginModule" flag="required" module="deployment.business-central.war"/>
```
I can start a case instance with any of users.
But when it comes to case instances list, all three users see only case instances which have _org.kie.server.user_ (**kieserver** in my case) as the case owner.

On Workbench side, all started case instances are shown in Process Instances perspective, but in Process Variables tab for each case instance, as initiator is always shown _org.kie.server.user_, no matter which user started the case and which user was set as case owner.

screenshot from 2017-04-18 11-04-50

bsremac:
@cristianonicolai Tested with deployable Casey.war and had the same results.
Only cases owned by _org.kie.server.user_ are listed for all logged-in users.

You can find below what I mentioned regarding login modules differentiation at runtime:
```java
<login-module code="org.kie.security.jaas.KieLoginModule" flag="optional" module="deployment.business-central.war"/>
<login-module name="org.kie.security.jaas.KieLoginModule-2" code="org.kie.security.jaas.KieLoginModule" flag="optional" module="deployment.jbpm-wb-case-mgmt-showcase-7.1.0-SNAPSHOT.war"/>
```

cristianonicolai:
Depends on:
https://github.com/kiegroup/kie-wb-common/pull/789
https://github.com/AppFormer/uberfire/pull/668

bsremac:
Was quite a ride :)
+1

cristianonicolai:
@bsremac definitely! :)

PR #656, opened 2017-03-21
JBPM-5800 - Included showcase app link if war is manually deployed

cristianonicolai:
Added new system property org.jbpm.casemgmt.showcase.url to allow specifying the context/url where the case management showcase app is deployed.


cristianonicolai:
@krisv last part needed for the jbpm installer.

krisv:
+1

PR #657, opened 2017-03-21
[BPMSPL-637] Add new fields to task list

rbarriuso:


cristianonicolai:
ok to test

bsremac:
@cristianonicolai The last comment in implementing JIRA mentions implementing Task Perspective in such way so that it has two variants depending on the role of user accessing it.
How is that comment related to the JIRA topic? Will certain fields/columns in the task list be hidden from non-admin user?

cristianonicolai:
@bsremac there would be no difference regarding the columns that are accessible in both views. The comment is about moving the Admin tab into a separated menu, where only admins would be allowed.

bsremac:
@rbarriuso thank you for the clarification

rbarriuso:
I've fixed the problem about the test failing.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/DataSetDefsBootstrap.java
```diff
@@ -35,149 +36,120 @@
...
+            "t.processSessionId, " +
+            "t.status, " +
+            "t.taskId, " +
+            "t.workItemId, " +
+            "pil.correlationKey, " +
+            "pil.processInstanceDescription, " +
+            "lmd.lastModificationDate ";
+    
+    private static final String SQL_SUBQUERY_LAST_MOD_DATE = "(" +
+            "select te.taskId, max(te.logTime) as lastModificationDate " +
```
cristianonicolai: @mswiderski I'm a bit concerned about the query performance adding another join, should we look into adding the last update date into the AuditTaskImpl object, it seems to me that this object is always updated anyway before triggering a new TaskEventImpl.

mswiderski: fully agree, that will be much better approach and not so big issue as we can provide update script to populate that column with data based on TaskEvent anyway

cristianonicolai: @mswiderski, I created [JBPM-5855](https://issues.jboss.org/browse/JBPM-5855) to handle that specifically.
@rbarriuso perhaps you could break this change into two, adding the process related fields now and the last update once JBPM-5855 is done.

rbarriuso:
Added changes which depend on https://github.com/kiegroup/jbpm/pull/793

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/DataSetDefsBootstrap.java
rbarriuso: Done

cristianonicolai:
@rbarriuso just a note, to get the build to work across repositories, you need to match the branch names for you PR's.

cristianonicolai:
Jenkins please retest this

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/resources/org/jbpm/workbench/ht/client/resources/i18n/Constants.properties
```diff
@@ -34,6 +34,9 @@ Process_Name=Process Name
 Process_Instance_Id=Process Instance Id
 Process_Definition_Id=Process Definition Id
 Process_Instance_Details=Process Instance Details
+Process_Instance_Correlation_Key=Process instance correlation key
```
cristianonicolai: Just because this String is used as the column header I would suggest to drop the _instance_ and use only _Process Correlation Key_

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/resources/org/jbpm/workbench/ht/client/resources/i18n/Constants.properties
```diff
@@ -34,6 +34,9 @@ Process_Name=Process Name
...
+Process_Instance_Description=Process instance description
```
cristianonicolai: Same as above, _Process Description_

cristianonicolai:
@rbarriuso looks good, just a minor comment regarding the column headers string.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/resources/org/jbpm/workbench/ht/client/resources/i18n/Constants.properties
jhrcek: Since the correlation key is really associated with a process instance, I find _Process Correlation Key_ a bit confusing. If you consider shortening for the sake of shortness for UI display, I'd prefer _Correlation Key_ @cristianonicolai WDYT?

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/resources/org/jbpm/workbench/ht/client/resources/i18n/Constants.properties
cristianonicolai: keeping it as _Process instance description_

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/resources/org/jbpm/workbench/ht/client/resources/i18n/Constants.properties
cristianonicolai: @jhrcek agreed, keeping it as _Correlation Key_

rbarriuso:
Changed column labels as agreed

cristianonicolai:
@rbarriuso thanks

PR #654, opened 2017-03-16
[BPMSPL-632] Add correlation key and last modified date to process in…

rbarriuso:
…stance list

Removed dead code.

PR #663, opened 2017-03-28
Remove unused code for handling null table settings

cristianonicolai:


cristianonicolai:
@nmirasch applying the same login to the other presenters following @rbarriuso's change in here: https://github.com/kiegroup/jbpm-wb/pull/654

nmirasch:
+1
@cristianonicolai, it would never happen that the dataset API wouldn't be applied. Thanks!

PR #665, opened 2017-03-29
Fix compilation issue, remove unused DashboardFactory

cristianonicolai:


jhrcek:
+1 Without this jbpm-wb doesn't compile after dashbuilder refactoring

PR #668, opened 2017-03-31
JBPM-5770: Use fixed height for blocks/cards

nmirasch:
Updated the css styles with the latest ones
Added scroll in milestone to ensure the fixed height
Added a min-height restriction over the displayed cards
Changed the page_size of actions and roles to fit with the design provided by UX


cristianonicolai:
@nmirasch looks good to me, only minor comment is If you can just clean up some of the comment out code in the `case-mgmt.less` file.

bsremac:
Looks nice Neus. Just a couple of observations.
- What determines the length of Stage block? Why are they made uneven?
stages

- It seems that the height of the last (3rd) filed in Actions, Roles and Comments list is slightly smaller, causing the text to be shadowed in case of Actions and Comments, and a bit tight for role name in roles

actions
comments
roles

nmirasch:
@cristianonicolai I have removed the commented styles that wouldn't be used
@bsremac This designs are the ones that provided us the UX team. The first the active and the last stages are show wider and the rest of them share the rest of space. The shadow at the beginning or at the end of the tables with pagination, they thought that is a good way to highlight that there are more elements ahead or behind.

bsremac:
@nmirasch Thank you for the clarifications :)
+1

PR #669, opened 2017-03-31
JBPM-4647: Case Roles. Redesign

nmirasch:
This is the same commit as https://github.com/kiegroup/jbpm-wb/pull/650 but with the right jira reference to Case Management Roles. 
@cristianonicolai  @bsremac could you take a look?
Thanks!


jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.java
```diff
@@ -16,84 +16,266 @@
...
+    ConfirmPopup confirmPopup;
+
+    @Inject
+    protected TranslationService translationService;
+
+    @Inject
+    @AutoBound
+    private DataBinder<CaseRoleAssignmentSummary> caseRoleAssignmentSummary;
+
+    boolean listenLoadLineEvent=true;
```
cristianonicolai: If this is only set to false when the instance is destroyed, is it needed at all?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -16,201 +16,223 @@
...
-                    final Integer roleCardinality = cds.getRoles().get(role);
-                    if (roleCardinality == -1) {
-                        return true;
-                    }
-                    final Integer roleInstanceCardinality = cis.getRoleAssignments().stream().filter(ra -> role.equals(ra.getName())).findFirst().map(ra -> ra.getGroups().size() + ra.getUsers().size()).orElse(0);
-                    return roleInstanceCardinality < roleCardinality;
-                }
-        ).collect(Collectors.toSet());
+    public void editAction(CaseRoleAssignmentSummary caseRoleAssignmentSummary) {
+        CaseRoleAssignmentSummary previousCaseRoleAssignmentSummary= getRoleAssignmentByName(caseRoleAssignmentSummary.getName());
```
bsremac: It looks to me that the only piece of data which connects _editAction_, _assignToRole_ and _storeRoleAssignments_ methods is the role name in  _CaseRoleAssignmentSummary_ model that is passed to _editAction_.
Maybe it would be more readable if we could make this connection clear and strip away methods from unnecessary computing of data which they don't use.
I'm thinking something along the lines of:
- remove _getRoleAssignmentByName_ call in _editAction_
- change _assignToRole_ signature so it doesn't take any arguments (the ones it takes now it doesn't use but just forwards)
- remove _oldUserAssignments_ and _oldGroupsAssignments_ lists from _storeRoleAssignments_ arguments and get them from that **cras** reference before you start changing the object it points to.

nmirasch:
@cristianonicolai @bsremac includes the las UX team too. Let's see it this is the good one. :)

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/EditRoleAssignmentViewImpl.java
```diff
@@ -124,20 +163,14 @@ public void hide() {
...
-        if (roleNameEmpty) {
-            roleNameList.getElement().focus();
-            roleNameHelp.setTextContent(translationService.format(PLEASE_SELECT_ROLE));
-            roleNameGroup.setValidationState(ValidationState.ERROR);
-            validForm=false;
-        }
 
-        if (isNullOrEmpty(userNameInput.getValue()) && isNullOrEmpty(groupNameInput.getValue())) {
-            userNameInput.focus();
+        if (model.getUsers().size() + model.getGroups().size() == 0 ) {
```
cristianonicolai: model.hasAssignments()

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -16,203 +16,214 @@
...
-        }else {
-            if (!isNullOrEmpty(userName)) {
-                addUserToRole(userName, roleName);
-            }
-            if(!isNullOrEmpty(groupName)){
-                addGroupToRole(groupName,roleName);
-            }
+    public void editAction(CaseRoleAssignmentSummary caseRoleAssignmentSummary) {
+
+        editRoleAssignmentView.setValue(CaseRoleAssignmentSummary.builder()
```
nmirasch: here the role assignment is clone to allow the cancel action and restore the old values

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -16,203 +16,214 @@
...
-        caseService.call(
-                (Void) -> findCaseInstance()
-        ).assignGroupToRole(serverTemplateId, containerId, caseId, roleName, groupName);
+    protected CaseRoleAssignmentSummary getRoleAssignmentByName(String roleName) {
+        return allUnfilteredElements.stream()
+                .filter(caseRoleAssignmentSummary -> roleName.equals(caseRoleAssignmentSummary.getName()))
+                .findFirst().get();
+    }
+
+    protected void storeRoleAssignments(String roleName, List<String> users, List<String> groups) {
```
nmirasch: this method has those parameters to allow the removeAllRoleAssignments actions passing empty list as users and groups. I can do an specific method for that purpose.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -16,203 +16,214 @@
...
+        CaseRoleAssignmentSummary previousCaseRoleAssignmentSummary = getRoleAssignmentByName(roleName);
+        List<String> previousUserAssignments = previousCaseRoleAssignmentSummary.getUsers();
+        List<String> previousGroupsAssignments = previousCaseRoleAssignmentSummary.getGroups();
+        previousCaseRoleAssignmentSummary.setGroups(new ArrayList<>(groups));
+        previousCaseRoleAssignmentSummary.setUsers(new ArrayList<>(users));
+        filterElements();
+
+        List<String> usersToRemove = new ArrayList<String>(previousUserAssignments);
+        usersToRemove.removeAll(users);
+        usersToRemove.forEach(
```
nmirasch: I assumed the distinct was applied in the previous assignments addition, but if you think is a good idea to ensure that, I can add it

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
nmirasch: Sorry,  I understand now pass the model instead the roleName.. yes, I'll do that change

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
```diff
@@ -16,203 +16,214 @@
...
         return translationService.format(ROLES);
     }
 
+    private CaseDefinitionSummary caseDefinition;
+
     @Override
     protected void clearCaseInstance() {
-        view.disableNewRoleAssignments();
         view.removeAllRoles();
+        allUnfilteredElements = new ArrayList();
```
nmirasch: sure

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-api/src/main/java/org/jbpm/workbench/cm/model/CaseRoleAssignmentSummary.java
```diff
@@ -64,6 +64,10 @@ public void setUsers(final List<String> users) {
         this.users = ofNullable(users).orElse(new ArrayList<>());
     }
 
+    public boolean hasAssignment(){
+        return getUsers().size() +getGroups().size() > 0;
+    }
+
```
bsremac: Since the method is moved here, maybe we don't need to fetch users and groups through getters but rather use them directly.

nmirasch:
@bsremac changed, thanks!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesPresenter.java
cristianonicolai: @nmirasch I could actually see duplicated user/group names in the demo version at least. I think it could be because of the fact that we're using lists and the converter also doesn't prevent duplicates. Maybe the fix isn't in here specifically but needs a double check.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRolesViewImpl.java
```diff
@@ -50,102 +63,112 @@
...
     @Override
-    public void setupPagination() {
-        rolesBadge.setTextContent(String.valueOf(allElementsList.size()));
-        pagination.init(allElementsList, this, PAGE_SIZE);
+    public void setTotalNumberOfRolesBadge(String totalNumberOfRolesBadge) {
+        rolesBadge.setTextContent(totalNumberOfRolesBadge);
     }
 
     @Override
     public void setVisibleItems(List visibleItems) {
```
cristianonicolai: one pending from last review.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/events/CaseRoleAssignmentListLoadEvent.java
```diff
@@ -0,0 +1,40 @@
...
+    public int getMaxWidth() {
+        return maxWidth;
+    }
+
+    public CaseRoleAssignmentListLoadEvent() {
+    }
+
+    @Override
+    public String toString() {
+        return "CaseRoleAssignmentListLoadEvent{" +
```
cristianonicolai: Perhaps include maxWidth into toString

nmirasch:
@cristianonicolai  fixed the control of duplicated users or groups, applied code styling in some classes, used typed List in the component for pagination and changed the toString method of  CaseRoleAssignmentListLoadEvent. Could you take a look? Thanks!!

cristianonicolai:
@nmirasch great result, well done!

PR #670, opened 2017-04-04
Small adjustments to make case app and workbench work smoothly to com…

mswiderski:
…plete it order case instances

@cristianonicolai here are the fixes we talked about. With them and the PR for kie-wb-playground I was able to run completely IT Order case with workbench and case mgmt showcase app integrated with workbench. A small article will be published tomorrow about this.


cristianonicolai:
@mswiderski thanks a lot for this! Can you just check the _CaseRolesPresenterTest_ class which has some failing tests due to this change.

mswiderski:
test should be fixed now

cristianonicolai:
@mswiderski thanks!

PR #671, opened 2017-04-05
JBPM-5864 - Allow start case instance to load case definitions with s…

cristianonicolai:
…ame id and name from different containers

Add container name as sub-text in case there is multiple cases with same name.

### DIFF
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstancePresenter.java
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstancePresenter.java
@@ -16,14 +16,12 @@
 
 package org.jbpm.workbench.cm.client.newcase;
 
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.Optional;
 import javax.enterprise.context.Dependent;
 import javax.enterprise.event.Event;
 import javax.inject.Inject;
 
-import com.google.common.collect.Iterables;
 import org.jboss.errai.bus.client.api.messaging.Message;
 import org.jboss.errai.common.client.api.Caller;
 import org.jboss.errai.security.shared.api.identity.User;
@@ -38,15 +36,14 @@ import org.jbpm.workbench.cm.service.CaseManagementService;
 import org.uberfire.client.mvp.UberElement;
 import org.uberfire.workbench.events.NotificationEvent;
 
-import static com.google.common.base.Strings.isNullOrEmpty;
 import static java.util.Collections.singletonList;
+import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
-import static java.util.stream.Collectors.toMap;
 
 @Dependent
 public class NewCaseInstancePresenter extends AbstractPresenter<NewCaseInstancePresenter.NewCaseInstanceView> {
 
-    private Map<String, CaseDefinitionSummary> caseDefinitions = new HashMap<>();
+    private List<CaseDefinitionSummary> caseDefinitions = emptyList();
 
     private Caller<CaseManagementService> caseService;
 
@@ -63,17 +60,14 @@ public class NewCaseInstancePresenter extends AbstractPresenter<NewCaseInstanceP
     @Inject
     private User identity;
 
-    protected void loadCaseRoles(final String caseDefinition) {
+    protected void loadCaseRoles(final String caseDefinitionId) {
         view.clearRoles();
-        if (isNullOrEmpty(caseDefinition)) {
-            return;
-        }
-        final CaseDefinitionSummary caseDefinitionSummary = caseDefinitions.get(caseDefinition);
-        if (caseDefinitionSummary == null) {
-            return;
-        }
-        final List<CaseRoleAssignmentSummary> roles = caseDefinitionSummary.getRoles().keySet().stream().filter(r -> "owner".equals(r) == false).map(r -> CaseRoleAssignmentSummary.builder().name(r).build()).collect(toList());
-        view.setRoles(roles);
+        final Optional<CaseDefinitionSummary> cds = getCaseDefinitionSummary(caseDefinitionId);
+        cds.ifPresent(d -> view.setRoles(d.getRoles().keySet().stream().filter(r -> "owner".equals(r) == false).map(r -> CaseRoleAssignmentSummary.builder().name(r).build()).collect(toList())));
+    }
+
+    private Optional<CaseDefinitionSummary> getCaseDefinitionSummary(final String caseDefinitionId) {
+        return caseDefinitions.stream().filter(d -> d.getUniqueId().equals(caseDefinitionId)).findFirst();
     }
 
     public void show() {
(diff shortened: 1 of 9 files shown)


cristianonicolai:
@bsremac @nmirasch please have a look once you have a chance.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/util/Select.java
```diff
@@ -52,10 +53,18 @@ public void addOption(final String text, final String value) {
     }
 
     public void addOption(final String text, final String value, final Boolean selected) {
+        addOption(text, null, value, selected);
+    }
+
+    public void addOption(final String text, final String subText, final String value, final Boolean selected) {
```
bsremac: Do we need to expose the boolean argument?
It seems that in usages of all overloaded methods, only _false_ value is used.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
```diff
@@ -151,13 +152,24 @@ public void clearRoles() {
...
     @Override
-    public void setCaseDefinitions(final List<String> definitions) {
+    public void setCaseDefinitions(final List<CaseDefinitionSummary> definitions) {
         clearCaseDefinitions();
         caseTemplatesList.setValue("");
-        for (final String definition : definitions) {
-            caseTemplatesList.addOption(definition);
-        }
+        definitions.forEach(d -> {
+            long sameNameDefinitions = definitions.stream().filter(dn -> dn.getName().equals(d.getName()) && dn.getUniqueId().equals(d.getUniqueId()) == false).count();
```
bsremac: I would personally go with _findFirst()_ instead of _count()_ and then check if the Optional returned is empty or not in the if/else part. Just because the number returned by _count()_ (when it's greater than 0) is irrelevant .

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstancePresenter.java
```diff
@@ -104,37 +96,37 @@ protected void loadCaseDefinitions() {
     protected void createCaseInstance(final String caseDefinitionName,
                                       final String owner,
                                       final List<CaseRoleAssignmentSummary> assignments) {
-        final CaseDefinitionSummary caseDefinition = caseDefinitions.get(caseDefinitionName);
-        if (caseDefinition == null) {
+        final Optional<CaseDefinitionSummary> cds = getCaseDefinitionSummary(caseDefinitionName);
```
bsremac: Can we rename _caseDefinitionName_ argument to _caseDefinitionId_ so it's consistent with naming of arguments in _loadCaseRoles()_ and _getCaseDefinitionSummary()_?

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstancePresenter.java
```diff
@@ -88,14 +83,11 @@ protected void loadCaseDefinitions() {
                         return;
                     }
 
-                    caseDefinitions = definitions.stream().collect(toMap(s -> s.getName(),
-                                                                         s -> s));
+                    caseDefinitions = definitions;
+                    caseDefinitions.sort(comparing(d -> d.getName()));
```
cristianonicolai: done, I enforced the sorting of the results from the kie-server.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/util/Select.java
cristianonicolai: Although it is not used at the moment I would prefer to keep it as is. This is a general UI component that I would expect to be used in different context, so exposing the relevant attributes here.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
cristianonicolai: done

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java
```diff
@@ -151,13 +153,24 @@ public void clearRoles() {
...
     @Override
-    public void setCaseDefinitions(final List<String> definitions) {
+    public void setCaseDefinitions(final List<CaseDefinitionSummary> definitions) {
         clearCaseDefinitions();
         caseTemplatesList.setValue("");
-        for (final String definition : definitions) {
-            caseTemplatesList.addOption(definition);
-        }
+        definitions.forEach(d -> {
+            final Optional<CaseDefinitionSummary> cds = definitions.stream().filter(dn -> dn.getName().equals(d.getName()) && dn.getUniqueId().equals(d.getUniqueId()) == false).findFirst();
```
jhrcek: This condition seems rather confusing:
`dn.getName().equals(d.getName()) && dn.getUniqueId().equals(d.getUniqueId()) == false`

It took me a while to get the point :) (It's distinguishing case definition names which determine CD uniquely from the names that have more different CDs associated, right?)

Apart from readability I see other problem with this way of checking "which name identify the case uniquely" and that is quadratic complexity. For each case definition all other case definitions are checked. E.g. if you have 100 case definitions it will check 100x100 pairs of definitions etc.

The same functionality can be achieved by single linear pass through the list of definitions and aggregating them into the map by name. Like:
```
Map<String,List<Definitions> nameToDefinitions = definitions.stream().collect(Collectors.groupingBy(CaseDefinitionSummary::getName));

nameToDefinitions.entrySet().forEach(entry -> {
  if (entry.value().size() > 1) {
     // add disambiguated case def.
  } else {
     //add uniquely identified case def.
  }
});
```
@cristianonicolai WDYT?

cristianonicolai: @jhrcek  definitely a nice improvement, done.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-showcase/src/main/java/org/jbpm/workbench/cm/server/MockCaseManagementService.java
```diff
@@ -72,21 +72,26 @@
...
+            caseMilestoneList = readJsonValues(CaseMilestoneSummary.class, CASE_MILESTONES_JSON);
+            caseStageList = readJsonValues(CaseStageSummary.class, CASE_STAGES_JSON);
+            caseActionList = readJsonValues(CaseActionSummary.class, CASE_ACTIONS_JSON);
+            processDefinitionList = readJsonValues(ProcessDefinitionSummary.class, PROCESS_DEFINITION_JSON);
             LOGGER.info("Loaded {} case definitions", caseDefinitionList.size());
+    }
+
+    private <T> List<T> readJsonValues(final Class<T> type, final String fileName){
+        try {
+            final InputStream inputStream = Thread.currentThread().getContextClassLoader().getResourceAsStream(fileName);
```
triceo: FYI the stream is not properly closed here. If put inside `try ()`, this will be fixed.

cristianonicolai: @triceo done, thanks.

bsremac:
+1

PR #672, opened 2017-04-05
Case Stages - Remove empty message

cristianonicolai:


bsremac:
The message wasn't much useful indeed :)
+1

PR #676, opened 2017-04-06
Review task logs panel details

cristianonicolai:
Use list to handle items instead of building the html manually
Remove label string
Fix and reformat test code


bsremac:
Regarding  test code reformatting, should this type of formatting be our new practice from now on?

cristianonicolai:
@bsremac good question, to be honest, I was reluctant to do it because it creates a lot of unnecessary noise in the PR. Perhaps we should focus on either re-formating only PRs and making sure that new/changed code are properly formatted.

bsremac:
@cristianonicolai It makes sense to me that only new/re-factored code should be subject to this type of formatting.
Maybe also makes sense in case of re-factored PR to always mention this type of change in the PR, as you did for this one.
+1

PR #679, opened 2017-04-07
JBPM-4647: Case Roles. Redesign

nmirasch:
Case roles assignment management Building block


nmirasch:
@cristianonicolai trying to take the case roles redesign commit to  7.0.x branch.

PR #682, opened 2017-04-11
JBPM-4647: Case Roles. Redesign (#669)

nmirasch:


nmirasch:
@cristianonicolai cherrypicked the case roles implementation to 7.0.x branch

PR #683, opened 2017-04-12
JBPM-5864 - Allow start case instance to load case definitions with s…

cristianonicolai:
…ame id and name from different containers (#671)

Add container name as sub-text in case there is multiple cases with same name.
Minor improvements to loading JSON mock data


cristianonicolai:
Pushing changes from https://github.com/kiegroup/jbpm-wb/pull/671 into 7.0.x branch.

PR #692, opened 2017-04-18
JBPM-5695: Exception in Task list for restarted Kie server

nmirasch:
Catch KieServicesHttpException thrown when the kie-server still is not ready to execute dataset queries while it's registering dataSetDefinitions


nmirasch:
@cristianonicolai Just leave the ServerTemplateSelector reaction to the events fired  when a Server Template is deleted or when  a ServerInstanceRegistered ends with the  KieServerDataSetRegistered completion.
The removed ServerTemplateUpdated event is launched at the connect  and disconnect kie-server actions, both covered with the proposed approach. what do you think?

cristianonicolai:
@nmirasch  the only issue I can think of is the case when a new server template is added but there is not kie server instance connect to it. In that case, this server template would not show up in the UI. Although there are no server connected I guess we should try to keep consistency with the server management view. One way I can think to get around that is to actually test in the ServerTemplateUpdated event how many instance are connected, if there is at least one you could ignore the event and just wait for the KieServerDataSetRegistered event?

nmirasch:
@cristianonicolai  added the server template selector update when the  ServerTemplateUpdated event is fired with no instances. If that event brings a server template with instances this method does nothing, and the server template selector will be updated when the KieServerDataSetRegistered event indicates that the datatasets have been registered and ready to use.

cristianonicolai:
@nmirasch looks good, thanks.

PR #694, opened 2017-04-20
JBPM-5896: Case Roles - Only display Unassigned label if there are no…

nmirasch:
… assignments for the current role

### DIFF
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.html
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.html
@@ -24,6 +24,10 @@
                         <span class="fa fa-users"></span>
                         <div data-field="line-groups"></div>
                     </div>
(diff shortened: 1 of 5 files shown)


nmirasch:
@cristianonicolai To be consistent with this requirement I  have changed the two unassigned labels that appeared for users and roles, and have unified in one 'unassigned' label linked with the role,( not sure about the user icon that is shown there).  If you think that is a good idea to show the old two labels, I can revert that part of the PR. Could you take a look please? thanks!!

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/roles/CaseRoleItemView.html
```diff
@@ -24,6 +24,10 @@
                         <span class="fa fa-users"></span>
                         <div data-field="line-groups"></div>
                     </div>
+                    <div data-field="div-unassigned" class="hidden kie-full-width kie-single-role-assignment-height kie-unassigned-color ">
+                        <span class="fa fa-user"></span>
+                        <span data-i18n-key="Unassigned"></span>
```
cristianonicolai: please add a small space between the icon and the string, adding _padding-left: 5px;_ looks good.

nmirasch:
@cristianonicolai done, Added a 8px to have the same space than the one that have the assigned ones.

PR #696, opened 2017-04-21
JBPM-5896: Case Roles - Only display Unassigned label if there are no…

nmirasch:
… assignments for the current role (#694)


nmirasch:
Cherry pick change #694  into 7.0.x branch

PR #702, opened 2017-04-24
BPMSPL-632 & JBPM-5918: fixed sorting on last update column

rbarriuso:
Unified last modification date or process instance end date at SQL level.


cristianonicolai:
ok to test

PR #706, opened 2017-04-25
Fix database tests

cristianonicolai:
Add ExecutionErrorInfo class to JPA persistence unit


mswiderski:
thanks for adding that @cristianonicolai one question though - do we still need those tests in there? since entire backend/execution engine is outside of workbench...

cristianonicolai:
@mswiderski I think those tests are still useful, they check that any dataset query that will be registered into the kie server is valid. There is also a very common issue regarding sintax across the different databases which we could look into in the future.

mswiderski:
ok, fair enough @cristianonicolai , thanks

PR #715, opened 2017-05-19
RHBPMS-4712: Only 15 workitemhandlers can be added to kie-deployment-…

nmirasch:
…descriptor.xml  via business central console
Added the  unlimited page size at the table constructor to avoid to use the defaultPageSize that is 15.


cristianonicolai:
@nmirasch fix looks good but I also noticed that the parameters popup also has a limitation of 15 items, please take a look here: https://github.com/kiegroup/jbpm-wb/blob/4109ee74290f836a2041a07d91eedc82be607a22/jbpm-wb-integration/jbpm-wb-integration-client/src/main/java/org/jbpm/workbench/wi/client/editors/deployment/descriptor/DDParametersPopup.java#L65

nmirasch:
@cristianonicolai  could you take a look? thanks!

jbpm-console-ng-workbench-integration/jbpm-console-ng-workbench-integration-client/src/main/java/org/jbpm/console/ng/wi/client/editors/deployment/descriptor/DDParametersPopup.java
```diff
@@ -60,7 +60,7 @@
     @UiField
     FormGroup parametersGroup;
 
-    @UiField
+    @UiField(provided=true)
     CellTable<Parameter> parametersTable;
```
cristianonicolai: perhaps initialise the object in here and reuse the same constant for the page size?

nmirasch:
@cristianonicolai changed

PR #716, opened 2017-05-22
[do not merge]  - JBPM-5990 - Process instance list advanced search

cristianonicolai:


cristianonicolai:
@nmirasch @rbarriuso @bsremac preview of the new advanced search, this still work in progress but I think it can already give an idea of how is going to work. Feedback is welcome.
Known limitations:
* Filters work as AND expression but there isn't any logic to block the user from filtering multiple times with the same attribute.
* Missing filter by process
* Overflow for a large number of filters although it displays all filters this is not currently handled properly
* Missing date range filter
* State active should be an active filter when loading the search tab for the first time.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AdvancedSearchTable.java
```diff
@@ -0,0 +1,39 @@
...
+import org.uberfire.ext.services.shared.preferences.GridGlobalPreferences;
+
+public class AdvancedSearchTable<T extends GenericSummary> extends ExtendedPagedTable<T> {
+
+    private AdvancedSearchFiltersView advancedSearchFiltersView;
+
+    public AdvancedSearchTable(final GridGlobalPreferences gridPreferences) {
+        super(gridPreferences);
+        advancedSearchFiltersView = IOC.getBeanManager().lookupBean(AdvancedSearchFiltersView.class).getInstance();
+        topToolbar.add(advancedSearchFiltersView);
```
bsremac: I'm getting compilation error due to unknown variable _topToolbar_

cristianonicolai: @bsremac make sure you have this commit in your local uberfire version: https://github.com/AppFormer/uberfire/pull/712

PR #718, opened 2017-05-23
JBPM-5996: Click is not ignored in actions column on list views

rbarriuso:


ederign:
ok to test

cristianonicolai:
jenkins please retest this.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridView.java
```diff
@@ -399,5 +400,20 @@ public int getRefreshValue() {
     public void saveRefreshValue( int newValue ) {
         filterPagedTable.saveNewRefreshInterval( newValue );
     }
+    
+    public void addSelectionIgnoreColumn(Column<T, ?> column){
+        ignoreSelectionColumns.add(column);
+    }
+    
+    public boolean isSelectionIgnoreColumn(int colIx){
+        boolean ret = false;
```
cristianonicolai: I would suggest using a stream like alternative: .stream.filter.findFirst.isPresent or similar.

rbarriuso: Good idea

rbarriuso:
I've rebased, added tests and moved the "SelectionIgnoreColumn" functionality to the `ExtendedPagedTable` class, as I realized that it fits better there than in `AbstractMultiGridView`.

rbarriuso:
Applied functionality to `ProcessDefinitionListViewImpl`

PR #719, opened 2017-05-24
Updates to ProcessInstanceListViewImpl

cristianonicolai:
Removed unused code
Use GWT.create to initialize widgets so GWTMockito can properly mock classes for test
Renamed test class to mach impl


bsremac:
+1

PR #720, opened 2017-05-24
Updates to RequestListViewImpl

cristianonicolai:
Add enum for status instead of Strings
Renamed test class to mach impl


bsremac:
+1

PR #722, opened 2017-05-24
JBPM-5990 - Advanced search

cristianonicolai:
New search capabilities for Jobs, Process Instances and Tasks perspectives


cristianonicolai:
@nmirasch @rbarriuso @bsremac initial version for the advanced search for jobs, process instances and task perspectives. This still missing filter by process and date range which will come on separated PRs.
Closing https://github.com/kiegroup/jbpm-wb/pull/716 as it got mixed up with unrelated changes.

jhrcek:

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/AbstractTaskListPresenter.java
```diff
@@ -406,4 +409,74 @@ public void setTaskService(final Caller<TaskService> taskService) {
...
+                           v -> removeAdvancedSearchFilter(equalsTo(COLUMN_PROCESS_INSTANCE_DESCRIPTION,
+                                                                    v))
+        );
+
+        //TODO missing creation date
+
+        final FilterSettings settings = view.getAdvancedSearchFilterSettings();
+        final DataSetFilter filter = new DataSetFilter();
+        filter.addFilterColumn(equalsTo(COLUMN_STATUS,
+                                        TASK_STATUS_READY));
```
nmirasch: @cristianonicolai I think the default initial filter would be like the 'active' ones.. including :Ready, Reserved and InProgress status, do it make sense?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/requestlist/RequestListPresenter.java
```diff
@@ -235,4 +237,77 @@ public void requestCreated( @Observes RequestChangedEvent event ) {
...
+                             v -> removeAdvancedSearchFilter(equalsTo(COLUMN_STATUS,
+                                                                      v))
+        );
+
+        //TODO Missing process id and creation date
+
+        final FilterSettings settings = view.getAdvancedSearchFilterSettings();
+        final DataSetFilter filter = new DataSetFilter();
+        filter.addFilterColumn(equalsTo(COLUMN_STATUS,
+                                        RequestStatus.RUNNING.name()));
```
nmirasch: @cristianonicolai  the same for jobs, does it make sense to do the 'All' filter as default, I say this because maybe the user is used to access to job tab and see by the default the first tab, that is the 'All' one

cristianonicolai: @nmirasch that was my initial intention for both filters, the reason I only used one status is that the user can not create the same filter with multiple statuses (using OR) via the UI. I thought that would cause confusion. When adding new filters in the UI it always appends as an AND operation.

nmirasch:
@cristianonicolai  another doubt is about the domain columns. In this PR just came the task list (filter by name)  ,but later will happen with the process Instance id (filter by processId). My doubt is about if we have to prevent the addition of domain columns in the search tab, even when the column picker can grow a lot here, even the filter is removed, the columns will be still there. What do you think?

cristianonicolai:
@nmirasch good point, we can prevent adding domain columns for now.

PR #726, opened 2017-05-26
JBPM-5990 - Do not include domain specific columns for search tab res…

cristianonicolai:
…ults


cristianonicolai:
@nmirasch as suggested ;)

PR #731, opened 2017-05-29
JBPM-5990 - Add process name filter for perspectives using advanced search

cristianonicolai:


cristianonicolai:
Depends on https://github.com/AppFormer/uberfire/pull/724

cristianonicolai:
@nmirasch @rbarriuso @bsremac new dropwdown filter listing processes available for Jobs, Tasks, Tasks Admin and Process Instances perspectives.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/dataset/DataSetAwareSelect.java
```diff
@@ -0,0 +1,126 @@
...
+
+        try {
+            dataSetClientServices.lookupDataSet(ConsoleDataSetLookup.fromInstance(dataSetLookup,
+                                                                                  filterSettings.getServerTemplateId()),
+                                                new DataSetReadyCallback() {
+                                                    @Override
+                                                    public void callback(final DataSet dataSet) {
+                                                        select.refresh(s -> {
+                                                            s.removeAllOptions();
+                                                            for (int i = 0; i < dataSet.getRowCount(); i++) {
```
nmirasch: When there aren't process to show, do you think is a good idea disable Selector. I'm saying that because when there's no  process name to show a vertical line appears under the button (the dropdown with no results to show).
Does this make sense? I think we can leave this kind of improvements for later, this comment is just to be sure.

cristianonicolai: @nmirasch makes sense to me.

PR #732, opened 2017-05-29
JBPM-5990 - Add date range filter attributes for perspectives using a…

cristianonicolai:
…dvanced search

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AdvancedSearchFiltersViewImpl.java
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AdvancedSearchFiltersViewImpl.java
@@ -17,6 +17,7 @@
 package org.jbpm.workbench.common.client.list;
 
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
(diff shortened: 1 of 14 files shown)


jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AdvancedSearchFiltersViewImpl.java
```diff
@@ -115,11 +125,77 @@ public void addNumericFilter(final String label,
...
+        if (toDate.getValue() == null || fromDate.getValue() == null) {
+            return;
+        }
+        final Date toDateValue = UTCDateBox.utc2date(toDate.getValue());
+
+        final Date fromDateValue = UTCDateBox.utc2date(fromDate.getValue());
+        final DateRange dateRange = new DateRange(fromDateValue,
+                                                  toDateValue);
+        addActiveFilter(label,
+                        translationService.format(Constants.FROM)
```
rbarriuso: I'm curious: why translating via `translationService` instead of `...resources.i18n.Constants.INSTANCE.From()`?

cristianonicolai: @rbarriuso the general guideline is to use Errai api as much as possible to minimize the refactor for when we move to GWT 3. Not entirely sure if the translation/messages will be affected, but widgets in general are expected to be removed.

rbarriuso: Ok, good to know. Thanks!

PR #734, opened 2017-05-30
JBPM-5990 - Disable process dropdown when no items available

cristianonicolai:


cristianonicolai:
@bsremac @nmirasch as suggested.

PR #737, opened 2017-05-30
JBPM-5990 - Use dataset API like filtering for text field search

cristianonicolai:


cristianonicolai:
@nmirasch @bsremac @rbarriuso  alternative solution for text filtering. Using the dataset like API we can allow searches to include wildcards as specified in here: https://github.com/dashbuilder/dashbuilder/blob/79152058ea1878e004768d04024cda6e0e779b53/dashbuilder-shared/dashbuilder-dataset-api/src/main/java/org/dashbuilder/dataset/filter/CoreFunctionType.java#L34
 A simple string will work as equals but appending any combination of wildcards should allow more powerful filtering. It does need some hint in the UI for the user to be able to be aware of how it works.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/requestlist/RequestListPresenter.java
```diff
@@ -254,26 +254,32 @@ public void setupAdvancedSearchView() {
 
         view.addTextFilter(constants.BusinessKey(),
                            constants.FilterByBusinessKey(),
-                           v -> addAdvancedSearchFilter(equalsTo(COLUMN_BUSINESSKEY,
-                                                                 v)),
-                           v -> removeAdvancedSearchFilter(equalsTo(COLUMN_BUSINESSKEY,
-                                                                    v))
+                           v -> addAdvancedSearchFilter(likeTo(COLUMN_BUSINESSKEY,
```
nmirasch: @cristianonicolai  I think we have to add "%" + v'+ "%" to do the likeTo really useful and different from the equalsTo

jhrcek:
@cristianonicolai if we go for enabling wildcard search, we should make this somehow apparent in the UI, otherwise this becomes another feature that only developers are aware of. How about adding small question mark hint next to the search box or somewhere visibly that would explain what kind of wildcards are supported?

cristianonicolai:
@jhrcek sure, I think that is important. As I mentioned it needs to be some help in the UI explaining to the user how it works. Preferably I would like to do another PR just to handle that if it is ok.

jhrcek:
@cristianonicolai +1 for another PR. We'll add it to the doc so that it's not forgotten.

bsremac:
@cristianonicolai Looks powerful indeed. Only concern can be filtering performance on bigger data sets, but this can't be evaluated at this point. 
+1

PR #741, opened 2017-06-01
JBPM-6024: Move process instance error marker popover to Errai template

rbarriuso:


jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/list/ProcessInstanceSummaryErrorPopoverCell.java
```diff
@@ -0,0 +1,123 @@
...
+                });
+            function onViewErrorsClick(){
+                var processInstId = $wnd.jQuery(this).attr(procIdAttrName);
+                thisCellRef.@org.jbpm.workbench.pr.client.editors.instance.list.ProcessInstanceSummaryErrorPopoverCell::openErrorView(Ljava/lang/String;)(processInstId);
+            }
+        });
+    }-*/;
+    
+    private String sanitizeAttributeHtml(String html){
+        return html.replace("\"",  "'");
```
rbarriuso: Perhaps we could also remove newline characters. Wdyt @cristianonicolai @nmirasch ?

cristianonicolai: @rbarriuso I think data-content can actually be a function, that way you could simply call a method back in the class to get the inner HTML. That way you wouldn't need any manipulation in the html String, right?

cristianonicolai:
ok to test

cristianonicolai:
@rbarriuso this looks really good to me. My only comment would be regarding the Popover component, if we could try to reuse that in here and extend to the need you have.

rbarriuso:
@cristianonicolai The requested changes are made

bsremac:
@rbarriuso One small remark.
When one hovers over errors number, the popover with the link appears but then disappears immediately after the mouse pointer exits the number area. That makes sense if the mouse pointer is not moved to the popover. In this case, I would expect that the popover stays active and allow the user to click _View All_ link.

On the other hand, when one hovers over errors number, the mouse pointer changes, indicating clickable element, which makes me think that the whole popover is completely redundant in this case. The popover offers only one link and states the number of errors which is already exposed. So why not simply integrating the link when one clicks on the errors number?

rbarriuso:
@bsremac Regarding the first comment, I totally agree, but that's the behavior of the current implementation of [the Popover widget on Patternfly](https://www.patternfly.org/pattern-library/widgets/#popover). To make the popover stay the user needs to click.

Regarding the second comment, I'll change the mouse pointer when it's not really clickable

rbarriuso:
@bsremac Mouse pointer changed

bsremac:
+1

PR #743, opened 2017-06-03
JBPM-5990: Load user grid stored preferences.

nmirasch:


jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridView.java
```diff
@@ -249,9 +248,23 @@ public String getValidKeyForAdditionalListGrid( String baseName ) {
...
     }
 
-    protected ExtendedPagedTable<T> createExtendedPagedTable(final GridGlobalPreferences preferences,
+    protected ExtendedPagedTable<T> createExtendedPagedTable(final GridGlobalPreferences defaultPreferences,
                                                              final String key) {
-        return TAB_SEARCH.equals(key) ? createAdvancedSearchTable(preferences) : new ExtendedPagedTable<T>(preferences);
+        GridGlobalPreferences pref;
+        ExtendedPagedTable<T> table ;
+        if(TAB_SEARCH.equals(key)){
+            pref = new GridGlobalPreferences(defaultPreferences.getKey() + key,
```
jhrcek: The `if` and `else` branch have the same code for initializing `pref` - why not initialize it before if?

nmirasch: you are absolutely right, it' s only necessary  add the defaultPreferences.getKey() to the grid preference key at  the TAB_SEARCH, to avoid collision with other view  Tab search stored preferences.
@jhrcek  changed. Can you take a look? thanks

jhrcek: I actually meant moving pref initialization before the if block to remove some code duplication like this:


```
GridGlobalPreferences pref = new GridGlobalPreferences(defaultPreferences.getKey() + key,
                                             defaultPreferences.getInitialColumns(),
                                             defaultPreferences.getBannedColumns());

if (TAB_SEARCH.equals(key)){
    table = createAdvancedSearchTable(pref);
} else {
    table = createAdvancedSearchTable(pref);
}

```

But if you think it's better the way you changed it I'm ok with that:)

PR #744, opened 2017-06-05
Enable findbugs.failOnViolation

jhrcek:
Enable findbugs-maven-plugin to fail the build when serious bug has been identified. Findbugs plugin is only executed when the build is run with `-Dfull` (which is the case in kie-jenkins CI jobs).

At the same time I'm providing fixes to 2 issues identified by the plugin.

The configuration of the plugin can be found in [droolsjbpm-build-bootstrap/pom.xml](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L1165-L1182) and has been already fine-tuned by QE to only fail the build in cases where it's almost certain there's a bug.

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/UTCTimeBoxImplShared.java
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/UTCTimeBoxImplShared.java
@@ -28,22 +28,22 @@ public abstract class UTCTimeBoxImplShared extends Composite implements UTCTimeB
     protected DateTimeFormat timeFormat;
 
     // ----------------------------------------------------------------------
-    
+
     /**
      * Sets the DateTimeFormat for this UTCTimeBox. The HTML5
      * implementation will ignore this.
      */
     @Override
-    public void setTimeFormat(DateTimeFormat timeFormat) { 
+    public void setTimeFormat(DateTimeFormat timeFormat) {
         this.timeFormat = timeFormat;
     }
-    
+
     /**
      * Sets the visible length of the time input. The HTML5
      * implementation will ignore this.
      */
     @Override
-    public void setVisibleLength(int length) {}    
+    public void setVisibleLength(int length) {}
 
     /**
      * Validates the value that has been typed into the text input.
@@ -51,7 +51,7 @@ public abstract class UTCTimeBoxImplShared extends Composite implements UTCTimeB
      */
     @Override
     public void validate() {}
-    
+
 
     /**
      * Sets the time value (as milliseconds since midnight independent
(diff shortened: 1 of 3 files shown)


jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/UTCTimeBoxImplShared.java
```diff
@@ -60,26 +60,28 @@ public void validate() {}
...
+
     // ----------------------------------------------------------------------
-    
+
     public UTCTimeBoxImplShared() {
         if (fallbackTimeFormats == null) {
-            fallbackTimeFormats = new DateTimeFormat[fallbackFormatStrings.length];
-            for (int i=0; i<fallbackFormatStrings.length; i++) {
-                fallbackTimeFormats[i] = DateTimeFormat.getFormat(fallbackFormatStrings[i]);
+            // Initialize the array fully in local var before assigning to class field to avoid synchronization issues
```
jhrcek: findbugs plugin warning to clarify this fix:

UTCTimeBoxImplShared.java:68, LI_LAZY_INIT_UPDATE_STATIC, Priority: High

Incorrect lazy initialization and update of static field org.jbpm.workbench.common.client.util.UTCTimeBoxImplShared.fallbackTimeFormats in new org.jbpm.workbench.common.client.util.UTCTimeBoxImplShared()

This method contains an unsynchronized lazy initialization of a static field. After the field is set, the object stored into that location is further updated or accessed. The setting of the field is visible to other threads as soon as it is set. If the futher accesses in the method that set the field serve to initialize the object, then you have a very serious multithreading bug, unless something else prevents any other thread from accessing the stored object until it is fully initialized.

Even if you feel confident that the method is never called by multiple threads, it might be better to not set the static field until the value you are setting it to is fully populated/initialized.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
```diff
@@ -105,7 +105,7 @@ public ProcessInstanceSummary getProcessInstance(String serverTemplateId, Proces
             if(nis.getNodeType().equals("HumanTaskNode")){
                 logs.add(new RuntimeLogSummary(nis.getId(), prettyDateFormatter.format(nis.getDate()), "Task '" + nis.getName() + "' was created", "System"));
                 for(TaskEventInstance te : allTaskEventsByProcessInstanceId){
-                    if(te.getWorkItemId() != null && nis.getId() == te.getWorkItemId() &&
+                    if(te.getWorkItemId() != null && te.getWorkItemId().equals(nis.getId()) &&
```
jhrcek: This and the following methods compare two reference values using the == or != operator, where the correct way to compare instances of this type is generally with the equals() method.

jhrcek:
Hello @cristianonicolai could you please review/merge?

cristianonicolai:
@jhrcek thanks!

PR #749, opened 2017-06-09
Add error counter advanced search filter to process instance list

rbarriuso:
This is a proposal to add a true / false advanced search filter to list process instances with or without unacknowledged errors present


cristianonicolai:
ok to test

cristianonicolai:
jenkins please retest this.

PR #753, opened 2017-06-12
JBPM-6064 - Navigation links

cristianonicolai:
Navigation between process definition and instances
Navigation between process instance and errors


cristianonicolai:
@nmirasch @rbarriuso tests added, ready for review ;)

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridPresenter.java
```diff
@@ -32,6 +32,9 @@
 
 public abstract class AbstractMultiGridPresenter<T extends GenericSummary, V extends MultiGridView> extends AbstractScreenListPresenter<T> implements RefreshSelectorMenuBuilder.SupportsRefreshInterval {
 
+    public static final String SEARCH_PARAMETER_PROCESS_INSTANCE_ID = "processInstanceId";
+    public static final String SEARCH_PARAMETER_PROCESS_DEFINITION_ID = "processDefinitionId";
```
rbarriuso: Perhaps the search parameter IDs could be defined in `PerspectiveIds`, with the IDs of the perspectives which receive them as parameters. What do you think?

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridPresenter.java
```diff
@@ -86,33 +100,43 @@ protected ListView getListView() {
 
     public abstract void setupAdvancedSearchView();
 
+    public abstract void setupActiveSearchFilters();
```
rbarriuso: Perhaps a default implementation of `setupActiveSearchFilters()` could only call `setupDefaultActiveSearchFilters()` and let the child presenters decide if it's necessary to override `setupActiveSearchFilters()`

cristianonicolai: done

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridPresenter.java
cristianonicolai: @rbarriuso I agree we need a better place to keep this, although `PerspectiveIds` doesn't seem the right place to me as we share these parameters only across the runtime views. Another point is that I believe the same parameter could be reused across different screens, so appending the Perspective Id wouldn't be ideal. I end up extending the `PerspectiveIds`, would that be a better approach?

rbarriuso: Yes, I think now it's better. Thanks

rbarriuso:
+1

PR #754, opened 2017-06-13
JBPM-6011 - Avoid injecting DataSetEditorManager twice

cristianonicolai:


cristianonicolai:
@nmirasch I think I found the reason why we're getting two instances, DataSetEditorManager was injected twice, in the view and presenter, causing multiple instances of FilterEditorPopup.

PR #755, opened 2017-06-13
JBPM-6090: Navigation from jobs to process instances and vice versa

rbarriuso:
Depends on https://github.com/kiegroup/jbpm-wb/pull/753


jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/list/ProcessInstanceListViewImpl.java
```diff
@@ -429,32 +428,42 @@ public ProcessInstanceSummary getValue(ProcessInstanceSummary process) {
...
                 }
             }
-        } ) );
+        ));
 
-        CompositeCell<ProcessInstanceSummary> cell = new CompositeCell<ProcessInstanceSummary>( cells );
-        Column<ProcessInstanceSummary, ProcessInstanceSummary> actionsColumn = new Column<ProcessInstanceSummary, ProcessInstanceSummary>( cell ) {
+        cells.add(new ProcessInstanceSummaryActionCell(
+            constants.ViewJobs(),
+            ProcessInstanceSummaryActionCell.PREDICATE_TRUE,
```
rbarriuso: @cristianonicolai Any recommendation about how to implement the condition so the "View jobs" button only is displayed if the process instance has jobs? Should I use a dataset lookup to check the jobs associated to each process instance?

rbarriuso:
@nmirasch @cristianonicolai Here's an initial implementation of the navigation between process instances and jobs. Please feel free to review while https://github.com/kiegroup/jbpm-wb/pull/753 is merged (then I'll rebase)

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/list/ProcessInstanceListViewImpl.java
cristianonicolai: @rbarriuso I'm wondering if we should have the button always present without a data/number check. I'm concerned about adding yet another join to the dataset only to enable/disable this button. The drawback would be that the user could try to check Jobs for processes that don't have any. wdyt?

bsremac: Having the button disabled in that case sounds like the cleanest approach, but I understand your concern @cristianonicolai.
If we leave the button enabled, the only logical approach would be to show empty list in Jobs perspective. Everything else would be just too confusing.

rbarriuso: @bsremac Exactly, that's the current behavior: the button is always enabled and allows the user to navigate to the jobs view, but the job list may be empty if the selected process instance has no associated jobs

rbarriuso:
Rebased

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/requestlist/RequestListViewImplTest.java
```diff
@@ -181,5 +196,65 @@ public void initDefaultFiltersOwnTaskFilter() {
...
+        SafeHtmlBuilder cellHtmlBuilder = mock(SafeHtmlBuilder.class);
+        doAnswer(invocationOnMock -> {
+            invocationOnMock.callRealMethod();
+            verify(cellHtmlBuilder, times(shouldRender ? 1 : 0)).append(any());
+            return null;
+        }).when(cellMock).render(any(), any(), eq(cellHtmlBuilder));
+
+        cellMock.render(cellContext, val, cellHtmlBuilder);
+
+        verify(cellMock).render(cellContext, val, cellHtmlBuilder);
```
bsremac: What is the difference between the above method body and the following one:
```java
        RequestListViewImpl.ActionHasCell cellMock = spy(cell);
        SafeHtmlBuilder cellHtmlBuilder = mock(SafeHtmlBuilder.class);
        cellMock.render(cellContext, val, cellHtmlBuilder);
        verify(cellHtmlBuilder, times(shouldRender ? 1 : 0)).append(any());
```

rbarriuso: @bsremac I think both implementations are valid, though yours is shorter. I'll replace it.

rbarriuso: @bsremac PR including this change at https://github.com/kiegroup/jbpm-wb/pull/766

PR #760, opened 2017-06-15
JBPM-6112 - Make content in Source tab of Deployment Descriptor editor read-only

tsurdilo:


tsurdilo:
@mswiderski could you please review when you have the chance? thanks

mswiderski:
Could we make the force tab to be read only instead of removing it? I see a value in being Ankę to ser the structure of this file. Though if this is too complex then we can remove it

tsurdilo:
@mswiderski yep thats better idea. source tab content is read-only now. sorry about all the formatting changes :)

PR #765, opened 2017-06-20
JBPM-5990: Replace active filter when selecting new value

cristianonicolai:


cristianonicolai:
@bsremac @nmirasch as discussed, whenever a new filter is added, if there is an active filter for the same column, it gets replaced.

jbpm-wb-common/jbpm-wb-data-filters-client/src/test/java/org/jbpm/workbench/df/client/filter/FilterSettingsTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+        settings.addColumnFilter(filter);
+        settings.removeColumnFilter(filter);
+
+        assertEquals(0,
+                     settings.getDataSetLookup().getFirstFilterOp().getColumnFilterList().size());
+        assertFalse(settings.getDataSetLookup().getFirstFilterOp().getColumnFilterList().contains(filter));
+    }
+
+}
```
bsremac: I'd merge these two tests into one (_testAddRemoveColumnFilter_).
For example, add two filters, verify both are are there, remove one, verify only one is left.

The reason I'm proposing this is that the second test (_testRemoveColumnFilter_) can be false positive, if adding a filter fails.

cristianonicolai: @bsremac sure, changes done.

bsremac:
+1

PR #779, opened 2017-06-27
JBPM-6139: Wrap active filter list when overflow

nmirasch:


cristianonicolai:
@nmirasch wrap looks fine up to certain width, if the screen gets smaller they fall out the grey panel. IS that something we could prevent? Take a look at the screenshots below.

selection_123

selection_122

nmirasch:
@cristianonicolai changed a bit the active filter styles, could you take a look? Thanks!

cristianonicolai:
@nmirasch perfect, thanks

PR #787, opened 2017-07-02
JBPM-6166: Execution Error view. Hide by default the error Id and add…

nmirasch:
… Job type


jbpm-wb-executor-service/jbpm-wb-executor-service-api/src/main/java/org/jbpm/workbench/es/model/events/ExecErrorSelectionEvent.java
```diff
@@ -16,22 +16,37 @@
 package org.jbpm.workbench.es.model.events;
 
 import org.jboss.errai.common.client.api.annotations.Portable;
+import org.jbpm.workbench.es.model.ExecutionErrorSummary;
 
 @Portable
 public class ExecErrorSelectionEvent {
```
cristianonicolai: @nmirasch now that you're making changes here, would ou mind changing this class name to `ExecutionErrorSelectedEvent`  ? These events can also be moved into the client module as they are only used in there. 
Another note in the relation to error events, `ExecErrorSearchEvent` isn't in use, could be removed. `ExecErrorChangedEvent` is only sued in the presenter to refresh the grid, I would recommend deleting it too.

jbpm-wb-executor-service/jbpm-wb-executor-service-api/src/main/java/org/jbpm/workbench/es/model/events/ExecErrorSelectionEvent.java
```diff
@@ -16,22 +16,37 @@
...
     public ExecErrorSelectionEvent() {
     }
 
     public ExecErrorSelectionEvent(String serverTemplateId,
+                                   ExecutionErrorSummary errorSummary
+                                   ) {
+        this(serverTemplateId,
+             errorSummary.getDeploymentId(),
+             errorSummary.getErrorId(),
+             errorSummary.getProcessId() +
```
cristianonicolai: I think building the title String should be a responsibility of the presenter and not the event itself. I would encourage to leave the event to only hold the necessary data without much logic in it.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenterTest.java
```diff
@@ -362,6 +365,35 @@ public void testDefaultAdvancedSearchFilterTypes() {
...
+        assertEquals(deploymentId,
+                     captor.getValue().getDeploymentId());
+        assertEquals(errorId,
+                     captor.getValue().getErrorId());
+        assertEquals(errorSummary.getProcessId() +
+                             " - " + errorSummary.getProcessInstanceId() +
+                             " (" + errorSummary.getDeploymentId() + ")",
+                     captor.getValue().getErrorDetailsTitle());
+    }
+
```
bsremac: @nmirasch Can we add one test which asserts all the columns which should be shown by default?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenter.java
```diff
@@ -243,32 +231,27 @@ public Menus getMenus() {
...
             placeManager.closePlace(ExecutionErrorDetailsPresenter.SCREEN_ID);
         }
     }
 
-    public void onExecutionErrorSelectionEvent(@Observes ExecErrorWithDetailsRequestEvent event) {
-        placeManager.goTo(ExecutionErrorDetailsPresenter.SCREEN_ID);
-        execErrorSelectionEvent.fire(new ExecErrorSelectionEvent(event.getServerTemplateId(),
-                                                                 event.getDeploymentId(),
-                                                                 event.getErrorId()));
+    private String getErrorDetailTitle(ExecutionErrorSummary summary){
```
cristianonicolai: @nmirasch I think this should be done in the presenter that receives the event (`ExecutionErrorDetailsPresenter`). That presenter is the one responsible for building the UI data, the event would only have the necessary ids that identify the error that was selected.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/errordetails/ExecutionErrorDetailsPresenter.java
```diff
@@ -59,6 +59,8 @@
     private String serverTemplateId;
     private String deploymentId;
     private String errorId;
+    private String errorDetailsTitle;
```
cristianonicolai: Not sure if there is any need for holding the title state, perhaps remove this.

nmirasch:
@bsremac @cristianonicolai now the error detail title is generated at ExecutionErrorDetailsPresenter. Added a very basic test on the view about the default columns. Could you take a look? Thanks!

PR #788, opened 2017-07-02
JBPM-6168: Add view task action to execution error list for the error…

nmirasch:
… type Task

### DIFF
--- a/jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenter.java
+++ b/jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenter.java
@@ -194,10 +194,24 @@ public class ExecutionErrorListPresenter extends AbstractMultiGridPresenter<Exec
         return pis -> isUserAuthorizedForPerspective(PerspectiveIds.PROCESS_INSTANCES) && pis.getProcessInstanceId() != null;
     }
 
+    public Predicate<ExecutionErrorSummary> getViewTaskActionCondition() {
+        return pis -> ((isUserAuthorizedForPerspective(PerspectiveIds.TASKS_ADMIN) ||
+                isUserAuthorizedForPerspective(PerspectiveIds.TASKS)) &&
+                pis.getActivityId() != null  && ExecutionErrorType.TASK.getType().equals(pis.getType().getType()));
+    }
+
     public void goToProcessInstance(final ExecutionErrorSummary errorSummary) {
         navigateToPerspective(PerspectiveIds.PROCESS_INSTANCES,
                               PerspectiveIds.SEARCH_PARAMETER_PROCESS_INSTANCE_ID,
                               errorSummary.getProcessInstanceId().toString());
+    }
+
+    public void goToTask(final ExecutionErrorSummary errorSummary) {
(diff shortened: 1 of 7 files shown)


bsremac:
@nmirasch Can a test be added which tests that the action is only available in the Execution errors perspective when the error has Task type and that the action is not available with other two types?

bsremac:
@nmirasch @cristianonicolai Does this mean that the column with the number of errors will be added to Tasks Lists too and that it will be possible to navigate from Tasks perspective to Execution errors perspective?

cristianonicolai:
@bsremac at this stage only navigating from error to task.

cristianonicolai:
@rbarriuso see @bsremac's comment, would that be of interest for the task admin list?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenter.java
```diff
@@ -194,25 +194,38 @@ public void goToJob(ExecutionErrorSummary errorSummary) {
...
+    public Predicate<ExecutionErrorSummary> getViewJobActionCondition() {
         return pis -> isUserAuthorizedForPerspective(PerspectiveIds.JOBS) && pis.getJobId() != null;
     }
 
-    public Predicate<ExecutionErrorSummary> getViewProcessInstanceActionCondition(){
+    public Predicate<ExecutionErrorSummary> getViewProcessInstanceActionCondition() {
         return pis -> isUserAuthorizedForPerspective(PerspectiveIds.PROCESS_INSTANCES) && pis.getProcessInstanceId() != null;
     }
 
+    public Predicate<ExecutionErrorSummary> getViewTaskActionCondition() {
```
cristianonicolai: @nmirasch could you add some tests to verify the condition works as expected?

rbarriuso:
@cristianonicolai Currently there's no requirement about adding the error count column to the task admin list, but perhaps it would be a nice-to-have feature in the future

nmirasch:
@bsremac @cristianonicolai added some test and rebased. Could you mind take a look? Thanks!

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenterTest.java
```diff
@@ -341,6 +343,48 @@ public void testViewProcessInstanceActionCondition() {
...
+    @Test
+    public void testViewTaskActionCondition() {
+        doAnswer(new PerspectiveAnswer(TASKS_ADMIN)).when(authorizationManager).authorize(any(ResourceRef.class),
+                                                                                                eq(identity));
+
+        assertTrue(presenter.getViewTaskActionCondition().test(ExecutionErrorSummary.builder()
+                                                                       .type(ExecutionErrorType.TASK)
+                                                                       .activityId(Long.valueOf(1)).build()));
+
+        when(authorizationManager.authorize(any(ResourceRef.class),
```
cristianonicolai: @nmirasch because the authorization is denied from this point on, it invalidates the other tests you have like a different error type or error without `activityId`. You could either move those two tests before this point or split that in another test method.

nmirasch:
@cristianonicolai absolutely right, thanks! test updated.

PR #791, opened 2017-07-05
JBPM-6159 - Ensure search tab is always present independently from st…

cristianonicolai:
…ored grids

Search tab is always added as the first tab
Refactor and combined AbstractMultiGridView tests using shared parent class

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridView.java
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridView.java
@@ -17,6 +17,7 @@ package org.jbpm.workbench.common.client.list;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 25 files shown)


cristianonicolai:
@nmirasch @bsremac please take a look.

Here is how I tested this, start an older workbench version (7.0 for instance), create/remove a few search tabs from the runtime views (jobs, tasks, process instances). Deploy a new version using this fix, this will cause the workbench to start using the old preferences on the `.niogit` folder. As expceted results, you should see the new Search tab in all the views ( as the first tab ) plus any of the others left from your changes. Also, the Admin tab from task list should not be present (moved to Task Admin List).

cristianonicolai:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/467

bsremac:

bsremac:
@cristianonicolai 
Tested it and didn't find any issues.
To what you tried I also tried having the custom filter list spreading to next line - Search tab takes the first place and moves other filters to the left.
Deleting all filters, thus leaving the perspective empty - Search tab gets added along with all default filters restored.
Admin tab is not anymore part of Task List no matter what I tried to somehow trick it :)
+1

PR #793, opened 2017-07-06
BPMSPL-633: Improved UX of error count popover

rbarriuso:
Improves UX of process instance error count popover by managing visibility manually


ederign:
ok to test

PR #795, opened 2017-07-07
JBPM-6098 - Handle filter settings without key set

cristianonicolai:


jbpm-wb-common/jbpm-wb-common-client/src/test/java/org/jbpm/workbench/common/client/dataset/DataSetAwareSelectTest.java
```diff
@@ -79,6 +79,16 @@ public void testDifferentTableKey() {
     }
 
     @Test
+    public void testNoTableKey() {
```
bsremac: I find test method name confusing compared with the first statement in the test :)
However, it's difficult to suggest a name which wouldn't break the convention with testDifferentTableKey() test. Initially, I was thinking about the name along the lines of _testFilterSettingsWithoutKey_
Or maybe leaving the method name as it is and calling explicitly
```java
filterSettings.setKey(null);
```
for better clarity?

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/dataset/DataSetAwareSelect.java
```diff
@@ -68,7 +68,7 @@ public void setTableKey(String tableKey) {
 
     public void onDataSetReady(@Observes DataSetReadyEvent event) {
         final FilterSettings filterSettings = event.getFilterSettings();
-        if (filterSettings.getKey().equals(this.tableKey) == false) {
+        if (this.tableKey.equals(filterSettings.getKey()) == false) {
```
bsremac: This would be a perfect interview question :)
"How does this fix the bug"?

jbpm-wb-common/jbpm-wb-common-client/src/test/java/org/jbpm/workbench/common/client/dataset/DataSetAwareSelectTest.java
cristianonicolai: @bsremac lack of creativity from my side :P I've changed the test method name.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/dataset/DataSetAwareSelect.java
cristianonicolai: classic NPE question :)

bsremac:
+1

PR #799, opened 2017-07-14
JBPM-6182 - Add validation for Work Item handlers declarations in Dep…

mswiderski:
…loyment Descriptor Editor

### DIFF
--- a/jbpm-wb-integration/jbpm-wb-integration-backend/src/main/java/org/jbpm/workbench/wi/backend/server/dd/DDEditorServiceImpl.java
+++ b/jbpm-wb-integration/jbpm-wb-integration-backend/src/main/java/org/jbpm/workbench/wi/backend/server/dd/DDEditorServiceImpl.java
@@ -18,6 +18,7 @@ package org.jbpm.workbench.wi.backend.server.dd;
 
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
@@ -35,6 +36,7 @@ import org.jbpm.workbench.wi.dd.model.DeploymentDescriptorModel;
 import org.jbpm.workbench.wi.dd.model.ItemObjectModel;
 import org.jbpm.workbench.wi.dd.model.Parameter;
 import org.jbpm.workbench.wi.dd.service.DDEditorService;
+import org.jbpm.workbench.wi.dd.type.DDResourceTypeDefinition;
 import org.kie.internal.runtime.conf.AuditMode;
 import org.kie.internal.runtime.conf.DeploymentDescriptor;
 import org.kie.internal.runtime.conf.NamedObjectModel;
@@ -65,6 +67,9 @@ public class DDEditorServiceImpl
 
     @Inject
     private CommentedOptionFactory commentedOptionFactory;
+
+    @Inject
+    private DDResourceTypeDefinition resourceTypeDefinition;
 
     @Override
     public DeploymentDescriptorModel load(Path path) {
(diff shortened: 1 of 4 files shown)


manstis:
Added @cristianonicolai as a reviewer since this PR is in ```jbpm-wb```

hasys:

jbpm-wb-integration/jbpm-wb-integration-backend/src/main/java/org/jbpm/workbench/wi/backend/server/dd/DDEditorServiceImpl.java
```diff
@@ -157,6 +173,65 @@ public String toSource(Path path,
...
+    protected List<ValidationMessage> validateObjectModels(Path path, List<? extends ObjectModel> objectModels) {
+
+        final List<ValidationMessage> validationMessages = new ArrayList<ValidationMessage>();
+
+        objectModels.forEach(model -> {
+
+            String identifier = model.getIdentifier();
+
+            if (identifier == null || identifier.isEmpty()) {
+                validationMessages.add(newMessage(path, "Identifier cannot be empty for " + model.getIdentifier(), Level.ERROR));
```
hasys: Any chances to make localization for this and other messages?

mswiderski:
@hasys when it comes to the first one - mvel check is basic one only to warn (and not error out) potential issue. This might be on project class path but deployment descriptor does not have access to it's class loader - remember it is not built when the validation happen.

second thing I don't think it's a problem, name can be anything, you can still name your work item like "SOmeName `12~!#!@$#@$#@%$#^$%&^%&^&(&)()()+_)|}{|';';/.,:{|}" so editor can't validate that.

hasys:
Hi @mswiderski,

any changes to make warning about `Identifier` even more useful? At least extend message with explanation? Like
> Use String in format `new some.package.ClassName()`
and, probably some regexp to check **new** package.**ClassName()** format?

@tsurdilo what do you think about WID `Name` field is it possible to use any symbol there?

mswiderski:
@hasys we cannot limit mvel expression to new instances only. You can use pretty much any expression e.g. System.getProperty("my property") and that is valid as well. Similar you can create instance and call a method on it and so on. So there is no way to actually provide anything better - as far as I can see

hasys:
@mswiderski, OK I see, thank you for explanation. Maybe, in this case, we can add some link to detailed documentation which values Identifier can take?

mswiderski:
@hasys that would kind of mean mvel docs... not sure we need to be so "restrictive" here, let's trust a bit our users :)

hasys:
@mswiderski it is not about trust it is about convenience of our's users :)

Ok, you are right. In this case this PR is OK, just let's hear @tsurdilo about `Names` task names and some comments about Localization (we can create separate Jira for it).

P.S. @yurloc suggested another way how to help to user to understand what's going on and how to fix it. See [JBPM-6215](https://issues.jboss.org/browse/JBPM-6215)

mswiderski:
@hasys I see only once difference in the requested changes - enforce validation on save. Besides that this work should pretty much cover most immediate needs.

mswiderski:
enhanced implementation using BuildValidationHelper on deployment descriptor service

tsurdilo:
@hasys @mswiderski about the question on name property in wid - in bpmn2 the name of the node (which comes from the name param in wid) is wrapped in cdata block (nodename element) so it should allow anything there so we should be good there as well as maciej said. Thanks.

hasys:
@tsurdilo If I correct in Deployment Description -> Work Item Handlers -> Name field is not an Element Displayed name, but Element ID.
I mean in this WID Name from Work Item Handlers is `"name: "MyTask",` not `"displayName": "My Task"`. And in this case it will have some limitations. At leas I tried Name "SOmeName `12~!#!@$#@$#@%$#^$%&^%&^&(&)()()+_)|}{|';';/.,:{|}" and task didn't appear on canvas when I dragged it form library.
image

tsurdilo:
@hasys when designer reads all found wids it keeps all names as identifier during round-tripping bpmn2 to know which tasks are custom workitems. At the same time it dynamically creates each of the workitems SVG (to be displayed on the canvas) with a file name with "invalid" chars stripped out. I don't have my work laptop with me to test locally but would assume that when you drag this workitem onto the canvas, designer cannot find the associated svg file that it created (or not..would need to look at logs) and that is why it cannot draw it. I would suggest raising a jbpm-designer Jira for this if possible. Thanks.

hasys:

hasys:

PR #801, opened 2017-07-17
BPMSPL-633: Fix error count popover selectors

rbarriuso:
Fix error count popover selectors to avoid conflict with advanced search
filters


bsremac:

bsremac:
@rbarriuso There's a bug with the error count popover I noticed last week. I wanted to report it separately, but maybe this is a good time and place to mention it and possibly fix it as part of the same PR.

The issue is the the popover persists if not hovered over it with the mouse pointer. It persists even when perspective is changed. Have a look at the following two screenshots:
screenshot from 2017-07-17 13-17-35

screenshot from 2017-07-17 13-18-55

The funny thing around the second screenshot is that the link works :)
I would expect that the popover disappears if the user moves the mouse pointer anywhere else on the page, outside popover area.

rbarriuso:
@bsremac I see. The problem is that Patternfly's popover is hidden when the pointer leaves the popover source (in this case the error count cell), and it's difficult to workaround that behavior. That's the reason now it's opened on hover and kept open until clicked or hovered again.
I think it's better to open a separate issue about this.

bsremac:
@rbarriuso All right. Let's track this problem separately then.
+1 for the current changes in this PR.

cristianonicolai:
ok to test

PR #804, opened 2017-07-19
Fix Case Management Showcase app provisioning error

cristianonicolai:
Add missing provider-name parameter for ALA API


bsremac:
+1

PR #806, opened 2017-07-21
JBPM-5998 : Tasks: Completed tasks disappear from All tab when proce…

nmirasch:
…ss instance completes
Added fields to taskSelectionEvent to be able to show task details and task context using the audit data if the task is not persisted.
Modified human task services to not throw exception when the task is not persisted. 
Made task process context implementation consistent with list 'View process' action, allowing the deletion of  the jbpm-wb-process-runtime-api dependency

### DIFF
--- a/jbpm-wb-human-tasks/jbpm-wb-human-tasks-api/src/main/java/org/jbpm/workbench/ht/model/events/TaskSelectionEvent.java
+++ b/jbpm-wb-human-tasks/jbpm-wb-human-tasks-api/src/main/java/org/jbpm/workbench/ht/model/events/TaskSelectionEvent.java
@@ -17,6 +17,7 @@
 package org.jbpm.workbench.ht.model.events;
 
 import java.io.Serializable;
+import java.util.Date;
 
 import org.jboss.errai.common.client.api.annotations.Portable;
 
@@ -27,8 +28,16 @@ public class TaskSelectionEvent implements Serializable {
     private String containerId;
     private Long taskId;
     private String taskName;
-    private boolean forAdmin;
-    private boolean forLog;
+    private Boolean forAdmin;
+    private Boolean forLog;
+
+    private String description;
+    private Date expirationTime;
+    private String actualOwner;
+    private String status;
+    private Integer priority;
+    private Long processInstanceId;
+    private String processId;
 
     public TaskSelectionEvent() {
     }
@@ -51,14 +60,70 @@ public class TaskSelectionEvent implements Serializable {
                               String containerId,
                               Long taskId,
                               String taskName,
-                              boolean forAdmin,
-                              boolean forLog) {
+                              Boolean forAdmin,
+                              Boolean forLog) {
         this.serverTemplateId = serverTemplateId;
         this.containerId = containerId;
         this.taskId = taskId;
         this.taskName = taskName;
         this.forAdmin = forAdmin;
         this.forLog = forLog;
+    }
+
+    public TaskSelectionEvent(String serverTemplateId,
+                              String containerId,
+                              Long taskId,
+                              String taskName,
+                              Boolean forAdmin,
+                              Boolean forLog,
+                              String description,
+                              Date expirationTime,
+                              String status,
+                              String actualOwner,
+                              Integer priority,
+                              Long processInstanceId,
+                              String processId) {
+        this.serverTemplateId = serverTemplateId;
+        this.containerId = containerId;
+        this.taskId = taskId;
+        this.taskName = taskName;
+        this.forAdmin = forAdmin;
+        this.forLog = forLog;
+        this.description = description;
+        this.expirationTime = expirationTime;
+        this.status = status;
+        this.actualOwner = actualOwner;
+        this.priority = priority;
+        this.processInstanceId = processInstanceId;
+        this.processId = processId;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public Date getExpirationTime() {
+        return expirationTime;
+    }
+
+    public String getActualOwner() {
+        return actualOwner;
+    }
+
+    public String getStatus() {
+        return status;
+    }
+
+    public Integer getPriority() {
+        return priority;
+    }
+
+    public Long getProcessInstanceId() {
+        return processInstanceId;
+    }
+
+    public String getProcessId() {
+        return processId;
     }
 
     public String getTaskName() {
@@ -69,11 +134,11 @@ public class TaskSelectionEvent implements Serializable {
         return taskId;
     }
 
-    public boolean isForAdmin() {
+    public Boolean isForAdmin() {
         return forAdmin;
     }
 
-    public boolean isForLog() {
+    public Boolean isForLog() {
         return forLog;
     }
 
--- a/jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/RemoteTaskServiceImpl.java
+++ b/jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/RemoteTaskServiceImpl.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
+
 import org.jboss.errai.bus.server.annotations.Service;
 import org.jbpm.workbench.ht.model.CommentSummary;
 import org.jbpm.workbench.ht.model.TaskAssignmentSummary;
@@ -31,7 +32,7 @@ import org.jbpm.workbench.ht.model.TaskSummary;
 import org.jbpm.workbench.ht.service.TaskService;
 import org.jbpm.workbench.ks.integration.AbstractKieServerService;
 import org.kie.internal.identity.IdentityProvider;
-import org.kie.server.api.exception.KieServicesException;
+import org.kie.server.api.exception.KieServicesHttpException;
 import org.kie.server.api.model.instance.TaskComment;
 import org.kie.server.api.model.instance.TaskEventInstance;
 import org.kie.server.api.model.instance.TaskInstance;
@@ -43,6 +44,8 @@ import static java.util.stream.Collectors.toList;
 @Service
 @ApplicationScoped
 public class RemoteTaskServiceImpl extends AbstractKieServerService implements TaskService {
+
+    public static int NOT_FOUND_ERROR_CODE = 404;
 
     @Inject
     private IdentityProvider identityProvider;
@@ -61,10 +64,14 @@ public class RemoteTaskServiceImpl extends AbstractKieServerService implements T
             TaskInstance task = client.getTaskInstance(containerId,
                                                        taskId);
             return new TaskSummaryMapper().apply(task);
-        } catch (KieServicesException e) {
-            // task not found
-            return null;
+        } catch (KieServicesHttpException kieException) {
+            if (kieException.getHttpCode() == NOT_FOUND_ERROR_CODE) {
+                return null;
+            } else {
+                throw kieException;
+            }
         }
+
     }
 
     @Override
@@ -255,10 +262,18 @@ public class RemoteTaskServiceImpl extends AbstractKieServerService implements T
         UserTaskServicesClient client = getClient(serverTemplateId,
                                                   UserTaskServicesClient.class);
 
-        List<TaskComment> comments = client.getTaskCommentsByTaskId(containerId,
-                                                                    taskId);
+        try {
+            List<TaskComment> comments = client.getTaskCommentsByTaskId(containerId,
+                                                                        taskId);
+            return comments.stream().map(c -> build(c)).collect(toList());
+        } catch (KieServicesHttpException kieException) {
+            if (kieException.getHttpCode() == NOT_FOUND_ERROR_CODE) {
+                return emptyList();
+            } else {
+                throw kieException;
+            }
+        }
 
-        return comments.stream().map(c -> build(c)).collect(toList());
     }
 
     @Override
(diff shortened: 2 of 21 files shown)


nmirasch:
depends on https://github.com/kiegroup/jbpm/pull/920 to check if the user that access to task administration is defined at server as a business administrator

@cristianonicolai @bsremac could you take a look? thanks!

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskadmin/TaskAdminPresenterTest.java
```diff
@@ -0,0 +1,125 @@
...
+    private Caller<TaskService> remoteTaskServiceCaller;
+
+    @Mock
+    private EventSourceMock<TaskRefreshedEvent> taskRefreshedEvent;
+
+    private TaskAdminPresenter presenter;
+
+    @Before
+    public void initMocks() {
+        when(viewMock.getActualOwnerPanel()).thenReturn(GWT.create(FormControlStatic.class));
```
cristianonicolai: Ideally the presenter should not know about these GWT components, instead, I think the `TaskAdminView` interface should have methods to receive the necessary text Strings.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskcomments/TaskCommentsPresenterTest.java
```diff
@@ -133,4 +135,56 @@ public void removeCommentAdded() {
...
+                                                     "commentText",
+                                                     "ByTest",
+                                                     new Date());
+        when(commentsServiceMock.getTaskComments(eq(serverTemplateId),
+                                                 eq(containerId),
+                                                 eq(taskId))).thenReturn(Arrays.asList(comment1));
+
+        presenter.onTaskSelectionEvent(event);
+
+        verifyNoMoreInteractions(commentsServiceMock);
```
cristianonicolai: if all you want to check is that the `getTaskComments` isn't called, I would recommend using `verifyZeroInteractions` instead and perhaps even not do any mock preparation with `when`.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-api/src/main/java/org/jbpm/workbench/ht/model/events/TaskSelectionEvent.java
```diff
@@ -61,6 +70,62 @@ public TaskSelectionEvent(String serverTemplateId,
         this.forLog = forLog;
     }
 
+    public TaskSelectionEvent(String serverTemplateId,
```
cristianonicolai: Could you add the related methods `equals`, `hascode` and `toString`? Another suggestion would be to move `forAdmin` and `forLog` to Boolean instead.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskdetails/TaskDetailsPresenter.java
```diff
@@ -53,12 +53,14 @@
 
     private String currentContainerId;
 
+    private TaskSummary currentTaskSummary;
```
cristianonicolai: I'm not sure you need to hold the current task, in terms of testing you could check that the view methods get the expected values.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskdetails/TaskDetailsPresenter.java
```diff
@@ -131,6 +142,16 @@ public void onTaskSelectionEvent(@Observes final TaskSelectionEvent event) {
         this.currentTaskId = event.getTaskId();
         this.currentServerTemplateId = event.getServerTemplateId();
         this.currentContainerId = event.getContainerId();
+        this.currentTaskSummary =
```
cristianonicolai: `refreshTask` calls the backend, so I'm not sure if just doing `setTaskDetails` would be enough then.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskdetailsmulti/TaskDetailsMultiPresenter.java
```diff
@@ -146,17 +146,18 @@ public void onTaskSelectionEvent(@Observes final TaskSelectionEvent event) {
...
-            }
-        });
-        taskFormDisplayProvider.setup(new HumanTaskDisplayerConfig(new TaskKey(serverTemplateId,
-                                                                               containerId,
-                                                                               taskId)),
-                                      taskFormPresenter.getTaskFormView().getDisplayerView());
-
+        if (!event.isForLog()) {
+            taskFormPresenter.getTaskFormView().getDisplayerView().setOnCloseCommand(new Command() {
+                @Override
```
cristianonicolai: Shorter form using lambda? :)

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskdetails/TaskDetailsPresenter.java
nmirasch: the idea is the refresh have to do as much as possible. If the task is not persisted, the details would be the one that the task refresh event holds. But if the task is persisted the method getTask return the last task data. To allow the refresh operation in both cases, I thought to have the last load task details in the currenTaskSummary to be used when there's no task details  at DB. This task selection method will try to access to the last task data and call the setTaskDetails with the best data maintaining the previous behavior. Here we have to evaluate is we can rely always on the event data and avoid to call the getTask Method in any case. What do you think @cristianonicolai about this modification? does it make sense?

nmirasch:
@cristianonicolai changes done! Could you take another look? thanks!

jbpm-wb-human-tasks/jbpm-wb-human-tasks-api/src/main/java/org/jbpm/workbench/ht/model/events/TaskSelectionEvent.java
cristianonicolai: @nmirasch can you take a look at adding the methods mentioned above?

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskdetails/TaskDetailsPresenter.java
```diff
@@ -92,31 +84,23 @@ public void callback(Void nothing) {
...
-            }
-        }).getTask(currentServerTemplateId,
-                   currentContainerId,
-                   currentTaskId);
+    protected void setTaskDetails(String status,
+                                  String description,
+                                  String actualOwner,
+                                  Date expirationTime,
+                                  String priority) {
+        if (status.equals(TASK_STATUS_COMPLETED)) {
```
cristianonicolai: Usually is better to use the constant as the first element to compare, that way you can avoid potential NPE.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskdetails/TaskDetailsPresenter.java
cristianonicolai: @nmirasch looks good.

nmirasch:
@cristianonicolai changed

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/RemoteTaskServiceImpl.java
```diff
@@ -255,10 +255,14 @@ public void deleteTaskComment(String serverTemplateId,
...
-        List<TaskComment> comments = client.getTaskCommentsByTaskId(containerId,
-                                                                    taskId);
-
-        return comments.stream().map(c -> build(c)).collect(toList());
+        try {
+            List<TaskComment> comments = client.getTaskCommentsByTaskId(containerId,
+                                                                        taskId);
+            return comments.stream().map(c -> build(c)).collect(toList());
+        } catch (KieServicesException e) {
+            // task not found
```
bsremac: Having a look at _getTaskCommentsByTaskId()_ implementation in _org.kie.server.client.impl.UserTaskServicesClientImpl_, the exception is thrown when there's an actual failure response upon executing JMS command. Are we sure that it's safe to simply swallow this exception with a comment and continue?

nmirasch: I tried replicate the same behavior as for the getTask, because it's the same call only retrieving the comments information. The idea is when the operation of get comments fails just show an empty llist without showing any exception here

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskadmin/TaskAdminPresenterTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.mockito.Mock;
+import org.uberfire.mocks.CallerMock;
+import org.uberfire.mocks.EventSourceMock;
+
+import static org.mockito.Mockito.*;
+
+@RunWith(GwtMockitoTestRunner.class)
+public class TaskAdminPresenterTest {
+
+    private static final String CURRENT_USER = "Jan";
```
bsremac: Jan will be honored :)

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskdetails/TaskDetailsPresenterTest.java
```diff
@@ -15,31 +15,160 @@
...
+        verify(viewMock).setTaskPriority(priority);
+    }
 
-        verify(view).setTaskDescriptionEnabled(false);
-        verify(view).setDueDateEnabled(false);
-        verify(view).setDueDateTimeEnabled(false);
-        verify(view).setTaskPriorityEnabled(false);
-        verify(view).setUpdateTaskVisible(false);
+    private void verifyReadOnlyMode(int i,
+                                    boolean readOnly) {
```
bsremac: It's quite confusing to have the method which verifies read-only mode and then have an argument which determines whether it's read-only or not. 
Furthermore, the argument value opposes the argument name, having _false_ actually means read-only.

I would suggest removing the boolean argument and hard-code _false_ in all method calls inside. The method is used only with boolean being set to _false_ in tests anyway, so no ram is done and it's much clearer.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/RemoteTaskServiceImpl.java
bsremac: Thanks for the clarification. Since this approach was already approved some time in the past, I guess it's they way to go. I was just finding it a little bit suspicious to treat an error as an empty list, from user experience perspective.
If the list can't be retrieved, it would provide incorrect information to the user that the list is empty where in reality it can contains data, but data can't be displayed.

nmirasch:
@bsremac test changed

bsremac:
+1

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/RemoteTaskServiceImpl.java
cristianonicolai: @nmirasch I think @bsremac got a point here, we could be hiding other errors with this approach. Is there any information in the _KieServicesException_ that we could make sure it is related to a _TaskNotFoundException_ only?

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/RemoteTaskServiceImpl.java
```diff
@@ -61,10 +64,12 @@ public TaskSummary getTask(String serverTemplateId,
             TaskInstance task = client.getTaskInstance(containerId,
                                                        taskId);
             return new TaskSummaryMapper().apply(task);
-        } catch (KieServicesException e) {
-            // task not found
-            return null;
+        } catch (KieServicesHttpException kieException) {
+            if (kieException.getHttpCode() != NOT_FOUND_ERROR_CODE) {
```
cristianonicolai: just to make a bit easier to read this, can you update to:
```
if (kieException.getHttpCode() == NOT_FOUND_ERROR_CODE) {
  return null;
} else {
throw kieException;
}
```

nmirasch:
@cristianonicolai  done

tsurdilo:
Jenkins retest please

PR #807, opened 2017-07-24
Enable checkstyle validation

cristianonicolai:


pom.xml
```diff
@@ -51,6 +51,8 @@
 ]]>
     </checkstyle.header.template>
     <findbugs.failOnViolation>true</findbugs.failOnViolation>
+    <checkstyle.failOnViolation>true</checkstyle.failOnViolation>
+    <checkstyle.logViolationsToConsole>true</checkstyle.logViolationsToConsole>
```
bsremac: I couldn't find _checkstyle.logViolationsToConsole_ in here:
https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html

On this page, it is said the following property sets the value of _logViolationsToConsole_:
_checkstyle.console_

Am I referencing the wrong documentation?

jhrcek: @bsremac It seems you found an issue with checkstyle plugin configuration in kiegroup projects (declaring `checkstyle.logViolationsToConsole` should not have any effect). I'd recommend removing that property from configuration altogether, as the default value of the "correct" `checkstyle.console` property [is true](https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html#logViolationsToConsole) anyway

cristianonicolai: @bsremac @jhrcek thanks, removed _checkstyle.logViolationsToConsole_ property.

bsremac:
@cristianonicolai The other potential problem, from reading the documentation, can arise from having _failsOnError_ set to _true_ (line 193). This setting should fail the build immediately, hence not logging anything regardless of _logViolationsToConsole_  being set to _true_ implicitly.

cristianonicolai:
@bsremac thanks you're correct about _failsOnError_ so I removed. On the other hand, I had to set _logViolationsToConsole_ to ensure that messages get printed. Perhaps somewhere up in the pom.xml hierarchy it is set to false. See https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html#logViolationsToConsole

pom.xml
psiroky: See https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L178 and https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L573

bsremac:
@cristianonicolai Based on the references Petr pointed out, it seems that when it comes to _checkstyle.logViolationsToConsole_ property name it is a valid name after all :) and yes - it needs to be explicitly set to _true_  here.
+1

PR #820, opened 2017-08-01
[7.2.x] JBPM-6233 - Provide web socket based alternative for KIE Server to co…

mswiderski:
…ntroller communication

depends on https://github.com/kiegroup/droolsjbpm-integration/pull/1085


jhrcek:
Hello @mswiderski , Just being curious: Is the web socket implementation going to be used in console by default? I heard that this would mean no extra authentication would be required for WB <-> KS communication. Is that right?

mswiderski:
@jhrcek when web socket is used as the protocol then only kie server needs to have credentials to connect to controller but the controller does not have to know credentials to kie server.

When it comes to defaults, once it proves to be stable (Which it seems after my tests) it can be used as default.

PR #827, opened 2017-08-03
[master] JBPM-5406 Implement results pagination in Case Management App

annabaker:
- Added pagination parameters to getCaseInstances + getCaseMilestones
- Pagination changes reflected in presenters and view implementations
- il8n constant for Load More made less specific so it could be shared amongst other results lists
- Mock Case Management changes for testing
- Added live search to Case Definitions drop down
-  Unit Tests
- Small renaming/reorg of methods
- Small formatting changes (I applied these using IntelliJ with the codestyle xml)


annabaker:
jenkins retest

PR #849, opened 2017-08-18
[7.3.x] JBPM-6137 - Remove deprecated navbar search box related classes

cristianonicolai:
Fix dependencies for using the new admin page,


cristianonicolai:
Jenkis execute full downstream build

jbpm-wb-showcase/src/main/java/org/jbpm/workbench/client/ShowcaseEntryPoint.java
```diff
@@ -57,11 +58,14 @@
 
     protected WorkbenchMegaMenuPresenter menuBar;
 
+    protected DefaultAdminPageHelper adminPageHelper;
```
bsremac: Have just one question regarding introduction of this object. Why it's needed? I don't see it's used anywhere.

cristianonicolai: @bsremac it's used at `setupAdminPage` method

PR #850, opened 2017-08-18
JBPM-6137 - Remove deprecated navbar search box related classes

cristianonicolai:
Fix dependencies for using the new admin page,


cristianonicolai:

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/definition/list/ProcessDefinitionListPresenter.java
```diff
@@ -159,7 +158,7 @@ public boolean error(Message message,
                                                return onRuntimeDataServiceError();
                                            }
                                        }).getProcessesByFilter(getSelectedServerTemplate(),
-                                                               textSearchStr,
+                                                               null,
```
bsremac: Just one question here.
Instead of introducing ```null``` argument, why not switching to ```getProcesses``` method?

cristianonicolai: @bsremac good catch, fixed.

PR #851, opened 2017-08-22
JBPM-5406 Implement results pagination in Case Management App

annabaker:
- Added pagination parameters to getCaseInstances + getCaseMilestones
- Pagination changes reflected in presenters and view implementations
- il8n constant for Load More made less specific so it could be shared amongst other results lists
- Mock Case Management changes for testing
- Added live search to Case Definitions drop down
-  Unit Tests
- Small renaming/reorg of methods
- Small formatting changes (I applied these using IntelliJ with the codestyle xml)


jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/milestones/CaseMilestoneListPresenter.java
```diff
@@ -48,19 +59,56 @@ protected void clearCaseInstance() {
...
     }
 
-    protected void refreshData(String caseId) {
+    protected void loadMoreCaseMilestones() {
+        setCurrentPage(getCurrentPage() + 1);
+        milestonesServiceCall(caseId,
+                              getCurrentPage());
+    }
+
+    private void loadButtonToggle() {
```
cristianonicolai: Instead of loading the next page of items, the load button should be visible if loading the page returned the same number of items as the page size. This way we can avoid this extra call and keep consistency with other search UI behavior.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/comments/CaseCommentsPresenter.java
```diff
@@ -127,7 +122,11 @@ protected void addCaseComment(String caseCommentText) {
     protected void updateCaseComment(final CaseCommentSummary caseCommentSummary,
                                      String caseCommentNewText) {
         caseService.call(
-                (Void) -> refreshComments()
+            (Void) -> {
+                visibleCaseComments.remove(caseCommentSummary);
```
cristianonicolai: Not sure I understood why it is needed to remove and add the caseCommentSummary object into the list. Perhaps only the text should be updated in the object and not even call the refreshComments method.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/comments/CaseCommentsPresenter.java
```diff
@@ -138,7 +137,10 @@ protected void updateCaseComment(final CaseCommentSummary caseCommentSummary,
 
     protected void deleteCaseComment(final CaseCommentSummary caseCommentSummary) {
         caseService.call(
-                (Void) -> refreshComments()
+            (Void) -> {
+                visibleCaseComments.remove(caseCommentSummary);
+                refreshComments();
```
cristianonicolai: I think you can avoid calling `refreshComments` as you already removed the item from the list. You only need to update the view now.

jsoltes:
This pr is from august, @annabaker are you going to update this?

PR #854, opened 2017-08-22
[7.3.x] JBPM-5103 - Rename Request List to Jobs

cristianonicolai:


bsremac:
So "Jobs" is the way we want to go.
I was giving more chances to "Requests" really, because ```kie.server.client``` API uses "Requests", so I was sure when the time comes, we would align the name with the API.

Jobs is much more user friendly though, no doubt.

cristianonicolai:
@bsremac I had the same "conflict" but it seems that at the UI and documentation level Jobs is the better term. At the code level, it still has references to requests.

bsremac:
+1

PR #875, opened 2017-09-14
JBPM-6416 - Deployment descriptor validation fix

tsurdilo:


tsurdilo:
@mswiderski after looking more into this and trying to fix the issue @hasys found during testing I think best thing we can do is make the validation for mvel expressions a syntax only check. This fix does that. Validation will still report error in case of an invalid mvel expression being entered by the user but class/parameter type resolutions are not going to be checked.

mswiderski:
@tsurdilo fine with me, if that brings in syntax check then that's what was intended from the beginning, @hasys what do you think?

tsurdilo:
@hasys could you let us know please? thanks.

tsurdilo:
@hasys any news on this by chance?

hasys:
Hi @tsurdilo, looks good, just want to make final check and have a look how does it work in WB.

hasys:

PR #888, opened 2017-09-23
[7.3.x] JBPM-6281 - Use likeTo to filter Error Id

cristianonicolai:


jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenter.java
```diff
@@ -286,9 +286,9 @@ public void setupAdvancedSearchView() {
 
         view.addTextFilter(constants.Id(),
                            constants.FilterByErrorId(),
-                           v -> addAdvancedSearchFilter(equalsTo(COLUMN_ERROR_ID,
+                           v -> addAdvancedSearchFilter(likeTo(COLUMN_ERROR_ID,
```
bsremac: Would this also work if "_" operator is used?

cristianonicolai: @bsremac it should respect the wildcards usage as any other field using _likeTo_,  I believe both "_" and "%" should work just fine.

PR #894, opened 2017-09-25
[7.3.x] JBPM-6437 - Allow date filter to specify whether to use calendar max …

cristianonicolai:
…date or not


cristianonicolai:
@bsremac I tend to prefer using some kind of "options" object instead of adding multiple methods. Perhaps if we start adding more configuration options in the future, we would still keep one method and each presenter would set it up as needed. I guess is just a matter of preference anyway.

PR #911, opened 2017-10-03
[7.3.x] JBPM-6374: Listing pages unchecks previosuly checked rows

nmirasch:


nmirasch:
@bsremac, @cristianonicolai could you take a look? thanks!

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractListPresenter.java
```diff
@@ -133,6 +133,7 @@ public void updateDataOnCallback(List<T> instanceSummaries,
                                    instanceSummaries);
 
         updateRefreshTimer();
+        getListView().getListGrid().setVisibleSelectedItems();
```
cristianonicolai: the same should be applied in the _updateDataOnCallback_ method above. My suggestion would be to consolidate both methods, the one receiving a _PageResponse_ should call this one passing the related parameters.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/ExtendedPagedTable.java
```diff
@@ -180,10 +195,12 @@ public boolean clearCurrentSelection(CellPreviewEvent<T> event) {
                             final T domain = event.getValue();
                             if (domain.isSelected()) {
                                 input.setChecked(false);
-                                domain.setSelected(false);
+                                setItemSelection(domain,
+                                                 false);
```
bsremac: I'm not completely sure if ```domain.isSelected()``` and ```isItemSelected(item)``` are interchangeable here.
If they are, then instead of ```setItemSelection(domain, false)``` a direct call to ```selectedItems.remove(item)``` can be made here.

The same comment applies to ```else``` branch in this case.

nmirasch: @bsremac domain.isSelected is the new value that comes in the click on checkbox event. isItemSelected specifies if the item has been previously selected. SetItemSelection tries to encapsulate the interaction with that saved selections.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractListPresenter.java
nmirasch: done!

nmirasch:
@bsremac @cristianonicolai updated with your suggestions. Could you give it another try? thanks!

PR #916, opened 2017-10-10
JBPM-5651 - Add excluded owner table join to task data set

cristianonicolai:


cristianonicolai:
Depends on https://github.com/kiegroup/jbpm/pull/1023

bsremac:
I could add one Selenium test for this, if it makes sense. This test would cover UI and REST-based exchange with Kie Server.
As for the asset, I would create a stripped-down version of the process which is mentioned in the corresponding JIRA. It would contain only 2 user tasks with the latter one having ```ExcludedOwnerId``` property set as data input and assigning it to the tested user Id. In this way, the tested user is expected to see the first task (as a group owner) and is not expected to see the second one after the user claims and completes the first one, although the second task is assigned to the same group as the first one.

cristianonicolai:
@mswiderski @bsremac +1 for the Selenium test. I'll also take a look at adding the referenced process as a test before merging this. As for checking if the join is present, there is probably not much value on testing that.

PR #920, opened 2017-10-17
RHBPMS-5019: Job list: Job request details broken when command throws…

nmirasch:
… exception

Removed fileset that was causing the overflow problem in firefox and replace the text including it in a textarea.

### DIFF
--- a/jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPopup.java
+++ b/jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPopup.java
@@ -22,17 +22,16 @@ import javax.inject.Inject;
 
 import com.google.gwt.cell.client.TextCell;
 import com.google.gwt.core.client.GWT;
-import com.google.gwt.safehtml.shared.SafeHtmlUtils;
 import com.google.gwt.uibinder.client.UiBinder;
 import com.google.gwt.uibinder.client.UiField;
 import com.google.gwt.user.cellview.client.Column;
-import com.google.gwt.user.client.ui.HTML;
 import com.google.gwt.user.client.ui.Widget;
 import com.google.gwt.view.client.ListDataProvider;
 import org.gwtbootstrap3.client.shared.event.ModalShownEvent;
 import org.gwtbootstrap3.client.shared.event.ModalShownHandler;
 import org.gwtbootstrap3.client.ui.FormControlStatic;
 import org.gwtbootstrap3.client.ui.FormGroup;
+import org.gwtbootstrap3.client.ui.TextArea;
 import org.gwtbootstrap3.client.ui.constants.ButtonType;
 import org.gwtbootstrap3.client.ui.gwt.DataGrid;
 import org.jboss.errai.common.client.api.Caller;
@@ -62,7 +61,7 @@ public class JobDetailsPopup extends BaseModal {
     public DataGrid<RequestParameterSummary> executionParametersGrid;
 
     @UiField
-    public HTML errorsOccurredList;
+    public TextArea errorsOccurredTextArea;
 
     @UiField
     public FormGroup errorControlGroup;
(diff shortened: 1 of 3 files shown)


nmirasch:
@cristianonicolai @bsremac  could you take a look? Thanks!

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPopup.java
```diff
@@ -128,6 +127,9 @@ public String getValue(RequestParameterSummary rowObject) {
                                                                                        paramValueColumn));
 
         this.dataProvider.addDataDisplay(executionParametersGrid);
+
+        errorsOccurredList.setVisibleLines(10);
```
bsremac: Would only 10 lines of stack trace be enough to troubleshoot the error?
I guess it's not possible to make the number of lines shown dynamic i.e. to expand to fit the whole stack trace, right?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPopup.java
```diff
@@ -164,9 +166,11 @@ public void setRequest(RequestSummary r,
             errorControlGroup.setVisible(true);
             String html = "";
             for (ErrorSummary error : errors) {
-                html += "<strong>" + error.getMessage() + "</strong><br/>" + error.getStacktrace() + "<br><br>";
+                html += error.getMessage() +
```
bsremac: Maybe it makes sense for easier readability reasons to also rename the variable from ```html`` to, for example, ```textArea``` so it's aligned with the field type used.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPopupTest.java
```diff
@@ -58,6 +57,8 @@ public void errorGenerationTest() {
                                    errors,
                                    null);
 
-        verify(errorsOccurredList).setHTML(SafeHtmlUtils.fromTrustedString("<strong>" + error.getMessage() + "</strong><br/>" + error.getStacktrace() + "<br><br>"));
+        verify(errorsOccurredList).setText("errorMessage\n" +
+                                                   "stackTrace\n" +
+                                                   "\n");
```
bsremac: Just a cosmetic thing here. This String breakdown looks a bit off. Maybe fixing the indention so each part is exactly below the previous one would make it look nicer :)

nmirasch:
@bsremac the problem was with the fileset at firefox. I have removed it and fixes the Execution Parameters table overflow. I think this details popup have to be changed in future to be consistent with the other details views.
The 10 lines limit is to be according to the text area that shows the stacktrace at Execution error view and to avoid the big scrolling at the popup.
I'm going to rename the field and do the cosmetic thing . I'll let you know when its done. Thanks!

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPopup.java
jhrcek: @bsremac I think Neus implemented this correctly. the "visible lines" only refers to the number of rows the textarea will display. It doesn't "trim the stacktrace". The remaining lines are still accessible via textarea's scrollbar as demonstrated [here](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_textarea). I think that it's ok as it is - just show 10 lines and let user resize the area (= default browser behaviour).

nmirasch:
@bsremac changes done, could you take another look?

PR #922, opened 2017-10-19
JBPM-5833: Strange units used for displaying data in charts in Process and Task Dashboard

dgutierr:
@jhrcek @cristianonicolai Please can you review?

It fixes the first issue reported. As for the 2nd issue, see my last comment at JBPM-5833.

### DIFF
--- a/jbpm-wb-dashboard/jbpm-wb-dashboard-client/src/test/java/org/jbpm/dashboard/renderer/client/TaskDashboardTest.java
+++ b/jbpm-wb-dashboard/jbpm-wb-dashboard-client/src/test/java/org/jbpm/dashboard/renderer/client/TaskDashboardTest.java
@@ -20,6 +20,7 @@ import javax.enterprise.event.Event;
 import org.dashbuilder.common.client.error.ClientRuntimeError;
 import org.dashbuilder.dataset.DataSet;
 import org.dashbuilder.dataset.filter.DataSetFilter;
+import org.dashbuilder.displayer.ColumnSettings;
 import org.dashbuilder.displayer.DisplayerSettings;
 import org.dashbuilder.displayer.client.AbstractDisplayer;
 import org.dashbuilder.displayer.client.Displayer;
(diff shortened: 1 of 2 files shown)


jhrcek:
In the meantime @dgutierr can you please explain how adding that extra parameter fixes the formatting issue?

dgutierr:
@jhrcek This is up to the query service implementation supporting the required DataSetLookup group settings, which is more a question for @mswiderski I guess.

jhrcek:
@dgutierr No, I meant extra parameter `COLUMN_TASK_DURATION` to the `column` function in this PR..

dgutierr:
@jhrcek The kie-server is returning "duration" instead of "duration_average". This is why I'm forcing the resulting column as "duration". Without adding so, the format settings are not applied since the "duration_average" column is not found in the resulting dataset. I think this is another issue we could improve in the query service. @mswiderski ^

nmirasch:
@dgutierr the drilldown operation over this field launches an exception "The request could not be understood by the server due to malformed syntax: 2017" when passes the value. Could you take a look to see if it's related?

jhrcek:
@dgutierr I executed regression tests and the fix seems to be working fine.
@nmirasch I wasn't able to reproduce the issue you're describing. Could you please give more details about how to repreoduce that?

nmirasch:
@jhrcek I just created a task and go to 'Process & Task Dashboard"- "Tasks" and click over  Tasks By Creation Date, for example. If shows only the task created with the legend '2017 Tasks:1' . The user can drill down there clicking over the point and that creates a new date filter to be applied at the whole page dashboards. In case of dates dashboards it launches an exception because the date filter is created with that 2017 value instead a valid date filter value. My question to @dgutierr is if this new literal at legend can be the source of the problem or if it is a general problem with the date filters legends and filter creation in dashboards that have to be fixed.

dgutierr:
@nmirasch See my last comment at JBPM-5833. Any side effect of having a single "2017" point of data shall disappear as soon as the root caused is fixed.

dgutierr:
@nmirasch I've added format verification to the test. If ok, can you approve please?

jbpm-wb-dashboard/jbpm-wb-dashboard-client/src/test/java/org/jbpm/dashboard/renderer/client/TaskDashboardTest.java
```diff
@@ -291,6 +292,18 @@ public void testTasksByCreationDate() {
     public void testTasksByRunningTime() {
         Displayer displayer = presenter.getTasksByRunningTime();
         DataSet dataSet = displayer.getDataSetHandler().getLastDataSet();
+        DisplayerSettings displayerSettings = displayer.getDisplayerSettings();
+
+        ColumnSettings taskDurationSettings = displayerSettings.getColumnSettings(COLUMN_TASK_DURATION);
+        assertNotNull(taskDurationSettings);
+        assertEquals(taskDurationSettings.getValueExpression(), "value/60000");
```
nmirasch: @dgutierr Just one more little thing. The first argument of assertEquals have to be the object expected and the second one the object that you are evaluating.

dgutierr: Done.

PR #927, opened 2017-10-23
JBPM-5751: Process form dialog too narrow and cannot be resized

nmirasch:
Removed FormContentResizeListener use because the new forms don't have a predefined size to use at resize and they adapt themselves to their container.
The size of the modals popups that show forms have been changed to the ModalSize.LARGE (900px). 
Also I added a test to cover the QuickNewProcessInstancePopup, even it's not related directly with this fix.


nmirasch:
@pefernan @bsremac  @cristianonicolai  could you take a look please? thanks!

bsremac:
@nmirasch If I'm getting the approach correctly, we are just switching to the larger modal dialog thus increasing the probability that the content in all fields will fit, right? 
There's no (resize) option available to ensure that the content of each field will fit and be visible no mater its length.

nmirasch:
@bsremac The current forms, as far as I know, define fields and they disposition and they try to fit to their container. This popup size change is a temporary solution that, as you say, increase the probability that the displayed form looks nice, increasing the space available for the form. But the last goal have to be display the form in full-screen, it's pending to concrete with UX team. We managed the possibility to add the expand option to the popup that changes its size, but it's something too different to any popup used in all the platform.

bsremac:
@nmirasch I guess, with the full-screen form view on its way and due to the reason of maintaining alignment with other pop-ups, as you mentioned, this seems as the only available fix at the moment.

PR #929, opened 2017-10-25
JBPM-6509 - Support for sub cases both from within case and process

mswiderski:
depends on https://github.com/kiegroup/jbpm/pull/1041

### DIFF
--- a/jbpm-wb-integration/jbpm-wb-integration-backend/src/main/resources/CaseWorkDefinitions.wid
+++ b/jbpm-wb-integration/jbpm-wb-integration-backend/src/main/resources/CaseWorkDefinitions.wid
@@ -64,6 +64,26 @@
     "displayName" : "Milestone",
     "icon" : "defaultmilestoneicon.png",
     "category" : "Milestone"
+  ],
+  
+  [
+    "name" : "StartCaseInstance",
+    "parameters" : [
+      "DeploymentId" : new StringDataType(),
+      "CaseDefinitionId" : new StringDataType(),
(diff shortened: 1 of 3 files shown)


jbpm-wb-integration/jbpm-wb-integration-backend/src/main/resources/CaseWorkDefinitions.wid
```diff
@@ -64,6 +64,26 @@
...
     "category" : "Milestone"
+  ],
+  
+  [
+    "name" : "StartCaseInstance",
+    "parameters" : [
+      "DeploymentId" : new StringDataType(),
+      "CaseDefinitionId" : new StringDataType(),
+      "Data_" : new StringDataType(),
+      "UserRole_" : new StringDataType(),
```
tsurdilo: Whats the reason for the undescore?

mswiderski: this is a prefix/place holder as users should filed in their own data or roles to be mapped - see description of the handler https://github.com/kiegroup/jbpm/pull/1041/files#diff-7e89a760ce30fb700e734f14f68e15f1R50

so in reality it will become Data_name, Data_year where year and name are the names in the case to be set. Same for roles and restrictions

tsurdilo: ok thanks!

PR #935, opened 2017-10-31
[JBPM-6571] Write comminity tests to cover JBPM-6469

bsremac:
@cristianonicolai @nmirasch @triceo  Hi folks, these are the tests which I came up with in attempt to cover JBPM-6469 with a reproducer.
However, while writing the tests I realized that the issue is not on jBPM Workbench side, but rather on Kie Server side (```TaskSummaryList``` field doesn't get initialized in ```org.kie.server.api.model.instance.ProcessInstance```), which in turn returns ```null``` when queried in our backend and as a result an empty list is displayed in UI under ```Active user tasks``` in Process Instance Details.
Anyway, I added two tests (one for backend part and one for the client part), which would have caught this type of issue, if the issue had been on jBPM Workbench side.

I guess, the only way this issue would have been caught by tests in ```jbpm-wb```, if there had been an integration test which deploys Kie Server with a proper process and then uses real data instead of mocks.
I think, these types of issues are not worth taking into consideration with potential integration tests in jBPM Workbench, but rather left to be tested on Kie Server side.

@cristianonicolai @nmirasch I could open a new issue for Kie Server and then add ```caused by``` relationship with the existing issue, so that it's clear where the issue originates from. If you think this just introduces a bureaucracy overhead, then we can leave it as it is, since Neus has already been working on the fix.

### DIFF
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/test/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImplTest.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/test/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImplTest.java
@@ -18,16 +18,22 @@ package org.jbpm.workbench.pr.backend.server;
 
 import java.lang.reflect.Method;
 import java.util.Collection;
+import java.util.Date;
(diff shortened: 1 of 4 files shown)


cristianonicolai:
@bsremac nice work and I totally agree that an integration test should be the best way to catch this error. Anyway I think this is a good exercise that will help identify areas of improvement.
+1 to keep the existing issue for now.

PR #939, opened 2017-11-06
RHBPMS-4390: Exception stack trace overflows from Job Requests Detail…

nmirasch:
…s dialog

Replaced job details popup with the side panel with details.


nmirasch:
@bsremac, @cristianonicolai Need you review on this,. This solutions allow the user enlarge the paned to see the full value, as in other sections, but not sure if make sense right now, I have noted after my development, that the assigned jira has passed to 'propose to the release'
 Could you take a look and give me feedback? Thanks!!

cristianonicolai:
Jenkins please retest this

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobBasicDetailsViewImpl.java
```diff
@@ -0,0 +1,145 @@
...
+
+@Dependent
+@Templated(value = "JobBasicDetailsViewImpl.html")
+public class JobBasicDetailsViewImpl implements TakesValue<RequestSummary>,
+                                                IsElement {
+
+    @Inject
+    @DataField("businesskey_div")
+    Div businesskeyDiv;
+    @Inject
```
cristianonicolai: Please add new line between all these fields.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/requestlist/RequestListPresenter.java
```diff
@@ -97,12 +102,16 @@ public RequestListPresenter() {
...
-                                Event<RequestChangedEvent> requestChangedEvent
+                                Event<RequestChangedEvent> requestChangedEvent,
+                                Event<JobSelectedEvent> jobSelectedEvent,
+                                PlaceManager placeManager
     ) {
         this.view = view;
         this.executorServices = executorServices;
         this.dataSetQueryHelper = dataSetQueryHelper;
         this.requestChangedEvent = requestChangedEvent;
+        this.jobSelectedEvent= jobSelectedEvent;
```
cristianonicolai: Perhaps run a code reformat in this class, there are a few places in this class that need fixing.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/ExecutionParameterListViewImpl.java
```diff
@@ -0,0 +1,100 @@
...
+import com.google.gwt.user.client.ui.Widget;
+import com.google.gwt.view.client.ListDataProvider;
+import org.jbpm.workbench.es.client.i18n.Constants;
+import org.jbpm.workbench.es.model.RequestParameterSummary;
+
+import org.uberfire.ext.widgets.common.client.tables.SimpleTable;
+import org.uberfire.ext.widgets.table.client.ColumnMeta;
+
+@Dependent
+public class ExecutionParameterListViewImpl implements IsWidget {
```
cristianonicolai: Perhaps name it _JobParameterListViewImpl_?

nmirasch:
@cristianonicolai changes done

bsremac:
It seems that there is some issue unrelated to the changes introduced in this PR which prevents any data in Job Details panel to be displayed, only when the job execution has errors.
When I click on the job, this is the error which gets thrown:
jobwitherrros

When the error is acknowledged, Job Details panel contains no data.
jobdetails

nmirasch:
@bsremac thanks, I can't reproduce that effect only with the error acknowledgement that you commented. But running this views with kie-wd-distributions  showcase, I see a lot of messages related with "[ERROR] Security framework of XStream not initialized, XStream is probably vulnerable.". 
Could you pass me the full message about the forbiddenClassException? which class is it exactly?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobParameterListViewImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import org.uberfire.ext.widgets.common.client.tables.SimpleTable;
+import org.uberfire.ext.widgets.table.client.ColumnMeta;
+
+@Dependent
+public class JobParameterListViewImpl implements IsWidget {
+
+    public static final String COL_PARAM_ID = "ParamId";
+    public static final String COL_PARAM_VALUE = "ParamValue";
+    final Constants constants = Constants.INSTANCE;
+    protected SimpleTable<RequestParameterSummary> parameterSummarySimpleTable;
```
cristianonicolai: @nmirasch would it be reasonable to consider updating this implementation to use Errai instead? Something towards what we did in here: https://github.com/kiegroup/jbpm-wb/blob/a434fbd7440c2d81c9eb47dfb0a5037cd8f5ba92/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/newcase/NewCaseInstanceViewImpl.java#L107

cristianonicolai:
@nmirasch these warnings should be gone by now, after a series of PR's like this one: https://github.com/kiegroup/drools-wb/pull/661 we probably need to do another look to find where these are coming from.
@bsremac if this is coming from Kie Server, it could be that some class need explicit permission, see https://github.com/kiegroup/droolsjbpm-integration/blob/c9825c48ecc48599c86e96fa291e21dfd44e19d6/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/KieServerTypePermission.java

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPresenter.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Inject
+    private Event<ChangeTitleWidgetEvent> changeTitleWidgetEvent;
+
+    private PlaceRequest place;
+
+    private String serverTemplateId;
+    private String deploymentId;
+    private Long jobId;
+    private String processName;
```
cristianonicolai: @nmirasch I'm a bit reluctant to include the process related data in the details. Although that is visible on the table, this data is not part of the job domain information. If a job is part of a process, the instance id is shown the parameter table already. @bsremac wdyt?

bsremac: @cristianonicolai @nmirasch Yes, I agree here. I would limit the information exposed to that we are exposing to jobs which don't belong to a process definition.
The approach also adds to information consistency we are offering for jobs.

mswiderski:
it looks like XStream security related, so if it turns out there is a need to add a class or more to xstream security whitelist please add it here https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/marshalling/xstream/KieServerTypePermission.java#L39

cristianonicolai:
@bsremac which job class is causing this issue? Does that happens if you use something like:
- org.jbpm.executor.commands.PrintOutCommand
- org.jbpm.executor.commands.LogCleanupCommand
- org.jbpm.executor.commands.ExecutionErrorCleanupCommand

bsremac:
@cristianonicolai No, these correct commands won't do the trick. As I wrote in my original comment with attached screenshots, job execution needs to fail.
When it fails, then when you click on the job, the error will pop up and the job details won't be populated after you click on Ok button. The reason I tested this with a failing job, was to see how the stack trace looks now when we switched from pop-up panel to side panel, in addition to the rest of details.

I'm achieving this by setting asynchronous job within a process definition and then I just run the process with some bogus command which I assign to the job at process start. Trying to enter a bogus command via ```New Job``` dialog won't work as there's a check performed on ```Type``` field :)

nmirasch:
@cristianonicolai,  changed JobParameter table implementation using errai databinding and removed the process related info from details.
@bsremac I coudn't reproduce the xtream issue, (just to know if we need to add something to xstream security whitelist as Maciej said). My way of testing a failing job is similar but putting at the task script "throw new java.lang.RuntimeException()" and the job appears with the status ERROR and shows the stacktrace. Could you try it again please? Thanks!!

cristianonicolai:
@nmirasch now I get an error when trying to check the details:
selection_005
@bsremac got it, just check if you're running the latest snapshot from kie-server as there were several fixes regarding xstream issues lately.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsViewImpl.html
```diff
@@ -0,0 +1,41 @@
...
+            <div class="col-md-12">
+                <div class="form-horizontal">
+                    <div class="form-group" data-field="errorControlGroup">
+                        <label class="control-label col-md-4" data-i18n-key="Exceptions_Occurred"></label>
+                        <div class="col-md-8">
+                            <textarea class="col-sm-12"
+                                      rows="10" readonly data-field="job-errors"></textarea>
+                        </div>
+                    </div>
+                    <div class="form-group hidden" data-field="params-form-group">
```
nmirasch: @cristianonicolai I don't understand the error message you have, the data-field is here. Tested in my local environment it works.

bsremac:
@cristianonicolai The error is present even with the latest Kie Server snapshot. Just have in mind that I have been still running ```jbpm-wb``` with the initial state of this PR. I'm having problems building a new ```jbpm-wb``` locally due to issues with maven not finding artifacts it is looking for in nexus repo.

nmirasch:
@bsremac  I have the same issues, need to build uberfire and dashbuilder locally it seems nexus don't have the right artifacts today

nmirasch:
@cristianonicolai @bsremac  I updated the commit rebased with master and changed a typo error in the div variable linked with "params-form-group" As far as I  know it's not the cause for the error that found Cristiano, I mean "private Div paramssFormGroup" --> "private Div paramsFormGroup"

bsremac:
@cristianonicolai @nmirasch I have built ```kie-wb.war``` off of the last changes in this PR and I can't reproduce the error Cristiano found. 
One more additional piece of info when it comes to that error which I have been hitting. It happens also when a job is executed successfully, but it is part of a process instance and not an ad-hoc job. So it seems that there's something wrong with fetching details of a job in process context, in general.

jbpm-wb-executor-service/jbpm-wb-executor-service-api/src/main/java/org/jbpm/workbench/es/model/RequestSummary.java
```diff
@@ -123,6 +125,10 @@ public Integer getRetries() {
         return retries;
     }
 
+    public void setRetries(Integer retries) {
+        this.retries = retries;
+    }
+
```
bsremac: If this is the setter we want to have for ```retries```, I guess we could then delete the existing ```setRetries(int retries)```.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPresenter.java
```diff
@@ -0,0 +1,144 @@
...
+                    changeTitleWidgetEvent.fire(new ChangeTitleWidgetEvent(this.place,
+                                                                           getJobDetailTitle(requestDetails.getRequest())));
+                })
+                .getRequestDetails(serverTemplateId,
+                                   deploymentId,
+                                   jobId);
+    }
+
+    private String getJobDetailTitle(RequestSummary summary) {
+        String title = summary.getId() + " - ";
```
cristianonicolai: Can you move the extra String part (" - ") to whent key is not null(L120)? That way when key is null only the deployment id is added. This method is probably a good candidate for some isolated testing too, just to check the final String with all the variations.

jbpm-wb-executor-service/jbpm-wb-executor-service-api/src/main/java/org/jbpm/workbench/es/model/RequestSummary.java
cristianonicolai: +1

cristianonicolai:
@nmirasch exception is gone now, thanks.

@bsremac  I created a new JIRA to deal with the fobidden error as this is happening on master: https://issues.jboss.org/browse/JBPM-6615

cristianonicolai:
Related fix: https://github.com/kiegroup/droolsjbpm-integration/pull/1271

nmirasch:
@bsremac @cristianonicolai changed the jobDetailsTitle and added more test on it 
I replaced the "( deploymentId ) " by " - deploymentId" I thinks it looks a bit extrange, if you have a better solution, would be very welcome.

bsremac:
@nmirasch When it comes to Job Details title, I would rather go with the format ```Job Id - Process Name``` than ```Job Id - Deployment Id```, as Process Name seems as more suitable replacement for Business Key in this case.

nmirasch:
@bsremac the point is that processName information is not filled by the kie-server service to retrieve the job details. It is displayed at the table but if we need to show it at the details we need propagate that info modifying the event. We removed that to maintain consistency between the  service and the data displayed. I can bring that again at least for processName, make it sense @cristianonicolai?

bsremac:
@nmirasch @cristianonicolai Let's not complicate things then.
Having another look from the UI side, I realized that Business Key should never be ```null```, so each job should have ```job Id - Business Key``` part in the title anyway. 
Not sure about ```Deployment Id``` part, as I don't see much value in adding more data to the title, in general. All the details which are in the title are also present among job details, so I missing to see the use case where the title info makes the difference.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPresenter.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public void onRefresh() {
+        refreshExecutionErrorDataRemote(serverTemplateId,
+                                        deploymentId,
+                                        jobId);
+    }
+
+    public void refreshExecutionErrorDataRemote(final String serverTemplateId,
```
bsremac: ```refreshJobDetailsDataRemote``` maybe?

nmirasch: @bsremac absolutely right, Thanks!

cristianonicolai:
@nmirasch @bsremac I think the title is fine now, although the business key is mandatory when creating a new job, you still can have jobs created by the engine (ex async process task) that do not include the business key, so it would still make sense to handle not having this info.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPresenterTest.java
```diff
@@ -0,0 +1,140 @@
...
+    public void getJobDetailTitleTest() {
+        Long jobId = 1L;
+        String key = "key";
+        String deploymentId = "evaluation.1.0.1";
+        RequestSummary job = new RequestSummary();
+        job.setId(jobId);
+        job.setKey(key);
+        job.setDeploymentId(deploymentId);
+
+        assertEquals(job.getId() + " - " + job.getKey() + " - " + job.getDeploymentId(),
```
cristianonicolai: I think for these expected values, it would be more interesting to have the final full String there, even for visualizing the result I guess it makes sense.

bsremac:
@cristianonicolai Actually, I took into account the jobs which are part of the process too and it looks like in that case the engine assigns a numeric business key to the job. Have a look at the screenshot below. That 's why I concluded that ```job Id - Business Key``` part will always be present in the job details title and that maybe there's not need to expand that further.
jobinprocessbusinesskey

nmirasch:
@cristianonicolai added more assertions.

cristianonicolai:
@bsremac sure, that is auto generated but I was just not sure if it is always the case, as the core engine allows to create processes without providing a business key. Anyway, I'm fine to keep only id + business key in the title.

nmirasch:
@cristianonicolai @bsremac  I can do that. just want to add a comment...  in my test case that business key is empty.
Anyway, it you think it's better we can remove the deploymentId from title. I just wanted your confirmation

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/events/JobSelectedEvent.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public String toString() {
+        return "JobSelectedEvent{ " +
+                "serverTemplateId=" + serverTemplateId + " jobId=" + jobId + " deploymentId=" + deploymentId + '}';
+    }
+
+    @Override
+    @SuppressWarnings("PMD.AvoidMultipleUnaryOperators")
+    public int hashCode() {
```
triceo: Given how important this `~~` thing is in GWT, I keep wondering how come we haven't already created utility classes for this. Something similar to [`Objects.hash(Object... values)`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#hash(java.lang.Object...)) from the JDK.

This entire method would then just be `return Objects.hash(this.serverTemplateId, this.deploymentId, this.jobId);`

@manstis @jhrcek Do you think that this would be useful? If so, where would you place such a utility method, so that we can get everyone using it?

Anyway, just a suggestion, shouldn't block this PR.

jhrcek: Sounds reasonable to me. @manstis Any suggestion where such utility should be placed? Probably somewhere in uberfire..

bsremac:
@nmirasch On your last question. I would remove deployment Id from the title regardless of the fact your test case generates a job which doesn't have any kind of business key assigned to it. In that case we would have only plain ```job Id``` in the title and that is it (which is probably not that usual case). 
Deployment Id is also an optional value which is empty for Ad-hoc jobs. I think handling two optional values in the title might cause too much inconsistency in the title form.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/events/JobSelectedEvent.java
nmirasch: @triceo oks, thanks for reviewing!

nmirasch:
@bsremac  removed the deploymentId from the job details title.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/events/JobSelectedEvent.java
cristianonicolai: +1

PR #943, opened 2017-11-17
JBPM-6462: Process Instances: Aborting notification doesn't appear up…

nmirasch:
…on single instance abort
Added a notification when a single Process instance is aborted, modified the signaling notification to be consistent with the aborting one and updated  i18n resource bundles.(removing unused constants)


nmirasch:
@cristianonicolai @bsremac could you take a look? thanks!

PR #944, opened 2017-11-19
[JBPM-6626] Enhance Job Details tests

bsremac:
@cristianonicolai @nmirasch Enhanced unit tests as far as they could go, as I'm not planning to add any Selenium or integration tests in the future for Job Details.
Also, I introduced a small fix for NPE in case of clicking on a job which is still being shown in the Jobs list, but has already been deleted in the background by executor's log clean-up command (named test ```testStaleJobSelected```, but I'm open to suggestions for a maybe clearer name :))

### DIFF
--- a/jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPresenter.java
+++ b/jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/jobdetails/JobDetailsPresenter.java
@@ -15,6 +15,7 @@
  */
 package org.jbpm.workbench.es.client.editors.jobdetails;
 
+import java.util.List;
 import javax.enterprise.context.Dependent;
 import javax.enterprise.event.Event;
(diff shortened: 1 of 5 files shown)


cristianonicolai:
@bsremac nice, overall looks good to me, my only comment is regarding the _testJobSelectedEventObserved_ method, which I'm not sure there is much value in verifying if the annotation is present.

bsremac:
@cristianonicolai Yeah, I was having second doubts about adding this. 
The idea behind this test was to ensure that when you select a row, an action will be triggered thus replacing a classic Selenium test which performs the select operation and then waits for the expected result.
I couldn't find a better way to mimic this in unit tests realm.

cristianonicolai:
@bsremac to fully replicate that behavior, you need a CDI container in place to fire and handle the event. Usually, Arquillian would be a good candidate for that, I just think it might be more of an integration test than a unit test. That's why usually we "fake" the event with a new instance and call the method with the observer to test the logic in a unit test.

bsremac:
@cristianonicolai I understand. I guess I was trying here to go for the next best thing when the container is not used, which is definitely far from ideal :(
Deleted the test.

PR #945, opened 2017-11-22
Fix missing parameter in i18n resources for different langagues

cristianonicolai:


cristianonicolai:
@nmirasch I decided to revert part of your change because I'm afraid that if I fix the resources files manually, they could get overridden by a future sync with Zanata. Plus, adding back the full profile so Jenkins can catch similar issues in future PRs.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/resources/org/jbpm/workbench/pr/client/resources/i18n/Constants.properties
```diff
@@ -55,7 +55,7 @@ Clear=Clear
 Save=Save
 Variables_Name=Variable Name
 Variable_Value=Variable Value
-Signalling_Process_Instance=Signaling Process Instance (id={0})
+Signalling_Process_Instance=Signaling Process Instance
```
nmirasch: @cristianonicolai maybe we can remove this one? and use Signaling_Process_Instance

cristianonicolai: Sure, will send a new PR for it

PR #948, opened 2017-11-24
RHBPMS-4926: “Process 'Evaluation' was created” is shown twice in pro…

nmirasch:
…cess instance logs
Changed the log line generation to the presenter allowing i18n. Unified the service that provide the RuntimeLogSummary list. Added the 'Completed' i18n literal to nodes that have been completed to be differentiated from the creation ones.

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/DateUtils.java
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/DateUtils.java
@@ -21,6 +21,8 @@ import java.util.Date;
 import com.google.gwt.i18n.client.DateTimeFormat;
 import org.kie.workbench.common.services.shared.preferences.ApplicationPreferences;
 
+import static org.uberfire.client.views.pfly.widgets.Moment.Builder.moment;
+
 /**
  * Provides utility methods for manipulating with {@link Date}s.
  */
@@ -88,6 +90,13 @@ public class DateUtils {
         if (date != null) {
             DateTimeFormat format = DateTimeFormat.getFormat(getDateTimeFormatMask());
             return format.format(date);
+        }
+        return "";
+    }
+
+    public static String getPrettyTime(Date date) {
+        if (date != null) {
+            return moment(date.getTime()).fromNow();
         }
         return "";
     }
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
@@ -35,11 +35,9 @@ import org.kie.server.api.model.definition.ProcessDefinition;
 import org.kie.server.api.model.definition.UserTaskDefinitionList;
 import org.kie.server.api.model.instance.NodeInstance;
 import org.kie.server.api.model.instance.ProcessInstance;
-import org.kie.server.api.model.instance.TaskEventInstance;
 import org.kie.server.api.model.instance.TaskSummary;
 import org.kie.server.client.ProcessServicesClient;
 import org.kie.server.client.QueryServicesClient;
-import org.ocpsoft.prettytime.PrettyTime;
 
 import static java.util.Collections.emptyList;
 import static java.util.stream.Collectors.toList;
@@ -79,7 +77,7 @@ public class RemoteProcessRuntimeDataServiceImpl extends AbstractKieServerServic
         List<NodeInstance> nodeInstances = processServicesClient.findActiveNodeInstances(deploymentId,
                                                                                          processInstanceId,
                                                                                          0,
-                                                                                         100);
+                                                                                         Integer.MAX_VALUE);
 
         for (NodeInstance instance : nodeInstances) {
             NodeInstanceSummary summary = new NodeInstanceSummary(instance.getId(),
@@ -98,52 +96,22 @@ public class RemoteProcessRuntimeDataServiceImpl extends AbstractKieServerServic
     }
 
     @Override
-    public List<RuntimeLogSummary> getBusinessLogs(String serverTemplateId,
-                                                   String deploymentId,
-                                                   String processName,
-                                                   Long processInstanceId) {
+    public List<RuntimeLogSummary> getProcessInstanceLogs(String serverTemplateId,
+                                                          String deploymentId,
+                                                          Long processInstanceId) {
         if (serverTemplateId == null || serverTemplateId.isEmpty()) {
             return emptyList();
         }
 
-        List<NodeInstance> processInstanceHistory = getProcessInstanceHistory(serverTemplateId,
-                                                                              deploymentId,
-                                                                              processInstanceId);
-        List<TaskEventInstance> allTaskEventsByProcessInstanceId = new ArrayList<TaskEventInstance>();//taskAuditService.getAllTaskEventsByProcessInstanceId(processInstanceId, "");
-        List<RuntimeLogSummary> logs = new ArrayList<RuntimeLogSummary>(processInstanceHistory.size() + allTaskEventsByProcessInstanceId.size());
-        PrettyTime prettyDateFormatter = new PrettyTime();
+        ProcessServicesClient processServicesClient = getClient(serverTemplateId,
+                                                                ProcessServicesClient.class);
 
-        for (int i = processInstanceHistory.size() - 1; i >= 0; i--) {
-            NodeInstance nis = processInstanceHistory.get(i);
+        List<NodeInstance> processInstanceHistory = processServicesClient.findNodeInstances(deploymentId,
+                                                                                            processInstanceId,
+                                                                                            0,
+                                                                                            Integer.MAX_VALUE);
 
-            if (nis.getNodeType().equals("HumanTaskNode")) {
-                logs.add(new RuntimeLogSummary(nis.getId(),
-                                               prettyDateFormatter.format(nis.getDate()),
-                                               "Task '" + nis.getName() + "' was created",
-                                               "System"));
-                for (TaskEventInstance te : allTaskEventsByProcessInstanceId) {
-                    if (te.getWorkItemId() != null && te.getWorkItemId().equals(nis.getId()) &&
-                            (te.getType().equals("CLAIMED") || te.getType().equals("RELEASED") || te.getType().equals("COMPLETED"))) {
-                        logs.add(new RuntimeLogSummary(nis.getId(),
-                                                       "- " + prettyDateFormatter.format(te.getLogTime()),
-                                                       "Task '" + nis.getName() +
-                                                               "' was " + te.getType().toLowerCase() + " by user " + te.getUserId(),
-                                                       "Human"));
-                    }
-                }
-            } else if (nis.getNodeType().equals("StartNode")) {
-                logs.add(new RuntimeLogSummary(nis.getId(),
-                                               prettyDateFormatter.format(nis.getDate()),
-                                               "Process '" + processName + "' was created",
-                                               "Human"));
-            } else if (nis.getNodeType().equals("EndNode")) {
-                logs.add(new RuntimeLogSummary(nis.getId(),
-                                               prettyDateFormatter.format(nis.getDate()),
-                                               "Process '" + processName + "' was completed",
-                                               "System"));
-            }
-        }
-        return logs;
+        return processInstanceHistory.stream().map(new RuntimeLogSummaryMapper()).collect(toList());
     }
 
     @Override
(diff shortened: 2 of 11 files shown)


nmirasch:
@bsremac @cristianonicolai  could you take a look please? thanks!

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
```diff
@@ -98,51 +97,28 @@ public ProcessInstanceSummary getProcessInstance(String serverTemplateId,
...
-                                               "System"));
-            }
-        }
+        List<NodeInstance> processInstanceHistory = processServicesClient.findNodeInstances(deploymentId,
+                                                                                            processInstanceId,
+                                                                                            0,
+                                                                                            100);
+        List<RuntimeLogSummary> logs = new ArrayList<>();
+        PrettyTime prettyDateFormatter = new PrettyTime();
+        processInstanceHistory.forEach(nis -> logs.add(new RuntimeLogSummary(nis.getId(),
```
cristianonicolai: Perhaps use .map with a mapper class?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,55 +59,83 @@ public IsWidget getWidget() {
...
+    }
+
     public void refreshProcessInstanceData(final LogOrder logOrder,
                                            final LogType logType) {
+        processRuntimeDataService.call((List<RuntimeLogSummary> logs) -> {
+            final List<String> logsLine = new ArrayList<String>();
+            if (logOrder == LogOrder.ASC) {
+                Collections.reverse(logs);
+            }
+            logs.stream().forEach(rls -> addLogline(logsLine,
```
cristianonicolai: Same here, addLogline could be changed into a mapper function to be used in .map in the logs stream. That way you can also easily test the String result for a single line.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,55 +59,83 @@ public IsWidget getWidget() {
...
-                    view.setLogs(logsLine);
-                }
-            }).getRuntimeLogs(currentServerTemplateId,
-                              currentDeploymentId,
-                              currentProcessInstanceId);
+            String agent = constants.System();
+            if (NODE_HUMAN_TASK.equals(logSummary.getNodeType()) && logSummary.isCompleted()) {
+                agent = constants.Human();
+            } else if (NODE_START.equals(logSummary.getNodeType()) && !logSummary.isCompleted()) {
+                agent = constants.Human();
```
cristianonicolai: Same agent as Human Task?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,55 +59,83 @@ public IsWidget getWidget() {
...
+            }
+            logLines.add(logSummary.getPrettyTime() + ": " +
```
cristianonicolai: I think it makes a bit more readable if you use String.format instead of concatenating all these Strings together...

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
```diff
@@ -98,51 +97,28 @@ public ProcessInstanceSummary getProcessInstance(String serverTemplateId,
...
-                                               "Process '" + processName + "' was completed",
-                                               "System"));
-            }
-        }
+        List<NodeInstance> processInstanceHistory = processServicesClient.findNodeInstances(deploymentId,
+                                                                                            processInstanceId,
+                                                                                            0,
+                                                                                            100);
+        List<RuntimeLogSummary> logs = new ArrayList<>();
+        PrettyTime prettyDateFormatter = new PrettyTime();
```
cristianonicolai: We can't really format the date in here because it won't consider the user locale, we should try to reuse the Moment API at the presenter level for this.

nmirasch:
@cristianonicolai @bsremac this PR depends on https://github.com/kiegroup/appformer/pull/107 for the moment API interaction changes.  Could you take a look?
I did some changes taking in account your comments @cristianonicolai, the only thing I couldn't achieve was the String.format, because it's not available in GWT client side. After trying different things, I added a private function that solves our issue there, but I think it's not strong enough. Maybe you have a better idea about how to do that String formatting... Another option would be  create different i18n literal for each combination, but I think it's not too transparent for the translations.
Any idea is very welcome.
Thanks!

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,60 +65,117 @@ public IsWidget getWidget() {
...
+                              getCurrentProcessName(),
+                              constants.WasCompleted());
+            }
+        }
+        return null;
+    }
+
+    private String format(String pattern,
+                          Object... params) {
+        String formattedText = pattern;
```
cristianonicolai: @nmirasch true, String.format isn't available, you could give a go to GWT own templating then: http://www.gwtproject.org/doc/latest/DevGuideSecuritySafeHtml.html#SafeHtmlTemplates

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,60 +65,117 @@ public IsWidget getWidget() {
...
+    }
+
+    public void setCurrentDeploymentId(String currentDeploymentId) {
+        this.currentDeploymentId = currentDeploymentId;
+    }
+
     public void refreshProcessInstanceData(final LogOrder logOrder,
                                            final LogType logType) {
+        processRuntimeDataService.call((List<RuntimeLogSummary> logs) -> {
+            //List<String> logsLine = new ArrayList<String>();
```
cristianonicolai: remove? :)

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,60 +65,117 @@ public IsWidget getWidget() {
...
-                        logsLine.add(rls.getTime() + ": " + rls.getLogLine());
-                    }
-
-                    view.setLogs(logsLine);
-                }
-            }).getBusinessLogs(currentServerTemplateId,
-                               currentDeploymentId,
-                               currentProcessName,
-                               currentProcessInstanceId);
+            String prettyTime = moment(DateTimeFormat.getFormat("yyyy-MM-dd hh:mm:ss a")
```
cristianonicolai: See my comment on the other PR, just wondering if you could create the moment object without parsing the date, either by using the date object or date..getTime()

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,60 +65,117 @@ public IsWidget getWidget() {
         return view;
     }
 
+    public void setCurrentProcessInstanceId(Long currentProcessInstanceId) {
```
bsremac: Since you are re-factoring the whole thing I'm thinking of removing this ```current``` part from the field names, as it seems redundant.

cristianonicolai: do we even need these setter methods? In theory, the only place where the process instance data should change is when receiving an ProcessInstanceSelectionEvent.

nmirasch:
@cristianonicolai @bsremac added the requested changes.

cristianonicolai:
jenkins please retest this

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/test/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImplTest.java
```diff
@@ -232,4 +234,56 @@ public void testGetProcess() {
...
+                     logs.size());
+        assertRuntimeLogNodeInstance(nodeInstance1,
+                                     logs.get(0));
+        assertRuntimeLogNodeInstance(nodeInstance2,
+                                     logs.get(1));
+    }
+
+    private void assertRuntimeLogNodeInstance(NodeInstance node,
+                                              RuntimeLogSummary log) {
+        assertTrue(node.getId() == log.getId());
```
cristianonicolai: perhaps use assertEquals in here as well?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,64 +67,119 @@ public IsWidget getWidget() {
...
+                    .filter(Objects::nonNull)
+                    .collect(toList());
+
+            view.setLogs(logsLine);
+        }).getProcessInstanceLogs(serverTemplateId,
+                                  deploymentId,
+                                  processInstanceId);
+    }
 
+    protected String getLogline(RuntimeLogSummary logSummary,
```
cristianonicolai: _getLogLine_

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,64 +67,119 @@ public IsWidget getWidget() {
...
-                    }
-
-                    view.setLogs(logsLine);
-                }
-            }).getBusinessLogs(currentServerTemplateId,
-                               currentDeploymentId,
-                               currentProcessName,
-                               currentProcessInstanceId);
+
+            String prettyTime = moment().milliseconds(logSummary.getDate().getTime() - new Date().getTime()).fromNow();
```
cristianonicolai: You probably don't need to do getTime minus current getTime, fromNow should tell you that already...

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenter.java
```diff
@@ -53,64 +67,119 @@ public IsWidget getWidget() {
...
+    }
+
     public interface RuntimeLogView extends IsWidget {
 
         void init(final RuntimeLogPresenter presenter);
 
         void setLogs(List<String> logs);
     }
+
+    public interface LogTemplates extends SafeHtmlTemplates {
```
cristianonicolai: Nice :)

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/test/java/org/jbpm/workbench/pr/client/editors/instance/log/RuntimeLogPresenterTest.java
```diff
@@ -0,0 +1,474 @@
...
+
+    private Date logDate = new Date();
+    private String prettyTime;
+    private String techTime;
+
+    @Before
+    public void setup() {
+        processRuntimeDataServiceCaller = new CallerMock<ProcessRuntimeDataService>(processRuntimeDataService);
+        presenter.setProcessRuntimeDataService(processRuntimeDataServiceCaller);
+        prettyTime = moment().milliseconds(logDate.getTime() - new Date().getTime()).fromNow();
```
cristianonicolai: Calling moment API in the Java test probably won't return any value as it relies on JavaScript to work, so I'm not sure there is much value to do this call as prettyTime will always be an empty String.

nmirasch:
@bsremac @cristianonicolai I tried to encapsulate the moment API initialization at DateUtils exposing a simple getPrettyTime and passing the date.
I avoided the sorting to let the 'end node' complete at the end because I found that the timestamp sometimes is previous (and not exactly equals) to other completed nodes.
Please could you take a look? Thanks!!!!

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/DateUtils.java
```diff
@@ -92,4 +94,25 @@ public static String getDateTimeStr(Date date) {
         return "";
     }
 
+    public static String getPrettyTime(Date date) {
```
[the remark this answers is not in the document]
nmirasch: The idea was use DateUtils class for all data formatting interaction, and encapsulate there the moment API interaction. Maybe the thing here is that 'public static class MomentWrapper' have to be private to not to be exposed. What do you think?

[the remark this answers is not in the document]
bsremac: Yes, that's exactly what my comment was all about.
If we want to have ```getPrettyTime``` method be part of ```DateUtils``` interface at the same encapsulation level as all other methods there, then we should make it clear from the code that ```MomentWrapper``` methods are private and hence avoid any confusion which method is expected to be for data formatting.

nmirasch:
@bsremac added the optional implementation you commented. Not sure it its more readable, but  just allow us not work with null values. Could you take another look?

bsremac:
@nmirasch Yes, it looks like a trade-off between avoiding returning ```null``` and having the code look nicer at the point where the method's return value is used.
From readability perspective, I think both solutions are equally readable. It just comes down to whether we want to continue returning ```nulls``` from methods even now when there's an alternative in the form of ```Optional```.
Both approaches are valid, so whichever one we choose, there's no harm done :)

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/DateUtils.java
```diff
@@ -92,4 +94,25 @@ public static String getDateTimeStr(Date date) {
...
+
+    private static class MomentWrapper {
+
+        public static String DATE_FORMAT_MASK = "dd/MM/yyyy HH:mm:ss";
+        public static String MOMENT_FORMAT_MASK = "DD/MM/YYYY HH:mm:ss";
+
+        public static String getPrettyTime(Date date) {
+            DateTimeFormat format = DateTimeFormat.getFormat(DATE_FORMAT_MASK);
+
+            return moment(format.format(date),
```
cristianonicolai: @nmirasch I'm wondering if you use the moment constructor passing the Date object or the time in milliseconds (long) you could avoid parsing the date into different formats. Have you tried that?

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/DateUtils.java
```diff
@@ -92,4 +94,25 @@ public static String getDateTimeStr(Date date) {
...
+                          MOMENT_FORMAT_MASK)
+                    .utc()
```
cristianonicolai: Why is .utc needed in here?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
```diff
@@ -98,52 +96,22 @@ public ProcessInstanceSummary getProcessInstance(String serverTemplateId,
...
-            }
-        }
-        return logs;
+        ProcessServicesClient processServicesClient = getClient(serverTemplateId,
+                                                                ProcessServicesClient.class);
+
+        List<NodeInstance> processInstanceHistory = processServicesClient.findNodeInstances(deploymentId,
+                                                                                            processInstanceId,
+                                                                                            0,
+                                                                                            100);
```
cristianonicolai: @mswiderski we got a few places like in here where we need the complete list of items to be retrieved from a paginated service response. Just having a limited/fixed number of items to retrieve isn't a good solution. I wonder if we could define something like -1 for page size to bypass pagination and return all items to avoid multiple calls or that is something you would really prefer to avoid?

cristianonicolai:
@nmirasch overall looks good, I'm just concerned about that fixed number of node history we use for calling the kie server. That is an issue already so I think it would be fine to open a new JIRA to handle that separately if you want. 
Also, regarding the sorting that we discussed to try to get the logs better organized in the UI, is that something you still planning for this PR?

nmirasch:
@cristianonicolai I avoided the sorting we commented because I found that they don't have exactly the same timestamp. There are completed endNodes that are before than completed humanNodes. I thought it was dangerous change the order in visualization that could no contemplate all cases. Right now they are sorted as they come from the server. Maybe is a thing to fix there @mswiderski ?

nmirasch:
@cristianonicolai replaced the page size from 100 to  Integer.MAX_VALUE at backed service calls. Could you take a look? thanks!

PR #954, opened 2017-12-01
JBPM-6664 - CannotResolveClassException thrown when accessing active …

mswiderski:
…case

@cristianonicolai @bsremac this is in general a quick fix to not load case data as show case app does not really use that so there is no need to load it.

When it comes to long term solution, we need to have kie server client per containers deployed to kie server because data model that is inside kjar won't be found by showcase. Same thing was required in workbench and when you look at this in workbench you can actually work with the Survey class there.

So I suggest to merge this in for 7.5 community and maybe later on look for more advanced cases... though not sure this needs to be done since the case app is showcase only.

### DIFF
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
@@ -150,7 +150,7 @@ public class RemoteCaseManagementServiceImpl implements CaseManagementService {
                                                final String caseId) {
         return ofNullable(client.getCaseInstance(containerId,
                                                  caseId,
-                                                 true,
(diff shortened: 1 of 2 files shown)


bsremac:
@mswiderski @cristianonicolai Completely agree with the approach here, taking into account the current Casey's (nick we introduced for the app :)) capabilities.
The main motivation behind raising the issue was to provide as seamless as possible experience when using Casey with the only example case project we have at the moment and get rid of all the errors. I assume trying out the example projects is what a showcase app should be all about.

mswiderski:
the test failure does not seem to be related to my changes.

Jenkins retest this

mswiderski:
@cristianonicolai I guess we are waiting for this https://github.com/kiegroup/jbpm-wb/pull/956 to be merged, right?

cristianonicolai:
@mswiderski exactly, triggering build again now.

mswiderski:
Jenkins retest this

cristianonicolai:
The previous fix didn't really fix the integration test issue, trying with new PR https://github.com/kiegroup/jbpm-wb/pull/957

cristianonicolai:
Jenkins retest this

PR #965, opened 2017-12-26
JBPM-6174: Add error count column to the task admin list

mcivantos-tribalyte:
Add error count column to the task admin list


cristianonicolai:
ok to test

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridPresenter.java
```diff
@@ -16,6 +16,10 @@
 
 package org.jbpm.workbench.common.client.list;
 
+import static org.jbpm.workbench.common.client.PerspectiveIds.SEARCH_PARAMETER_IS_ERROR_ACK;
+import static org.jbpm.workbench.common.client.PerspectiveIds.SEARCH_PARAMETER_PROCESS_INSTANCE_ID;
+import static org.kie.workbench.common.workbench.client.PerspectiveIds.EXECUTION_ERRORS;
```
cristianonicolai: perhaps also use _org.jbpm.workbench.common.client.PerspectiveIds_ in here to keep consistency.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/AbstractTaskListView.java
```diff
@@ -87,9 +97,11 @@ public String getStyleNames(TaskSummary row,
     @Override
     public void initColumns(ExtendedPagedTable<TaskSummary> extendedPagedTable) {
         initCellPreview(extendedPagedTable);
-
+        
         Column actionsColumn = initActionsColumn();
+        Column<TaskSummary, ?> errorCountColumn = initErrorCountColumn();
```
cristianonicolai: The way is setup in here, the error columns gets visible to be used in the normal _Task List_, that should be avoided so it only appears in the _Task Admin List_

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskSummaryDataSetMapper.java
```diff
@@ -29,62 +29,126 @@
     @Override
     public TaskSummary apply(final DataSet dataSet,
```
cristianonicolai: I guess in this case is best to split this in two specific mapper classes, one for each data set.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-api/src/main/java/org/jbpm/workbench/ht/model/TaskSummary.java
```diff
@@ -45,6 +45,7 @@
     private Boolean isLogOnly = Boolean.FALSE;
     private Long parentId;
     private Map<String, String> domainData = new HashMap<>();
+    private Integer errorCount;
```
cristianonicolai: Should this class extend _GenericErrorSummary_ instead?

jbpm-wb-common/jbpm-wb-common-api/src/main/java/org/jbpm/workbench/common/model/GenericErrorSummary.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.workbench.common.model;
+
+public class GenericErrorSummary extends GenericSummary<Long>{
```
cristianonicolai: Please make this class abstract, also receiving a generic id type as in the _GenericSummary_.
Other important fixes in here should include, adding equals, hashcode and toString methods.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridPresenter.java
```diff
@@ -150,6 +154,15 @@ protected void navigateToPerspective(final String perspectiveId,
                              parameterValue);
         placeManager.goTo(request);
     }
+    
+    public void openView(final String itemId) {
```
cristianonicolai: This is probably too specific to be added in this class, ideally this should be in the target presenter classes.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
```diff
@@ -82,8 +87,9 @@ public void render(Context context,
         sb.appendHtmlConstant(tooltip.getOuterHTML());
     }
 
-    public void openErrorView(final String pid) {
-        viewPresenter.openErrorView(pid);
+    public void openErrorView(final String pid){
+        viewPresenter.openView(pid);
```
cristianonicolai: If you reuse _openView_ in here you will get an error list filtered by process instance id, when I believe we actually want is to filter by the task id,

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskslist/AbstractTaskListViewTest.java
```diff
@@ -73,7 +73,7 @@ public void initColumnsWithTaskVarColumnsTest() {
             @Override
             public Void answer(InvocationOnMock invocationOnMock) throws Throwable {
                 final List<ColumnMeta> columns = (List<ColumnMeta>) invocationOnMock.getArguments()[0];
-                assertTrue(columns.size() == 18);
+                assertTrue(columns.size() == 19);
```
bsremac: Can you please replace _assertTrue_ method with _assertEquals_, in the way it's used in _DataSetDefsBootstrapTest_ where you did the argument modification from 19 to 20?
If the assertion fails, the assertion error returned is much clearer in that case.

mcivantos-tribalyte: Done

mcivantos-tribalyte:
@cristianonicolai All your request changes are done.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminListPresenter.java
```diff
@@ -58,9 +63,73 @@ public FilterSettings createTableSettingsPrototype() {
...
+        }
+        builder.group(COLUMN_TASK_ID);
+
+        addCommonColumnSettings(builder);
+        return builder.buildSettings();
+    }
+    
+    @Override
+    protected void addCommonColumnSettings(FilterSettingsBuilderHelper builder) {
+        builder.setColumn(COLUMN_ACTIVATION_TIME,
```
cristianonicolai: I guess you can just add the error column in here and call super to add the remaining?

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminListViewImpl.java
```diff
@@ -16,15 +16,46 @@
...
 
     private static final String DATA_SET_TASK_LIST_PREFIX = "DataSetTaskAdminGrid";
     protected static final String TAB_ADMIN = DATA_SET_TASK_LIST_PREFIX + "_0";
+    
+    @Inject
+    private ManagedInstance<GenericErrorSummaryCountCell> popoverCellInstance;
+
 
+    @Override
+    public List<String> getInitColumns() {
```
cristianonicolai: because we now have two distinct implementations, _getInitColumns_ should also be removed from _AbstractTaskListView_ and added to _TaskListViewImpl_

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminListViewImpl.java
```diff
@@ -41,5 +72,126 @@ public void initDefaultFilters() {
     public String getGridGlobalPreferencesKey() {
         return DATA_SET_TASK_LIST_PREFIX;
     }
+    
+    @Override
+    public void initColumns(ExtendedPagedTable<TaskSummary> extendedPagedTable) {
+        initCellPreview(extendedPagedTable);
```
cristianonicolai: ideally, this method should be shared with the parent class to avoid duplication. only the error column is new here.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminSummaryDataSetMapper.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.function.BiFunction;
+
+import org.dashbuilder.dataset.DataSet;
+import org.jbpm.workbench.ht.model.TaskSummary;
+
+import static org.jbpm.workbench.common.client.util.DataSetUtils.*;
+import static org.jbpm.workbench.ht.model.TaskDataSetConstants.*;
+
+public class TaskAdminSummaryDataSetMapper implements BiFunction<DataSet, Integer, TaskSummary> {
```
bsremac: Do we really need two separate mappers here?
Both mappers are returning the same type (_TaskSummary_) and there's a difference in only one additional setter and the boolean value _isForAdmin_ method is taking as an argument.

So what do you think, for example, about having mapper which doesn't implement _BiFunction_, but rather exposes custom _apply_ method which would take three arguments: 
```
apply(final DataSet dataSet, 
      final Integer row, 
      final boolean isForAdmin)
```
@cristianonicolai Wdyt?

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskslist/AbstractTaskListPresenterTest.java
```diff
@@ -372,8 +372,15 @@ public void testTaskSummaryAdmin() {
...
-            final TaskSummary summary = new TaskSummaryDataSetMapper().apply(dataSetMock,
-                                                                             0);
+            final TaskSummary summary;
+            if(HUMAN_TASKS_WITH_ADMIN_DATASET.equals(dataSet)) {
+                summary = new TaskAdminSummaryDataSetMapper().apply(dataSetMock,
+                                                                                      0); 
+            }else {
+                summary = new TaskSummaryDataSetMapper().apply(dataSetMock,
+                                                                                      0);
+            }
```
bsremac: Indentations broken, else without preceding space.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminListViewImplTest.java
```diff
@@ -54,11 +77,57 @@ public AbstractTaskListPresenter getPresenter() {
         return Arrays.asList(TAB_SEARCH,
                              TAB_ADMIN);
     }
+    
+    @Override
+    public List<String> getExpectedInitialColumns() {
```
bsremac: For consistency reasons, this method should be overridden in _TaskListViewImplTest_ too, by copying over the implementation from _AbstractTaskListViewTest_ and leaving the method also abstract in _AbstractTaskListViewTest_ .

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminSummaryDataSetMapper.java
mcivantos-tribalyte: @bsremac I had only one mapper at first, but @cristianonicolai suggested having two separate mappers.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
```diff
@@ -65,10 +73,10 @@ public void render(Context context,
             tooltipClasses.remove(ERROR_PRESENT_STYLE);
         }
 
-        if (viewPresenter.getViewErrorsActionCondition().test(value)) {
+        if (viewPresenter.isUserAuthorizedForPerspective(EXECUTION_ERRORS) && value.getErrorCount() != null && value.getErrorCount() > 0) {
```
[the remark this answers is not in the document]
mcivantos-tribalyte: @bsremac The purpose of the second and third condition is showing a tooltip with the message "Click to view {n} error" when the number of errors is more than 0 or it's not null. Otherwise nothing is shown (in case of zero or null).
This line set the value of the errorCount in the cell, in order to set the value to zero when is null or the correct value in any other cases.
https://github.com/mcivantos-tribalyte/jbpm-wb/blob/5f4fc0f2772210b94a0d683571b3a2acf120b8f2/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java#L65

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminSummaryDataSetMapper.java
bsremac: @mcivantos-tribalyte I suppose the reason is having the consistent implementation of all mapper classes, I can't see any other reason to be honest, since the "functional" part of the mapper implementation is not even being used.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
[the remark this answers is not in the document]
bsremac: @mcivantos-tribalyte I see now where ```errorCount``` is used inside _if - else_ clause.
However, most of my previous comment still holds.
The call to ```value.getErrorCount()``` is already processed in line 65 and if you use ```errorCount``` instead of ```value.getErrorCount()``` in your condition, the current condition can be reduced to:
```
if (viewPresenter.isUserAuthorizedForPerspective(EXECUTION_ERRORS) && errorCount > 0)
```
Furthermore, the satisfaction of the condition ```errorCount > 0``` has already been addressed in line 70 when it comes to tooltip visibility and behavior, so I'm not sure why aren't these two conditional branches merged into one.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminSummaryDataSetMapper.java
cristianonicolai: @mcivantos-tribalyte @bsremac I agree, the mapper implementation isn't significantly different to justify a separated class. I guess I initially thought of this without checking the columns that are used in each case. So my primary concern is to not repeat this entire logic for both scenarios. 
Perhaps something as simple as this could work?
```java
.errorCount(HUMAN_TASKS_WITH_ADMIN_DATASET.equals(dataSet.getUUID()) ? getColumnIntValue(dataSet,
                                                                                                             COLUMN_ERROR_COUNT,
                                                                                                             row) : null)
                    .isForAdmin(HUMAN_TASKS_WITH_ADMIN_DATASET.equals(dataSet.getUUID()))
´´´

**Comment by mcivantos-tribalyte** [comment]
@cristianonicolai  @bsremac  all your requested changes done.

File: jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java [hunk 16]
```diff
@@ -82,8 +90,13 @@ public void render(Context context,
...
     }
 
-    public void openErrorView(final String pid) {
-        viewPresenter.openErrorView(pid);
+    public void openErrorView(final String pid){
+        final PlaceRequest request = new DefaultPlaceRequest(EXECUTION_ERRORS);
+        request.addParameter(SEARCH_PARAMETER_PROCESS_INSTANCE_ID,
+                             pid);
+        request.addParameter(SEARCH_PARAMETER_IS_ERROR_ACK,
+                             Boolean.toString(false));
```
bsremac: @mcivantos-tribalyte @cristianonicolai I think we need to add another parameter to the active filters list in this case, so it matches the DB query we making in _DataSetDefsBootstrap_ and that is **Type: Task**.
Otherwise, navigation to Execution errors view using error count tooltip can show a different number of errors from the one errors column is showing in the task admin view, which would look inconsistent.

cristianonicolai: +1

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/TaskAdminListViewImpl.java
```diff
@@ -41,5 +68,48 @@ public void initDefaultFilters() {
...
+        
+        Column<TaskSummary, ?> errorCountColumn = initErrorCountColumn();
+        extendedPagedTable.addSelectionIgnoreColumn(errorCountColumn);
+        
+        extendedPagedTable.addColumn(errorCountColumn, constants.Errors());
+        extendedPagedTable.setColumnWidth(errorCountColumn,
+                                          65,
+                                          Style.Unit.PX);
+        
+    }*/
```
bsremac: @mcivantos-tribalyte This commented out code should be removed completely, if it is superfluous.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
mcivantos-tribalyte: Done.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
```diff
@@ -64,32 +74,43 @@ public void render(Context context,
...
-        viewPresenter.openErrorView(pid);
+    public void openErrorView(final String id){
+        final PlaceRequest request = new DefaultPlaceRequest(EXECUTION_ERRORS);
+        
+        if(TASK_TYPE.equals(type)) {
+            request.addParameter(SEARCH_PARAMETER_TASK_ID,
+                                 id);
+        } else {
+            request.addParameter(SEARCH_PARAMETER_PROCESS_INSTANCE_ID,
+                                 id);
```
bsremac: @mcivantos-tribalyte I'm failing to see the benefits of the increased complexity here. Why are we checking the error type when it is clear form the DB query that the error type we are exclusively interested in is the task one:
```"errInfo.error_type=\'Task\'"```
This means that the error count associated with the specific task will only show the number of Task type errors. This is the information we want to show to task admins, right?
When we click on that number, the Execution errors view needs to match that number in order to maintain the consistency between the error count displayed in the admin task list and the number of errors listed in Execution errors view. To achieve this, the parameters being added to the request have to match the DB query in _DataSetDefsBootstrap_
Am I missing some use case scenario?

Comment (mcivantos-tribalyte) [reply jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java]: @bsremac  Yes, you are right, the information we want to show in Task Admin view is the number of Task type error. So, this query (https://github.com/kiegroup/jbpm-wb/pull/965/files#diff-51ef7bbb4674fdc3238f8ec3494c30a6L116) is related to the Task Admin view where we are already filtering by "errInfo.error_type=\'Task\'", as you said.
When we click on that number we navigate to the Execution errors view. There, the query is (https://github.com/kiegroup/jbpm-wb/blob/1aea7ee22f5468f2bcaaef0936ad4f2041c99486/jbpm-wb-executor-service/jbpm-wb-executor-service-backend/src/main/java/org/jbpm/workbench/es/backend/server/DataSetDefsBootstrap.java#L92) and we were filtering by ProcessInstanceId when we navigate from Process List View and I added  request.addParameter(SEARCH_PARAMETER_TASK_ID, id); in order to filter by the taskId when you are clicking when we navigate from Task Admin View. 
So the use case scenarios that I've understood are:
- Navigate from Process List view, in that case we have to show all the errors related to this process.
- Navigate from Task Admin View: in that case we have to show only the errors related to this task.
I guess, you suggest that the correct behaviour is showing all the task errors related to one process instead of the errors related to the task you are selecting when clicking.
@cristianonicolai  what's the correct behaviour?

Comment (bsremac) [reply jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java]: @mcivantos-tribalyte No, I'm not suggesting any change in the DB query, with that being retrieving only task type errors for the particular task we are selecting. If that is what we want, that is fine. I just wanted to confirm that based on your latest changes, because a task can generate various types of errors (not only Type: Task) depending on the process design, task work item handlers etc.

Now going back to the actual implementation of the navigation to Execution Errors. What looks troubling in the introduced logic is that we are now deciding on the set of active filters in Execution Errors view based on some task type parameter we introduced just so we are able to differentiate where we are navigating from.
My suggestion is to either:
1. Split this method into 2 methods in which case one would handle only navigation from Process Instances View and the other only navigation from Task Admin View or
2.  If we want to stay on the track of having one universal method which will cover both navigation paths, introduce a clearer (unambiguous) criterion based on which the differentiation is being made. 
The current implementation assumes that a task can generate only task type error. This means that if a certain task has generated a task error, but also some other at the same time, when we navigate to Execution Errors view the list will show both of them and not only task type error (because both errors belong to the same task id).

Comment (mcivantos-tribalyte) [reply jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java]: @bsremac Ok, I understand it. In the jira related to this PR, we only talk about Error_Type = 'Task' (https://issues.jboss.org/browse/JBPM-6174). So, I am going to implement the changes you suggest on the navigate to Execution Errors method in order to show only  Error_Type = 'Task'.

Comment (mcivantos-tribalyte) [reply jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java]: @bsremac  Done

File: jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java [hunk 19]
```diff
@@ -104,8 +125,10 @@ private native void initTooltips(String dataAttrVal,
                 .tooltip()
                 .on("click", function () {
                     var processInstId = $wnd.jQuery(this).attr(procIdAttrName);
-                    thisCellRef.@org.jbpm.workbench.pr.client.editors.instance.list.ProcessInstanceSummaryErrorCountCell::openErrorView(Ljava/lang/String;)(processInstId);
+                    thisCellRef.@org.jbpm.workbench.common.client.util.GenericErrorSummaryCountCell::openErrorView(Ljava/lang/String;)(processInstId);
```
bsremac: I don't know much about JS and this looks to me as a purely cosmetic comment, but for readability reasons it would be good to change the variable name in the last bracket from ```(processInstId)``` to just ```(id)```.

mcivantos-tribalyte: Done

cristianonicolai:
jenkins please retest this

bsremac:
@mcivantos-tribalyte I have just realized that _GenericErrorSummaryCountCell.html_has got a hard-coded class value ```process-instance-error-count```. It would be good to have that class dynamically generated depending on the view. For the task view it could be, for example: ```task-error-count```

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
cristianonicolai: @mcivantos-tribalyte in here, I believe this class should not dependend on the domain ( task or process ), you could simply delegate back to the presenter that would know which paramters to add into the place request. Same goes for the _TASK_INSTANCE_ATTRIBUTE_ and _PROCESS_INSTANCE_ATTRIBUTE_ that could be generic as _data-jbpm-id_.

bsremac:
@mcivantos-tribalyte @cristianonicolai We have another problem.
When navigated to Execution Errors view from Task Admin view, the advanced search doesn't work properly for subsequent queries.
Look at this example. We have a process instance which generated two errors. One is of type _DB_ and the other is of type _Task_.
If we navigate to Execution Errors view from Process Instances view, using the tool tip, we can see both errors:
fromprocessinstances
However, if we navigate to Execution Errors view from Task Admin view, using the tool tip, and then we modify the advanced search query so it matches the one we have when navigating from Process Instances view, the _DB_  type error is not listed.
fromtaskadmin

cristianonicolai:
@mcivantos-tribalyte @bsremac regarding your last comment, I believe the issue is in this code: https://github.com/mcivantos-tribalyte/jbpm-wb/blob/259a627a2f3294c61ea12accd6f707eaeaf04351/jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenter.java#L366

What happens is that the condition in _removeAdvancedSearchFilter_ needs to match with _addAdvancedSearchFilter_ otherwise the filter is not properly removed. 

In this case, I guess the best fix is to add to individual filters, one for type and another for the task id.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
```diff
@@ -64,37 +71,41 @@ public void render(Context context,
...
     }
 
-    public void openErrorView(final String pid) {
-        viewPresenter.openErrorView(pid);
+    public void openErrorView(final String id){
+        final PlaceRequest request = new DefaultPlaceRequest(EXECUTION_ERRORS);
+        
+        request.addParameter(SEARCH_PARAMETER_IS_ERROR_ACK,
+                             Boolean.toString(false));
+        viewPresenter.openView(request,id);
```
cristianonicolai: @mcivantos-tribalyte I still think that all of these should be moved to the presenter, in here you should only call the _openErrorView_ in the presenter.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
```diff
@@ -64,37 +71,41 @@ public void render(Context context,
         } else {
             tooltipClasses.remove(ERROR_PRESENT_STYLE);
         }
-
-        if (viewPresenter.getViewErrorsActionCondition().test(value)) {
+        
+        if (viewPresenter.isUserAuthorizedForPerspective(EXECUTION_ERRORS) && errCount > 0) {
```
cristianonicolai: @mcivantos-tribalyte can we keep the previous format with a condition.test to evaluate this?

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
bsremac: +1

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridPresenter.java
```diff
@@ -150,6 +150,14 @@ protected void navigateToPerspective(final String perspectiveId,
                              parameterValue);
         placeManager.goTo(request);
     }
+    
+    public void openView(final PlaceRequest request) {
+        placeManager.goTo(request);
+    }
+    
+    public void openView(final PlaceRequest request, final String parameterValue) {
+        placeManager.goTo(request);
```
bsremac: This method really looks odd. It does exactly the same thing as the one above, but for some reason takes two arguments.
If we can't avoid having this two-arguments method here, can we at least make it abstract and rely on overriding it wherever it is used?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/errorlist/ExecutionErrorListPresenter.java
```diff
@@ -353,6 +353,36 @@ public void setupActiveSearchFilters() {
             isDefaultFilters = false;
         }
         
+        final Optional<String> taskIdSearch = getSearchParameter(PerspectiveIds.SEARCH_PARAMETER_TASK_ID);
+        if (taskIdSearch.isPresent()) {
+            final String taskId = taskIdSearch.get();
+            view.addActiveFilter(constants.Task(),
+                                 taskId,
+                                 taskId,
```
bsremac: Why are we passing the same argument twice?

mcivantos-tribalyte: The first one is the label value that is shown in the top of the view and the second one is the real value by which we are filtering. I guess that in this case it's the same value, but it could be any other cases with different values.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/list/ProcessInstanceListPresenter.java
```diff
@@ -655,15 +665,6 @@ public void openTaskView(final String pid) {
                               pid);
     }
 
-    public void openErrorView(final String pid) {
```
cristianonicolai: seems like the same code, I guess we could just keep it?

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridPresenter.java
```diff
@@ -150,11 +152,16 @@ protected void navigateToPerspective(final String perspectiveId,
...
-
+    
     public boolean isUserAuthorizedForPerspective(final String perspectiveId) {
         final ResourceRef resourceRef = new ResourceRef(perspectiveId,
                                                         ActivityResourceType.PERSPECTIVE);
         return authorizationManager.authorize(resourceRef,
                                               identity);
     }
+    
+    public abstract void openErrorView(final String parameterId);
```
cristianonicolai: You can make these two methods non abstract, that way you only override where needed.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
```diff
@@ -27,32 +27,35 @@
...
+import org.jbpm.workbench.common.client.list.AbstractMultiGridPresenter;
+import org.jbpm.workbench.common.model.GenericErrorSummary;
+import org.jbpm.workbench.common.client.resources.i18n.Constants;
+
 
 @Dependent
-@Templated(value = "ProcessInstanceSummaryErrorCountCell.html", stylesheet = "ProcessInstanceSummaryErrorCountCell.css")
-public class ProcessInstanceSummaryErrorCountCell extends AbstractCell<ProcessInstanceSummary> implements IsElement {
+@Templated(value = "GenericErrorSummaryCountCell.html", stylesheet = "GenericErrorSummaryCountCell.css")
+public class GenericErrorSummaryCountCell<C extends GenericErrorSummary> extends AbstractCell<C> implements IsElement {
```
cristianonicolai: This looks good :)

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/GenericErrorSummaryCountCell.java
```diff
@@ -27,32 +27,35 @@
...
 
     @Inject
     @DataField(TOOLTIP_NAME)
     private Anchor tooltip;
 
-    private ProcessInstanceListPresenter viewPresenter;
-
-    public ProcessInstanceSummaryErrorCountCell init(final ProcessInstanceListPresenter viewPresenter) {
+    private AbstractMultiGridPresenter viewPresenter;
+
```
cristianonicolai: just get rid of some extra spacing, there are quite a few places in the PR like this.

mcivantos-tribalyte: @cristianonicolai @bsremac All your requested changes are already done :)

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridPresenter.java
```diff
@@ -150,11 +152,17 @@ protected void navigateToPerspective(final String perspectiveId,
...
         final ResourceRef resourceRef = new ResourceRef(perspectiveId,
                                                         ActivityResourceType.PERSPECTIVE);
         return authorizationManager.authorize(resourceRef,
                                               identity);
     }
+    
+    public void openErrorView(final String parameterId) {}
+    
+    public Predicate<T> getViewErrorsActionCondition(){
+        return null;
```
bsremac: @mcivantos-tribalyte I would avoid returning ```null``` and make this method abstract instead, but it's your choice.

mcivantos-tribalyte: @bsremac At first I made this method abstract, but Cristiano suggested change it in order to overriding only where needed.

cristianonicolai:
@mcivantos-tribalyte thanks!

PR #963, opened 2017-12-20
[do not merge] JBPM-6659 - Update details pages to take full screen

cristianonicolai:

PR #967, opened 2018-01-10
Consolidate screens id

cristianonicolai:


cristianonicolai:
@nmirasch @bsremac just taking this refactor part out of https://github.com/kiegroup/jbpm-wb/pull/963 as that PR won't be merged.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/PerspectiveIds.java
```diff
@@ -23,4 +23,18 @@
...
+    String PROCESS_DEFINITION_LIST_SCREEN = "Process Definition List";
+    String DASHBOARD_SCREEN = "DashboardScreen";
+    String EXECUTION_ERROR_DETAILS_SCREEN = "Execution Error Details";
+    String EXECUTION_ERROR_LIST_SCREEN = "Execution Error List";
+    String JOB_LIST_SCREEN = "Job List";
+    String JOB_DETAILS_SCREEN = "Job Details";
+    String TASK_ADMIN_LIST_SCREEN = "Task Admin List";
+    String TASK_LIST_SCREEN = "Task List";
+    String TASK_DETAILS_SCREEN = "Task Details Multi";
+
```
bsremac: @cristianonicolai Looks good. Have just one question regarding keeping the names for some identifiers, which look a bit cryptic (e.g. ```"DataSet Process Instance List With Variables"```) . I see also a couple of others which might use some trimming and aligning with the rest.
Is this the place and time to address something like this or are these about to change anyway moving to new perspectives in the future?

nmirasch: @bsremac @cristianonicolai maybe some literal could be more explanatory and maybe is a good moment to unify with the other workbench screens.

cristianonicolai: I agree, now that we combine these ids, it makes easier to visualize the inconsistency. I would be fine for us to update it now. In general, I see the following patterns: append Screen as suffix and remove any space. Will do a quick update and you let me know what you think.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/PerspectiveIds.java
```diff
@@ -23,4 +23,18 @@
     String SEARCH_PARAMETER_TASK_ID = "taskId";
     String SEARCH_PARAMETER_JOB_ID = "jobId";
     String SEARCH_PARAMETER_IS_ERROR_ACK = "isErrorAck";
-}
+
+    String PROCESS_INSTANCE_DETAILS_SCREEN = "ProcessInstanceDetailsScreen";
```
cristianonicolai: @bsremac @nmirasch wdyt?

bsremac: @cristianonicolai Looks good. Clear and concise.

nmirasch: @cristianonicolai good!

PR #969, opened 2018-01-12
[JBPM-6742] Test case roles assignment during case life cycle

bsremac:
@cristianonicolai @nmirasch Last addition to case roles tests in community. Had to do some re-factoring when moving non-trivial logic form the views to the presenter in order to make the logic testable and clearer here and there. While doing so, also fixed a couple of smaller issues:
1.  Owner role not shown among case roles when the owner role is not defined in the case definition. 
2. Pagination shown when case roles are loaded for the first time, even when the total number of listed roles doesn't exceed page size (page size is currently hard-coded to 2)
3.  Changed the message being displayed when the selected filter returns an empty list from ```No Roles Assignments Found``` to ```No Case Roles Found```, as the former message sounds illogical when the filter is set to ```Unassigned```. Furthermore, we are filtering case roles based on their assignments, not assignments per se.


cristianonicolai:
@bsremac looks really nice, the only minor comment is related to some formatting when you create some _CaseRoleAction_ inner classes.

bsremac:
@cristianonicolai Yeah, that formatting is as weird as it gets :)
Actually, I thought that we have some formatting rule around this, so I tried to mimic that when migrating similar constructs from view classes.
Fixed it by the rule of a common sense :)

PR #970, opened 2018-01-13
RHBA-31: Unable to type Due Date when scheduling a Job

nmirasch:
Replace the quickNewJobPopup by the newJob component using errai Ui, including the integration of the DateRangePicker configured to show a single date-time Picker.

Depends on https://github.com/kiegroup/appformer/pull/138


jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobViewImpl.java
```diff
@@ -0,0 +1,463 @@
...
+        inlineNotification.setType(InlineNotification.InlineNotificationType.DANGER);
+    }
+
+    protected EventListener<KeyboardEvent> getNumericInputListener() {
+        return (KeyboardEvent e) -> {
+            int keyCode = e.getKeyCode();
+            if (keyCode <= 0) { //getKeyCode() returns 0 for numbers on Firefox 53
+                keyCode = e.getWhich();
+            }
+            if (!((keyCode >= KeyCodes.KEY_NUM_ZERO && keyCode <= KeyCodes.KEY_NUM_NINE) ||
```
bsremac: Does this mean that we don't allow entering 0 retries?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobViewImpl.java
```diff
@@ -0,0 +1,463 @@
...
+        options.setAutoApply(true);
+        options.setTimePicker(true);
+        options.setTimePickerIncrement(1);
+        options.setSingleDatePicker(true);
+        options.setMinDate(moment());
+
+        return options;
+    }
+
+    boolean validateForm() {
```
bsremac: Can we move this validation to Presenter?
This type of validation was a common generator of negative Selenium tests, which we want to avoid and test via unit tests.

One of the options is to move it under createJob(...) method in presenter, as a whole or each validation divided into it's own method, and accompany it with a couple of more methods in the view like showTypeErrorMessage().
As for _else_ branches, I think they are superfluous as you are calling cleanErrorMessages() at the validation start. This cleanup can be moved to onCreateClick() before presenter.createJob(...) is called.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobViewImpl.java
nmirasch: yeah, we allow keycode == KeyCodes.KEY_ZERO

nmirasch:
@bsremac thanks for all, I added the changes suggested.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobViewImpl.java
nmirasch: @bsremac Here I only add the same control that was used at advanced search. Tested on Chrome it works fine, allowing entering the 0 value. If you find it don't work in any case, we can review that

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenter.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Inject
+    protected NewJobView view;
+
+    @PostConstruct
+    public void init() {
+        view.init(this);
+    }
+
+    @WorkbenchPartView
```
cristianonicolai: Are this method and annotation needed?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/resources/org/jbpm/workbench/es/client/i18n/Constants.properties
```diff
@@ -22,6 +22,8 @@ Message=Message
 Key=Key
 Value=Value
 Due_On=Due On
+Launch_Now=Launch now
```
cristianonicolai: @nmirasch @bsremac _Launch now_ or _Run now_?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenterTest.java
```diff
@@ -0,0 +1,244 @@
...
+import org.mockito.stubbing.Answer;
+import org.uberfire.mocks.CallerMock;
+import org.uberfire.mocks.EventSourceMock;
+import org.uberfire.workbench.events.NotificationEvent;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+@RunWith(GwtMockitoTestRunner.class)
+@WithClassesToStub({Text.class})
```
cristianonicolai: No need to use _WithClassesToStub_

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/resources/org/jbpm/workbench/es/client/i18n/Constants.properties
cristianonicolai: If we change to _Run now_, should we also change _Schedule_ to _Run later_?

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AdvancedSearchFiltersViewImpl.html
```diff
@@ -13,7 +13,7 @@
                     <div data-field="filters-input"></div>
                     <div class="filters-input">
                       <span>
-                        <a data-field="filters-input-help" class="hidden" role="button" data-toggle="popover" data-trigger="hover click"
+                        <a data-field="filters-input-help" class="hidden" role="button" data-toggle="popover" data-trigger="hover"
```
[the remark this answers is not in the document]
nmirasch: Maybe I have to create another jira for this, but the point is if you click on the 'info' icon, it opens the popover  and when the user change the screen without clicking it again, it remains open and impossible to close. Talking with Walter, he found this and he  applied the same solution at his code, remove the click trigger. Make sense open another jira for this small change? What do you think?

nmirasch:
@cristianonicolai  thanks, updated, I changed 'Launch now' and 'Schedule' by' Run now' and 'Run later'. Of course if @bsremac  have a better label for that, we can change it again. :)

bsremac:
@nmirasch @cristianonicolai I don't think we can't go more consistent than _Run now_ and _Run later_, when it comes using the same verb in both actions, so +1.
I would however also rename data fields "launch-now" and "launch-later" to "run-now" and "run-later", and name "launching" to "run" for consistency reasons.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenter.java
```diff
@@ -0,0 +1,182 @@
...
+                                       jobType,
+                                       dueDate,
+                                       jobCtxMap);
+        } else {
+            view.showBasicPane();
+        }
+
+    }
+
+    protected boolean validateForm(String jobName, String jobType, String jobRetries) {
```
bsremac: You broke the arguments list convention here :)

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenterTest.java
```diff
@@ -0,0 +1,241 @@
...
+                                                      anyString(),
+                                                      any(Date.class),
+                                                      any(Map.class));
+
+        presenter.createJob(JOB_NAME,
+                                  new Date(),
+                                  JOB_TYPE,
+                                  JOB_RETRIES,
+                                  new ArrayList<RequestParameterSummary>());
+
```
bsremac: You have a couple of formatting issues of this type in this class.

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenter.java
```diff
@@ -0,0 +1,182 @@
...
+
+        void show();
+
+        void hide();
+
+        void showBasicPane();
+
+        void cleanErrorMessages();
+
+        void showNameErrorMessage();
```
bsremac: Maybe to be clearer and consistent with two types of errors for _Type_ field, we could state in the method name what type of error is for others too, i.e. ```showEmptyNameErrorMessage()``` and below ```showEmptyRetriesErrorMessage()```?

nmirasch:
@bsremac done

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenter.java
```diff
@@ -0,0 +1,185 @@
...
+    @PostConstruct
+    public void init() {
+        view.init(this);
+    }
+
+    public NewJobView getView() {
+        return view;
+    }
+
+    public void show(String serverTemplateId) {
```
bsremac: I can't shake of the feeling that this method should be divided in two - generic setter for _serverTemplateId_ and another method that will call only ```view.show()``` (maybe even named more descriptively e.g. ```openNewJobDialog()```).
The reason for this is that _serverTemplateId_ doesn't play any role in opening the new job dialog per se, but only later as part of calling the backend when the schedule request is sent to kie server.
However, this distinction is well hidden when the method is called in _RequestListPresenter_ and the reader is bound to assume the correlation whereas there is none.
Just a suggestion though :)

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AdvancedSearchFiltersViewImpl.html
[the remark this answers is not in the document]
cristianonicolai: hm, that´s fine. I find useful to click but not to a point that it will stay in the page forever :P

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenter.java
```diff
@@ -0,0 +1,185 @@
...
+
+    @Inject
+    protected NewJobView view;
+
+    @PostConstruct
+    public void init() {
+        view.init(this);
+    }
+
+    public NewJobView getView() {
```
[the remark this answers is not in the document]
cristianonicolai: +1

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenter.java
cristianonicolai: Mixed feelings here, I get your point @bsremac, but to me is also kind of a logical dependency. You're explicitly telling to create/show a job for that server template. Thinking of the domain data that this UI depends on, like if we had to show details of a process instance in a pop up, I would either pass its id or a complete object in here.

bsremac: @cristianonicolai Right, I get it. Considering everything now, I think there are more reasons to pass it then not. Let's leave it as it is.
@nmirasch Maybe just renaming the method then to something which reflects this approach (```setNewJobDialog```?), as we are setting in the method whats needed for creating a new job eventually.

nmirasch:
@bsremac @cristianonicolai  removed the getView method from presenter and renamed the show method to openNewJobDialog. 
@bsremac For me setNewJobDialog doesn't include the  meaning of 'show'.  I prefer your previous suggestion 'openNewJobDialog' for that. What do you think?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenterTest.java
```diff
@@ -0,0 +1,239 @@
...
+                            new ArrayList<RequestParameterSummary>());
+
+        verify(executorServicesMock).scheduleRequest(anyString(),
+                                                     anyString(),
+                                                     any(Date.class),
+                                                     any(HashMap.class));
+    }
+
+    @Test
+    public void jobNameIsPassedAsBusinessKeyTest() {
```
bsremac: What is the difference between this test and ```testScheduleUnSuccessfulJob()```?

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenterTest.java
```diff
@@ -0,0 +1,239 @@
...
+    @Test
+    public void invalidSchedulingError_EmptyNameTest() {
+        presenter.createJob("",
+                            new Date(),
+                            JOB_TYPE,
+                            JOB_RETRIES,
+                            new ArrayList());
+
+        verify(view).cleanErrorMessages();
+        verify(view).showEmptyNameErrorMessage();
```
bsremac: Maybe it makes sense to also add the verification that _ExecutorService_ never called _scheduleRequest_.

nmirasch:
@bsremac absolutely right about the tests, I changed it a bit including that index problems you found,  renaming some validation error test and adding a test for checking the service call parameters (old unsuccessfulScheduling). Could you take another look on that? Thanks!

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenterTest.java
cristianonicolai: +1

jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenterTest.java
```diff
@@ -0,0 +1,226 @@
...
+            public Void answer(InvocationOnMock invocationOnMock) throws Throwable {
+                assertTrue(invocationOnMock.getArguments()[0].equals(serverTemplateId));
+                assertTrue(invocationOnMock.getArguments()[1].equals(JOB_TYPE));
+                assertTrue(invocationOnMock.getArguments()[2].equals(dueOnDate));
+                final HashMap<String, String> ctxValues = (HashMap) invocationOnMock.getArguments()[3];
+                assertTrue(ctxValues.get(RequestDataSetConstants.COLUMN_BUSINESSKEY).equals(JOB_NAME));
+                assertTrue(ctxValues.get(RequestDataSetConstants.COLUMN_RETRIES).equals(String.valueOf(JOB_RETRIES)));
+                assertTrue(ctxValues.get(param_key).equals(String.valueOf(param_value)));
+                return null;
+            }
```
bsremac: The arguments are correct now, but the main problem is still there. Assertions in ```answer(...)``` method don't work.
Try to replace
```assertTrue(invocationOnMock.getArguments()[0].equals(serverTemplateId));``` 
with 
```assertTrue(invocationOnMock.getArguments()[0].equals("Fail test"));``` 
and the test will still pass.
Furthermore, the code in ```answer(...)``` method after the first assert statement is not executed at all.
I don't know the exact reason for this behavior, as it comes from Mockito internals obviously, but my hunch is that it has to do with using _doAnswer()_ for stubbing a non-void method.

File: jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/main/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenter.java [hunk 15]
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        return valid;
+    }
+
+    private boolean isInvalidCommandTypeError(Throwable throwable) {
+        String message = "";
+        if (throwable.getMessage() != null) {
+            message = throwable.getMessage();
+        } else if (throwable.getCause() != null) {
```
bsremac: As I mentioned, here we have a potential _NullPointerException_ gap. The _else if_ expression should be expanded to:
```else if (throwable.getCause() != null && throwable.getCause().getMessage() != null)```

**Comment by nmirasch** [comment]
@bsremac changed the condition to avoid the possible nullpointerexception. The issue about the test is the returning value null and any exception generated by failing exception have the same effect. I changed it a bit to know if the assertions have passed. Now, the test know if any of them fails.

File: jbpm-wb-executor-service/jbpm-wb-executor-service-client/src/test/java/org/jbpm/workbench/es/client/editors/quicknewjob/NewJobPresenterTest.java [hunk 16]
```diff
@@ -0,0 +1,229 @@
...
+                                                      any(Map.class));
+
+        presenter.createJob(JOB_NAME,
+                            dueOnDate,
+                            JOB_TYPE,
+                            JOB_RETRIES,
+                            Arrays.asList(new RequestParameterSummary(param_key,
+                                                                      param_value)));
+        assertEquals("Success",
+                     assertionsResults.toString());
```
bsremac: @nmirasch This approach definitely does the trick, but really looks more like a workaround rather than something we would like to repeat again somewhere else.
Also setting aside the look, the assert error is very scarce and gives no information which assertion failed and for which reason.
I found another, more elegant approach which I think you will like too. The solution relies on _AssertJ_ library, so you have to add to _jbpm-wb-executor-service-client/pom.xml_ the following dependency:
```
    <dependency>
      <groupId>org.assertj</groupId>
      <artifactId>assertj-core</artifactId>
      <scope>test</scope>
    </dependency>
```
Once you did that, you just have to replace a couple of things in your test:
1. ```final StringBuffer assertionsResults = new StringBuffer();``` 
with ```final SoftAssertions softly = new SoftAssertions();```
2. Each assertion statement ```assertTrue(invocationOnMock.getArguments()[1].equals(JOB_TYPE));```
with ```softly.assertThat(invocationOnMock.getArguments()[1]).isEqualTo(JOB_TYPE);```
3. 
```
assertEquals("Success",
             assertionsResults.toString());
```
with ```softly.assertAll();```

Now, you have a full description of a failure and even more better - if there are multiple assertion failures, all of them will be collected and presented :)

nmirasch:
@bsremac added the new magic to the test. Really nice this SoftAssertions, than allow to run the assertions after and at the same time, Interesting to see all the failing assertions messages together. Thanks Bojan!

PR #973, opened 2018-01-17
JBPM-6015: Process instance log makes no sense. Changed created by st…

nmirasch:
…arted


nmirasch:
@cristianonicolai could you take a look? Thanks!

PR #977, opened 2018-01-27
[JBPM-6783] Jobs: Test new job dialog

bsremac:
@cristianonicolai @nmirasch Added a couple of more tests and did a small re-factoring of the existing tests.
Two bugs fixed:
1. On new job dialog re-open, the basic tab/pane is not set by default, although I could see in the code that was the intended behavior.
2.  Bogus additional job parameters entries (added via Advanced pane) i.e. those which contain _Click to Edit_ in place of parameter key/value are sent to Kie Server and as a consequence shown in _Job Details_ panel.


cristianonicolai:
looks good to me, thanks @bsremac

PR #985, opened 2018-02-22
JBPM-6877 - Fix screen id for navigating to task and process instance…

cristianonicolai:
… details screens


bsremac:
On the latest Firefox (58.0.1), the main view's header gets cut off when details view is open, but that has nothing to do with the changes in this PR. On Chrome, it works fine.

screenshot from 2018-02-23 11-43-11
 
@jhrcek FYI

PR #988, opened 2018-02-25
JBPM-6892 - Remove spacing between table results and controls

cristianonicolai:
Depends on https://github.com/kiegroup/appformer/pull/202

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/ExtendedPagedTable.java
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/ExtendedPagedTable.java
@@ -53,7 +53,10 @@ public class ExtendedPagedTable<T extends GenericSummary> extends PagedTable<T> 
         super(DEFAULT_PAGE_SIZE,
               (T item) -> (item == null) ? null : item.getId(),
               gridPreferences,
-              true);
+              true,
+              false,
+              false,
+              false);
 
         selectedItems = new ArrayList<T>();
         dataGrid.addColumnSortHandler(new AsyncHandler(dataGrid));


bsremac:
@cristianonicolai I'm getting the following compilation error when trying to build _jbpm-wb_:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project jbpm-wb-integration-api: Compilation failure
[ERROR] /home/bsremac/Devel/github.com/bsremac/jbpm-wb/jbpm-wb-integration/jbpm-wb-integration-api/src/main/java/org/jbpm/workbench/wi/dd/type/DDResourceTypeDefinition.java:[25,8] org.jbpm.workbench.wi.dd.type.DDResourceTypeDefinition is not abstract and does not override abstract method getCategory() in org.uberfire.workbench.type.ResourceTypeDefinition
```
I have the dependent appformer commit in place, so not sure where is this error coming from.

cristianonicolai:
@bsremac give it another go, I rebased this PR

bsremac:
Spacing looks fine, but there seems to be some problems with that horizontal slider. It "eats up" the space of the last row and it doesn't always appear. Have a look at screenshots below.
Trims the last (and here the only) row:
screenshot from 2018-02-26 22-35-43

The last rows trimmed, but is the slider is not showing up:
screenshot from 2018-02-26 22-36-46

screenshot from 2018-02-26 22-40-16

cristianonicolai:
@bsremac thanks, I had to increase the spacing to fix the horizontal slider issue. Strangely enough, this seems to only happen on Firefox.

See https://github.com/kiegroup/appformer/pull/205

bsremac:
@cristianonicolai I don't see the horizontal slider now at all, when I navigate to perspectives which don't have an empty list:
screenshot from 2018-02-27 12-11-51

However, if there's more than one page, the slider appears on the next page:
screenshot from 2018-02-27 12-13-40

Again, I see this inconsistency only in Firefox.

As an aside, what is the expected behavior with these sliders anyway. When is it expected for them to appear?
Even when I enable all the columns, in Process Instances perspective for example (the first attached screenshot), the column widths are just adjusted accordingly and the slider is not shown. This behavior is the same in Firefox and Chrome.

cristianonicolai:
@bsremac ok, let me take another look, these sliders should not show up at any time, at least I can't find a reason for that, apart from the overall height in the table that somehow affects it.

bsremac:
@cristianonicolai After the latest changes in appformer PR, the horizontal slider deosn't appear anymore in Firefox - that's fine.
The only thing which sort of stands out a bit is that controls drag a bit after changing the number of items per page to the lower number. Have a look at the short recording below. It's the same behavior in Firefox and Chrome.

[Screencast from 27-02-18 16:20:28.zip](https://github.com/kiegroup/jbpm-wb/files/1763626/Screencast.from.27-02-18.16.20.28.zip)

cristianonicolai:
@bsremac yeah, that happens because the table height is recalculated. I tried getting away from that but it simply doesn't work for the GWT table implementation, it needs to know/calculate the table height. Ideally, the selector and paging should just be there after the table uses its space.

bsremac:
@cristianonicolai I get it. Well, I guess we will have to live with certain limitations :)
Thanks for sharing the workings. Now I know that I will need to increase the timeout when locating control elements in Selenium tests for this type of use case.

PR #989, opened 2018-02-26
Clean up auth properties files comments

cristianonicolai:


cristianonicolai:
@mareknovotny the auth issue is due to the Errai merging the content of these files with the ones from WildFly, somehow it happens that _$REALM_NAME_ is added twice causing an error with the login module.

mareknovotny:
thanks! this solves the issue with login. I will merge this change into my PR and close it

PR #990, opened 2018-02-28
JBPM-6894 - User should be able to define case, process, activity lev…

mswiderski:
…el SLA

depends on https://github.com/kiegroup/droolsjbpm-integration/pull/1378

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.workbench.common.client.util;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import elemental2.dom.DomGlobal;
+import elemental2.dom.Element;
+import org.kie.api.runtime.process.ProcessInstance;
+
+import com.google.gwt.cell.client.AbstractCell;
+import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
+
+public class SLAComplianceCell extends AbstractCell<Integer> {
+    
+    private static final String SLA_STYLE = "label";
+    private static final String SLA_NA_PRESENT_STYLE = "label-default";
+    private static final String SLA_PENDING_PRESENT_STYLE = "label-primary";
+    private static final String SLA_MET_PRESENT_STYLE = "label-success";
+    private static final String SLA_ABORTED_PRESENT_STYLE = "label-warning";
+    private static final String SLA_VIOLATED_PRESENT_STYLE = "label-danger";
+    
+    private List<String> descriptions;
+    
+    public SLAComplianceCell(final List<String> descriptions) {
+        this.descriptions = descriptions;
+    }
+
+    @Override
+    public void render(Context context,
+                       Integer value,
+                       SafeHtmlBuilder sb) {
+
+        List<String> tooltipClasses = new ArrayList<>();
+        
+        tooltipClasses.add(SLA_STYLE);
+        String description = descriptions.get(0);
+        switch (value) {
+            case ProcessInstance.SLA_NA:
+                description = descriptions.get(1);
+                tooltipClasses.add(SLA_NA_PRESENT_STYLE);
+                
+                break;
+            case ProcessInstance.SLA_PENDING:
+                description = descriptions.get(2);
+                tooltipClasses.add(SLA_PENDING_PRESENT_STYLE);  
+                break;
+            case ProcessInstance.SLA_MET:
+                description = descriptions.get(3);
+                tooltipClasses.add(SLA_MET_PRESENT_STYLE);
+                break;
+            case ProcessInstance.SLA_ABORTED:
+                description = descriptions.get(4);
+                tooltipClasses.add(SLA_ABORTED_PRESENT_STYLE);
+                break;
+            case ProcessInstance.SLA_VIOLATED:
+                description = descriptions.get(5);
+                tooltipClasses.add(SLA_VIOLATED_PRESENT_STYLE);               
+                break;
+            default:
+                description = descriptions.get(0);
+                tooltipClasses.add(SLA_NA_PRESENT_STYLE);
+                break;
+
+        }        
+
+        Element span = DomGlobal.document.createElement("span");
+        span.textContent = description;
+        tooltipClasses.forEach(c -> span.classList.add(c));
+        Element content = DomGlobal.document.createElement("span");
+        content.appendChild(span);
+        sb.appendHtmlConstant(content.innerHTML);
+    }
+
+}
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
@@ -206,6 +206,8 @@ public class RemoteProcessRuntimeDataServiceImpl extends AbstractKieServerServic
                 processInstance.getCorrelationKey(),
                 processInstance.getParentId(),
                 null,
+                processInstance.getSlaCompliance(),
+                processInstance.getSlaDueDate(),                
                 0
         );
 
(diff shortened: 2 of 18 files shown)


mswiderski:
anyone willing to review?

tsurdilo:
@mswiderski looks good to me +1 !

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final String SPAN_NAME = "sla-compliance";
+    private static final String SLA_STYLE = "label";
+    private static final String SLA_NA_PRESENT_STYLE = "label-default";
+    private static final String SLA_PENDING_PRESENT_STYLE = "label-primary";
+    private static final String SLA_MET_PRESENT_STYLE = "label-success";
+    private static final String SLA_ABORTED_PRESENT_STYLE = "label-warning";
+    private static final String SLA_VIOLATED_PRESENT_STYLE = "label-danger";
+
+    
+    private List<String> descriptions;
```
cristianonicolai: minor adjustment in spacing between elements

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
```diff
@@ -0,0 +1,122 @@
...
+import org.jboss.errai.ui.shared.api.annotations.DataField;
+import org.jboss.errai.ui.shared.api.annotations.Templated;
+import org.kie.api.runtime.process.ProcessInstance;
+
+import com.google.gwt.cell.client.AbstractCell;
+import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
+
+@Dependent
+@Templated(value = "SLAComplianceCell.html")
+public class SLAComplianceCell extends AbstractCell<Integer> implements IsElement {
```
cristianonicolai: @mswiderski we're moving to the new elemental API for new UI screens, I'm making some comments in here so it can be adjusted to the new API.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
```diff
@@ -0,0 +1,122 @@
...
+package org.jbpm.workbench.common.client.util;
+
+import java.util.List;
+
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+
+import org.jboss.errai.common.client.dom.DOMTokenList;
+import org.jboss.errai.common.client.dom.Span;
+import org.jboss.errai.ui.client.local.api.IsElement;
```
cristianonicolai: should be replaced by _org.jboss.errai.common.client.api.elemental2.IsElement_

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final String SLA_PENDING_PRESENT_STYLE = "label-primary";
+    private static final String SLA_MET_PRESENT_STYLE = "label-success";
+    private static final String SLA_ABORTED_PRESENT_STYLE = "label-warning";
+    private static final String SLA_VIOLATED_PRESENT_STYLE = "label-danger";
+
+    
+    private List<String> descriptions;
+    @Inject
+    @DataField(SPAN_NAME)
+    private Span content;
```
cristianonicolai: To adjust this element to the new API you could use:
`@Inject
    @DataField(SPAN_NAME)
@Named("span")
    private HTMLElement content;`

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
```diff
@@ -0,0 +1,122 @@
...
+                       SafeHtmlBuilder sb) {
+
+        DOMTokenList tooltipClasses = content.getClassList();
+        tooltipClasses.add(SLA_STYLE);
+        String description = descriptions.get(0);
+        switch (value) {
+            case ProcessInstance.SLA_NA:
+                description = descriptions.get(1);
+                tooltipClasses.add(SLA_NA_PRESENT_STYLE);
+                tooltipClasses.remove(SLA_PENDING_PRESENT_STYLE);
```
cristianonicolai: Perhaps is not necessary to remove all unrelated css classes as this element isnt reused? If so, it could be done at the start, removing all classes.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/details/ProcessInstanceDetailsPresenter.java
```diff
@@ -167,5 +190,7 @@ public void callback(final List<NodeInstanceSummary> details) {
         HTML getCorrelationKeyText();
 
         HTML getParentProcessInstanceIdText();
+        
+        HTML getSlaComplianceText();
```
cristianonicolai: Returning the html element is a practice we're trying to get away, please update to only set the relevant text, something like _setSlaCompliance(String text)_

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/details/ProcessInstanceDetailsViewImpl.java
```diff
@@ -43,6 +43,10 @@
     @Inject
     @DataField
     public HTML processVersionText;
+    
+    @Inject
+    @DataField
+    public HTML slaComplianceText;
```
cristianonicolai: These text fields need fixing as they don't match the html element from the template (HTML vs Paragraph). If you could try to use _HTMLParagraphElement_ that would get us on the right path.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/resources/i18n/Constants.java
```diff
@@ -131,7 +131,7 @@
 
     String Process_Instance_Details();
 
-    String Process_Definition_Version();
+    String Process_Definition_Version();
```
cristianonicolai: perhaps revert this change.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/details/ProcessInstanceDetailsPresenter.java
```diff
@@ -118,6 +118,29 @@ public void callback(final ProcessInstanceSummary process) {
                     default:
                         break;
                 }
+                view.getStateText().setText(statusStr);
+                
+                String slaComplianceStr = constants.Unknown();
```
cristianonicolai: Would it be work moving this logic into a separated method that could also be easily tested?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
```diff
@@ -206,6 +206,8 @@ protected ProcessInstanceSummary build(ProcessInstance processInstance) {
                 processInstance.getCorrelationKey(),
                 processInstance.getParentId(),
                 null,
+                processInstance.getSlaCompliance(),
```
cristianonicolai: Would be nice to add a test that ensures the mapping between ProcessInstance and ProcessInstanceSummary works as expected.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
```diff
@@ -0,0 +1,113 @@
...
+                tooltipClasses.add(SLA_VIOLATED_PRESENT_STYLE);               
+                break;
+            default:
+                description = descriptions.get(0);
+                tooltipClasses.add(SLA_NA_PRESENT_STYLE);
+                break;
+
+        }        
+        content.textContent = description; 
+        sb.appendHtmlConstant(content.innerHTML);
```
mswiderski: @cristianonicolai @mbarkley how to render complete HTMLElement here? There is no outerHtml thingy in elemental2.... so with this refactored code there are no coloured labels any more which pretty much makes this custom cell useless...

mswiderski:
@cristianonicolai I made the changes you requested but so far I didn't find a way to achieve what was there before. Hints more than welcome.

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
cristianonicolai: @mswiderski seems like outerHTML is actually missing in the API: https://github.com/google/elemental2/issues/37

I think in this case we could simplify things and just wrap with another element, it also seems that using Errai template would be unnecessary and actually share the element across multiple render calls, have a look in my suggestions here: https://gist.github.com/cristianonicolai/59d6411865dd1b1610cc2f66a315b2eb

mswiderski: @cristianonicolai thanks a lot, this works perfectly. all good I believe, do you mind to take a look again?

cristianonicolai:
looks good, thanks @mswiderski !

PR #998, opened 2018-03-16
JBPM-7007 - Use table header height as offset to calculate total height

cristianonicolai:


cristianonicolai:
@bsremac pushing https://github.com/cristianonicolai/jbpm-wb/pull/35 into master

PR #999, opened 2018-03-19
RHBA-578 & JBPM-6886 PrimaryActionMenu updated

nmirasch:


nmirasch:
@cristianonicolai  @bsremac Updated the PrimaryActionMenuBuilder. Could you take a look?
I have included two commits, covering both issues. Do you think that makes sense squash them?
any thought is, as always very welcome. Thanks!

cristianonicolai:
@nmirasch I think squashing both commits is fine, that is what is going to happen anyway when we merge into master. 
My only comment is that you could reuse the isEnabled/setEnabled and addEnabledStateChangeListener methods from the MenuItem API to toggle those buttons. That way you avoid having to keep a reference to the menu builder too.

bsremac:
@nmirasch Looks good.
@nmirasch @cristianonicolai Have one question regarding the text inside the button which opens up the dialog for starting new process instance.
In Process Instances perspective the text inside the button is _New Process Instance_ and in Process Defintions Details Panel the button with the same action has text _New Instance_ .
How about aligning these two? I have no preference regarding the text itself as long as it is the same in both places.

cristianonicolai:
@bsremac +1 for _New Process Instance_

nmirasch:
@cristianonicolai  @bsremac changed the button text and done a bit refactor on PrimaryActionMenuBuilder to work with listeners and avoid to have the actions references for that purpose.
@cristianonicolai I didn't find the way to that setEnabled notify the listeners via API but added there the methods just in case. Could you take a look? 
Thanks!

bsremac:
+1

PR #1004, opened 2018-03-20
JBPM-6865 - Support basic filtering by multiple status

cristianonicolai:


cristianonicolai:
@bsremac @nmirasch please note that is missing the expand/collapse in the checkbox section. That will come in a separated PR.

bsremac:
@cristianonicolai I'm getting these errors when compiling _jbpm-wb_:
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.kie.workbench.forms:kie-wb-common-forms-serialization:jar is missing. @ org.jbpm:jbpm-wb-forms-backend:[unknown-version], /home/bsremac/Devel/github.com/bsremac/jbpm-wb/jbpm-wb-forms/jbpm-wb-forms-backend/pom.xml, line 105, column 17
[ERROR] 'dependencies.dependency.version' for org.kie.workbench.forms:kie-wb-common-forms-serialization:jar is missing. @ org.jbpm:jbpm-wb-showcase:[unknown-version], /home/bsremac/Devel/github.com/bsremac/jbpm-wb/jbpm-wb-showcase/pom.xml, line 1083, column 17
```

cristianonicolai:
@bsremac rebased PR

bsremac:
Maybe we should have the same behavior here as with errors when all states are checked i.e. cancel state filter?

screenshot from 2018-03-21 16-16-36

cristianonicolai:
@bsremac hm, good point... I guess that would be a common behavior for all places where we use this multi select.

cristianonicolai:
@bsremac done.

bsremac:
@cristianonicolai Looks good.
A couple of more comments from my side.
Should we clear all checkbox selections when navigating back to the perspective? Otherwise we can end up with something like this, which looks confusing in my opinion

screenshot from 2018-03-22 11-56-27

Just making sure that this filter hasn't been forgotten, as all others of the same kind have been changed to checkboxes.

screenshot from 2018-03-22 12-06-49

nmirasch:
@cristianonicolai I think the same about the checkboxes cleanning. It happends with the error checkboxes too, that show selected the last filter addition there

cristianonicolai:
@bsremac sure, need to get the state cleared when navigating back to the UI. 
Regarding the Acknowledged filter I left it as is intentionally as I really don't see multiple options in there, although we could discuss the _Errors_ too :)

bsremac:
@cristianonicolai hahahaha....I really wouldn't want to jump on that train again :)
We can leave this one as is and Errors with checkboxes and then wait for the feedback from the field - maybe they would find one of these two illogical - now I'm pointing my eyes towards Errors :)

cristianonicolai:
@bsremac done, fixed the checkbox state.

PR #1006, opened 2018-03-21
RHBA-604 - Decrease padding in select column

cristianonicolai:


bsremac:
It's fixed.

screenshot from 2018-03-23 12-37-35

PR #1007, opened 2018-03-22
JBPM-7001: Missing notification when abort process instance from deta…

nmirasch:
…ils page


nmirasch:
@cristianonicolai @bsremac could you take a look? Thanks!

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/details/ProcessInstanceDetailsPresenter.java
```diff
@@ -241,5 +245,7 @@ public IsWidget getProcessDiagramView() {
         void displayAllTabs();
 
         void displayOnlyLogTab();
+
+        void displayNotification(String text);
```
cristianonicolai: @nmirasch @bsremac  not sure if you agree but I've mixed feelings about this display notification in the view classes, because it is an event, we could consider that it could leave in the presenter class, wdyt?

bsremac: @cristianonicolai @nmirasch Now when mentioned, I agree it makes more sense to handle it inside the presenter

nmirasch:
@cristianonicolai @bsremac I have removed the displayNotification from the view.

PR #1010, opened 2018-03-23
JBPM-7003 - Update page size selector style to match PatternFly dropdown

cristianonicolai:

PR #1012, opened 2018-03-23
JBPM-6868: Review new filter popup

nmirasch:


jbpm-wb-common/jbpm-wb-data-filters-client/src/main/java/org/jbpm/workbench/df/client/filter/FilterEditorPopup.ui.xml
```diff
@@ -22,61 +22,52 @@
...
              xmlns:b="urn:import:org.gwtbootstrap3.client.ui">
 
-    <ui:with field="i18n" type="org.jbpm.workbench.df.client.i18n.FiltersConstants"/>
-    <ui:style>
-        .tab {
-            margin-top: 15px;
-        }
-    </ui:style>
+  <ui:with field="i18n" type="org.jbpm.workbench.df.client.i18n.FiltersConstants"/>
+  <ui:style>
```
cristianonicolai: Looks like this custom style could be removed as well.

nmirasch:
@cristianonicolai @bsremac removed the creation filter popup tabs. Could you take a look? Thanks!

nmirasch:
@cristianonicolai ups, absolutely right, updated!

bsremac:
We have two bugs. I can raise separate JIRAs for these or you can try to fix them as part of this PR.
1. The error messages which are expected to be shown only when trying to save the filter with empty name or without any query created, pop out also during the filter creation, depending on the order in which the fields are being populated. This is the case where the filter description is entered first and then the selection is moved to filter name - at that moment both error messages pop out.

screenshot from 2018-03-25 20-41-01

2. When trying to delete added query using that X button next to it, the NPE is thrown:

screenshot from 2018-03-25 20-33-45

cristianonicolai:
@bsremac I can reproduce the first error, even without these changes but still trying the second one.
Is that related to some column and/or filter type in particular?
If we don't consider these blockers I'm fine to handle it with individual JIRAs.

nmirasch:
@bsremac couldn't reproduce the second one about removing conditions with X. But for the other issue, I removed the field change handlers that were doing that validation. Now the validation is only done when the Ok button is clicked.

bsremac:
@cristianonicolai @nmirasch I've given it another look at the reproducer for issue No.2. 
I realize now it affects only Task Inbox perspective. Try it out there and you will see the issue. Column you select is irrelevant.

nmirasch:
@bsremac yes! let me see what is happening there. Thanks Bojan!

nmirasch:
@bsremac @cristianonicolai  it wasn't related with an specific item list, it was related with the scope of the components that made some instances remained live and observing the delete filter event. To fix that, the popup has been defined application scope as its object container and every time that is opened/closed the filter editor is created/destroyed.

bsremac:
@nmirasch If you want me to try it out one more time, you would have to probably rebase this PR as it doesn't compile anymore due to the new appformer master state.
The compilation error I'm getting is:
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/bsremac/Devel/github.com/bsremac/jbpm-wb/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridView.java:[137,24] cannot find symbol
  symbol:   method createPageSizesListBox(int,int,int)
  location: variable newListGrid of type org.jbpm.workbench.common.client.list.ListTable<T>
[INFO] 1 error
```

nmirasch:
@bsremac its retaled with a parcially merge of the https://github.com/kiegroup/jbpm-wb/pull/1010 (only done on appformer :( )

nmirasch:
@bsremac rebased!

jbpm-wb-common/jbpm-wb-data-filters-client/src/main/java/org/jbpm/workbench/df/client/filter/FilterEditorPopup.java
```diff
@@ -100,18 +83,15 @@
     DataSetClientServices dataSetClientServices;
 
     @Inject
-    public FilterEditorPopup(DataSetClientServices dataSetClientServices,
-                             DataSetFilterEditor filterEditor) {
-        this.filterEditor = filterEditor;
-        this.dataSetClientServices = dataSetClientServices;
-        setBody(uiBinder.createAndBindUi(FilterEditorPopup.this));
+    private ManagedInstance<DataSetFilterEditor> filterEditorsProvider;
```
cristianonicolai: Wondering if we need to use manage instance in here, isn't enough changing the scope to ApplicationScoped?

nmirasch:
@cristianonicolai I think it makes sense. Changed!

PR #1013, opened 2018-03-23
JBPM-7069 - Add BA to task admin query

cristianonicolai:


cristianonicolai:
Depends on https://github.com/kiegroup/jbpm/pull/1166

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/DataSetDefsBootstrap.java
```diff
@@ -119,25 +119,35 @@ protected void registerDataSetDefinitions() {
                 .dataSource(JBPM_DATA_SOURCE)
                 .dbSQL("select " +
                                SQL_SELECT_COMMON_COLS + "," +
+                                    "oe.id, " +
```
bsremac: Not familiar with data sets concept in-depth....will this "oe.id" column be exposed automatically in Tasks table in UI solely by changes made in this PR?
If yes, what defines it's name in the table header?

cristianonicolai: @bsremac adding this column in the data set does not expose in the UI, its only needed so the preprocessor can match the org entities in here: https://github.com/kiegroup/jbpm/pull/1166/files#diff-9f2264d54484a3cfe270ee03fb477ad8R64

PR #1042, opened 2018-04-09
RHBA-71 - Task is still in the reserved state after forwarding it to …

cristianonicolai:
…another user


cristianonicolai:
Cherry pick into master

PR #1049, opened 2018-04-10
[JBPM-7139] Process Instances: Add backend tests for bulk actions

bsremac:
@cristianonicolai @nmirasch Adding tests which are verifying API calls used when performing bulk actions.
With this piece added to existing presenter tests and the latest tests you @cristianonicolai added as part of the fix for bulk actions failing issue, I think we are in good shape to delete Selenium tests which are testing bulk signal/abort, with practically minimal risk.

### DIFF
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/test/java/org/jbpm/workbench/pr/backend/server/RemoteProcessServiceImplTest.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/test/java/org/jbpm/workbench/pr/backend/server/RemoteProcessServiceImplTest.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/test/java/org/jbpm/workbench/pr/backend/server/RemoteProcessServiceImplTest.java
```diff
@@ -0,0 +1,149 @@
+package org.jbpm.workbench.pr.backend.server;
```
cristianonicolai: missing copyright

bsremac: The copyright added.

bsremac:
@cristianonicolai Actually, _processServicesClient_ is the mock on which all verifications in tests are made. Basically, I wanted to test that the proper method from the API is being called and that the argument list is correct in all cases.
Alternatively, I could add a couple of helper methods which would wrap the client and then keep it inside methods. I would then be able to spy _RemoteProcessServiceImpl_ and perform the same type of verification via those helper methods.

cristianonicolai:
@bsremac yep, you can still have the _processServicesClient_ mock variable in the test class because you always return it in test setup method. Your test already works with current version of _RemoteProcessServiceImpl_ :)

cristianonicolai:
jenkins please retest this

PR #1061, opened 2018-04-18
[7.7.x] RHPAM-823 - Misssing showcase logo in navbar

cristianonicolai:


bsremac:
Is it possible to remove the hand-like mouse pointer when hovering over, since the logo itself is read-only?

cristianonicolai:
@bsremac done

PR #1065, opened 2018-04-19
RHPAM-716: Cannot see buttons on Task form

nmirasch:


nmirasch:
@cristianonicolai @bsremac added a padding-bottom on the tab content section to keep the last part visible by the scrolling. It affected to process definition an process instances details screens.
Could you take a look? Thanks!

PR #1067, opened 2018-04-19
JBPM-7173 - On workitem installs create the deployment descriptor if …

tsurdilo:
…one does not exist yet for the project


tsurdilo:
@mswiderski this fixes the issue on installing workitems from repository inside a business process
when the deployment descriptor does not exist yet in the project. I think every time anything wants to interact with the DD we should create it if it does not exist and then process the interaction.

tsurdilo:
@mswiderski do we need a 7.7.x pr for this too?

cristianonicolai:
@tsurdilo +1 for including into 7.7.x branch.

PR #1090, opened 2018-04-26
[7.7.x] RHPAM-861 - Missing start option for available actions

cristianonicolai:


cristianonicolai:
@bsremac what I found is that none of the actions that are in the available status have an id set, its only when they are loaded with a node instance, which happens for in progress and completed.

PR #1092, opened 2018-04-26
[7.7.x] RHPAM-862 - Case Management showcase not listing all available stages

cristianonicolai:


bsremac:
There are two issues when it comes to listing case stages:
1. If the case stage is not named in the case definition, then the case stages' time-line looks like this. Not sure whether we should go with some generic names in that case e.g. Stage 1, Stage 2 etc.?

screenshot from 2018-04-26 17-00-43

2. Stage drop-down list when creating dynamic tasks/sub-processes contains some bogus data where the case stage name should be. This holds for case stages with and without name.

screenshot from 2018-04-26 17-08-30

cristianonicolai:
@bsremac ok so I fixed the stage dropdown which was showing the id instead of the name.
Regarding the missing stage name, there are several places where this will break the UI, for instance in the available actions list, it returns an adhocfragment with no name and there is no information that I could link it to a stage. I've done a partial fix to show the id when the name isn't available but I believe a better fix would be to enforce the name at the designer level.

bsremac:
@cristianonicolai Yes, I agree. The proper fix should be to not let the stage be defined without a name, but then again this is the old designer, so I'm not sure anyone would commit to fixing it. On the other hand, Case Modeler Tool is out of the scope for GA, which has the potential to remedy this in the future.

This makes me think we are on our own to figure this out for the time being. Having stage Id shown in place of stage name, when the stage name is not defined, seems to me like the only viable solution because Stage Id is unique.
The downside of having stage Id shown is that this Id won't mean much to the case worker, as they wouldn't know which stage this Id refers to, but in the combination with the stages execution flow diagram shown in Casey and the process instance diagram shown in the workbench, they could figure it out. It's better than having it empty.

As for the look in the UI, there's one more issue. Now even when the stage has a name, the stage Id is still being shown for the available, non-dynamic ad-hoc action.

screenshot from 2018-04-27 12-28-23

cristianonicolai:
jenkins please retest this

PR #1094, opened 2018-04-26
[7.7.x] RHPAM-865 and RHPAM-866 - Filter Milestones and AdHoc tasks in availa…

cristianonicolai:
…ble actions

### DIFF
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
@@ -20,18 +20,13 @@ import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-
+import java.util.function.BiPredicate;
+import java.util.function.Predicate;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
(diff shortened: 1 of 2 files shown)


jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-backend/src/main/java/org/jbpm/workbench/cm/backend/server/RemoteCaseManagementServiceImpl.java
```diff
@@ -441,4 +463,30 @@ public void triggerAdHocAction(String containerId,
...
+
+    public static class MilestoneNodePredicate implements Predicate<CaseActionSummary> {
+
+        @Override
+        public boolean test(final CaseActionSummary summary) {
+            return NODE_TYPE_MILESTONE.equals(summary.getType());
+        }
+    }
+
+    public static class MilestonePredicate implements BiPredicate<CaseActionSummary, String> {
```
bsremac: I'm failing to the see the value in introducing this ```BiPredicate```.
It's sole use is in ```findMilestone()``` method as far as I can see, which is called only when the expression 
```new MilestoneNodePredicate().negate().test(action)``` has already been evaluated to ```false```, which means the part 
```new MilestoneNodePredicate().test(summary)``` in BiPredicate's test method will always evaluate to ```true```.
In that case the whole filtering in ```findMilestone()``` can be simplified to comparing names only.

cristianonicolai: Completly agree, removed ```MilestoneNodePredicate```

PR #1108, opened 2018-05-02
JBPM-7211 - Handle case actions with similar name

cristianonicolai:


cristianonicolai:
@bsremac updated PR including changes for adding the node type description.

PR #1124, opened 2018-05-11
JBPM-7213: Tasks First Due Date update writes null to task log

nmirasch:
The logs can Introduce characters like ' that are replaced at  SafeHtmlUtils.htmlEscape. To be displayed we need setInnerHTML instead  setTextContent


nmirasch:
@cristianonicolai could you take a look? Thanks!

PR #1126, opened 2018-05-18
JBPM-7105: Missing i18n for task status in basic filters panel

nmirasch:


nmirasch:
@cristianonicolai  @bsremac could you take a look?

PR #1127, opened 2018-05-21
RHPAM-718 - Remove default active filters

cristianonicolai:


cristianonicolai:
@bsremac totally agree and advised not to proceed with this change, but it has been a request from UX and PM.

PR #1142, opened 2018-06-04
JBPM-7313: Case List: Should have refresh option

bxf12315:
@cristianonicolai  I think the PR is not ready.
I just want to make sure my PR on right way.
Actually, I have two way for the issue.
first, it is the PR.
if we choice it, I think we need choice it's style , it is like start case button, or I will try to copy  RefreshMenuBuilder's style.
another, We can copy [code](https://github.com/Cory-jbpm/jbpm-wb/blob/0a4ffbd3045d0bd22d6f03dcbd4d46a0aaf60671/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/list/ProcessInstanceListPresenter.java#L387)  .
But I find , jbpm-case-mgmt don't have many gwt modeles, I can not inherit all modeles.
In the same time, I find case list page different to process instance list page.

I think, the PR miss unit test, can you give me example for it.
Please review it.
Thanks

### DIFF
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewImpl.java
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewImpl.java
@@ -166,6 +166,11 @@ public class CaseInstanceListSearchViewImpl extends AbstractView<CaseInstanceLis
                      true);
     }
 
+    @EventHandler("refresh-case-list")
+    public void onRefreshCaseClick(final @ForEvent("click") MouseEvent event) {
+        presenter.refreshData();
+    }
+
     private void onSortChange(final HTMLElement toHide,
                               final HTMLElement toShow,
                               final Boolean sortByAsc) {
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListViewImpl.html
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListViewImpl.html
@@ -28,9 +28,14 @@
                             <span class="fa fa-sort-numeric-desc"></span>
                         </button>
                     </div>
-                    <div class="form-group">
-                        <button class="btn btn-primary" type="button" data-field="start-case" data-i18n-key="StartCase"></button>
+
+                    <div class="form-group pull-right">
+                            <button class="btn btn-link" type="button" data-field="refresh-case-list" title="refresh-case-list-title">
+                                <i class="fa fa-refresh"></i>
+                            </button>
+                            <button class="btn btn-primary" type="button" data-field="start-case" data-i18n-key="StartCase"></button>
                     </div>
+
                 </form>
             </div>
         </div>
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.workbench.cm.client.list;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CaseInstanceListSearchViewTest {
+
+    @InjectMocks
+    CaseInstanceListSearchViewImpl actions;
+
+    @Mock
+    CaseInstanceListPresenter presenter;
+
+    @Test
+    public void testRefreshCaseList(){
+        actions.onRefreshCaseClick(null);
+
+        verify(presenter).refreshData();
+    }
+
+    @Test
+    public void testCreateCaseClick(){
+        actions.onCreateCaseClick(null);
+
+        verify(presenter).createCaseInstance();
+    }
+
+}
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/client/resources/i18n/Constants.properties
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/resources/org/jbpm/workbench/cm/client/resources/i18n/Constants.properties
@@ -112,3 +112,4 @@ More=...more
 SubProcess=Sub Process
 HumanTask=Human Task
 Milestone=Milestone
+refresh-case-list-title=Refresh


cristianonicolai:
ok to test

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewImpl.java
```diff
@@ -166,6 +166,11 @@ public void onSortStartedDesc(final @ForEvent("click") MouseEvent event) {
                      true);
     }
 
+    @EventHandler("refresh-case-list")
+    public void OnRefreshCaseClick(final @ForEvent("click") MouseEvent event){
```
cristianonicolai: lower case _onRefresh_...

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListViewImpl.html
```diff
@@ -31,6 +31,10 @@ <h1 data-i18n-key="CaseList"></h1>
                     <div class="form-group">
                         <button class="btn btn-primary" type="button" data-field="start-case" data-i18n-key="StartCase"></button>
                     </div>
+
+                    <div class="form-group">
+                        <button class="btn btn-primary" type="button" data-field="refresh-case-list" data-i18n-key="RefreshCaseList"></button>
```
cristianonicolai: class: `btn btn-link` with an icon: `<i class="fa fa-refresh"></i>`

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListViewImpl.html
```diff
@@ -31,6 +31,10 @@ <h1 data-i18n-key="CaseList"></h1>
                     <div class="form-group">
                         <button class="btn btn-primary" type="button" data-field="start-case" data-i18n-key="StartCase"></button>
                     </div>
+
+                    <div class="form-group">
```
cristianonicolai: probably add `pull-right` to get it right aligned.

bxf12315:
@cristianonicolai 
I updated the PR. Please review it.
If it is ok, I will rebase two commits.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListViewImpl.html
```diff
@@ -31,6 +31,12 @@ <h1 data-i18n-key="CaseList"></h1>
                     <div class="form-group">
                         <button class="btn btn-primary" type="button" data-field="start-case" data-i18n-key="StartCase"></button>
                     </div>
+
+                    <div class="form-group pull-right">
+                        <i class="fa fa-refresh"></i>
+                        <button class="btn btn-link" type="button" data-field="refresh-case-list">
```
cristianonicolai: In here, just a minor change, the icon should be inside the button, so no need for the span. We only want to show the refresh icon. Refresh should show as a title for this button when you hover it.
For that to work you need to add `title="..."` to the button and add the following key in the properties bundle: `refresh-case-list-title=Refresh`.

bxf12315:
@cristianonicolai updated. 
BTW, Do i rebase all commit of the PR?

cristianonicolai:
@bxf12315 yep, would be good if you rebase it into a single commit.

bxf12315:
@cristianonicolai I have rebased it.
Please review it.

cristianonicolai:
jenkins please retest this

bsremac:
I would place the refresh button next to _Start Case_ button, as we have done in all Manage perspectives, for consistency reasons. E.g.

screenshot from 2018-06-06 15-39-10

@cristianonicolai @nmirasch Not sure how do you feel about this idea.

cristianonicolai:
@bsremac I'm ok with that as long as we keep it all right aligned then? I mean push both refresh and start case button to right.

bsremac:
@cristianonicolai I guess, that makes sense as well.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListPresenterTest.java
```diff
@@ -138,4 +151,15 @@ public void testSelectCaseInstance() {
...
+
+    @Test
+    public void testRefreshCaskList(){
+        actions.onRefreshCaseClick(event);
+
+        verify(caseManagementService).getCaseInstances(view.getCaseInstanceSearchRequest());
+        final ArgumentCaptor<List> captor = ArgumentCaptor.forClass(List.class);
+        verify(view).setCaseInstanceList(captor.capture());
+        assertEquals(caseInstanceSummaryList.size(),
+                     captor.getValue().size());
```
bsremac: I don't see any point in re-testing the same thing which is already tested by ```testRefreshData()```.
I would keep this test as simple as possible and test only that a proper presenter method is called when the refresh button is clicked i.e.
```
@Test
    public void testRefreshCaskList(){ // Mind the typo you made in naming the method
        actions.onRefreshCaseClick(event);

        verify(presenter).refreshData(); // You would have to add @Spy annotation to presenter object
    }
```

bxf12315: ok you are right

bxf12315:
@cristianonicolai  updated it.
Please review it again

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceListViewImpl.html
```diff
@@ -28,9 +28,14 @@ <h1 data-i18n-key="CaseList"></h1>
                             <span class="fa fa-sort-numeric-desc"></span>
                         </button>
                     </div>
-                    <div class="form-group">
-                        <button class="btn btn-primary" type="button" data-field="start-case" data-i18n-key="StartCase"></button>
+
+                    <div class="form-group pull-right">
```
cristianonicolai: styling looks good.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListPresenterTest.java
```diff
@@ -73,6 +82,10 @@ public void init() {
         when(caseManagementService.getCaseInstances(any(CaseInstanceSearchRequest.class))).thenReturn(caseInstanceSummaryList);
         presenter.setCaseService(caseService);
         when(view.getCaseInstanceSearchRequest()).thenReturn(new CaseInstanceSearchRequest());
+        doAnswer((i) ->{
+            presenter.refreshData();
```
cristianonicolai: I gave a second thought on this test class and I think in here you're actually mocking the bit that needs to be tested.
It would be better if you create a new test class for `CaseInstanceListSearchViewImpl` and then mock the presenter to verify it was actually called.

bxf12315: ok, So Do I  need create test for all method (publlic ) of  CaseInstanceListSearchViewImpl?

bxf12315:
@cristianonicolai  I create new unit test class, it include two test method .
Please review it.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/AbstractCaseInstanceListTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Spy;
+import org.uberfire.client.mvp.PlaceManager;
+import org.uberfire.mocks.CallerMock;
+
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.when;
+
+public class AbstractCaseInstanceListTest {
```
bsremac: Introducing this type of class seems redundant when already having this class available ```org.jbpm.workbench.cm.client.util.AbstractCaseInstancePresenterTest```. Most of what you need is already there, you can just add the mocks and parts in init() method which you need.
Re-using this class instead will also make you rethink the approach about injecting mocks in a superclass. I strongly recommend moving it to the designated test class, even if it means duplicating injection across test classes. The price is not too big and in return we will have easier to understand test classes.

cristianonicolai: Another point is that it is mixing view and presenter test in here, I would also recommend leaving the presenter test as is and just add the necessary code for the view test class.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CaseInstanceListSearchViewTest extends AbstractCaseInstanceListTest{
+    @Mock
```
cristianonicolai: This is the actual test class, you should use `@InjectMocks` in here.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
```diff
@@ -0,0 +1,77 @@
...
+    MouseEvent event;
+
+    @Mock
+    NewCaseInstancePresenter newCaseInstancePresenter;
+
+    @Before
+    public void init() {
+        super.init();
+        doAnswer((i) ->{
+            newCaseInstancePresenter.show();
```
cristianonicolai: Not sure why you need these, the presenter should be a plain mock that you simply what to check if a method was called. The logic inside the presenter should be tested in the presenter test class.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/AbstractCaseInstanceListTest.java
bxf12315: Ok, I do not know the abstract test class. AbstractCaseInstancePresenterTest have many sub class(IMO, it is base class of all test of caseinstancePresenter), I think we should not add more CaseInstancePresenterList's logic in here. I would recommend let AbstractCaseInstanceListTest
extends it, move common logic(like: CaseManagementService caseManagementService;) into it.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
bxf12315: if call presenter.createCaseInstance, it will call newCaseInstancePresenter.show().
So I need add it.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/AbstractCaseInstanceListTest.java
bxf12315: @cristianonicolai I find we can not extends AbstractCaseInstancePresenterTest, it's presenter is AbstractCaseInstancePresenter total different to AbstractPresenter.

cristianonicolai: @bxf12315 maybe try to keep it simple first, just add a new test class...

bxf12315:
Please review it again.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CaseInstanceListSearchViewTest extends AbstractCaseInstanceListTest{
```
cristianonicolai: @bxf12315 the abstract class is not needed for the test you're executing, it is just mixing tests between presenter and view from my perspective. If you remove the extends you test class works anyway.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @InjectMocks
+    CaseInstanceListSearchViewImpl actions;
+
+    @Mock
+    MouseEvent event;
```
cristianonicolai: You can pass null in the event for the view, no need to create a mock for that.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Mock
+    CaseInstanceListPresenter presenter ;
```
cristianonicolai: formatting...

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Before
+    public void init() {
```
cristianonicolai: Please remove empty method.

bxf12315:
@cristianonicolai  I update it. Please review it.

jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/test/java/org/jbpm/workbench/cm/client/list/CaseInstanceListSearchViewTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CaseInstanceListSearchViewTest {
+
+    @InjectMocks
+    CaseInstanceListSearchViewImpl actions;
```
bsremac: Looks good. 
Would just recommend putting the actual object in which all mocks have been injected down to the end of the mock-chain.
It helps readability as it kid of makes sense that first you list what is being mocked and then where all those mocks are being injected. This is the practice used across all tests classes we have.

bxf12315:
@cristianonicolai @bsremac done.
Thanks

PR #1145, opened 2018-06-05
RHPAM-1127: Missing i18n for task status in basic filters panel

nmirasch:


nmirasch:
@cristianonicolai @bsremac cherrypicked from master to 7.7.x. Could you review? Thanks!

PR #1154, opened 2018-06-12
JBPM-7314 Case actions and case status should be aligned

bxf12315:
@cristianonicolai  Please review it.

### DIFF
--- a/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceViewImpl.html
+++ b/jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceViewImpl.html
@@ -1,6 +1,6 @@
 <div class="list-group-item" data-field="list-item">
     <div class="list-view-pf-actions">
-        <button class="btn btn-default hidden" data-field="complete" data-i18n-key="Complete"></button>
+        <button class="btn btn-default hidden" data-field="cancel" data-i18n-key="Cancel"></button>
         <div class="dropdown pull-right dropdown-kebab-pf hidden" data-field="kebab">
             <button class="btn btn-link dropdown-toggle" type="button" id="dropdownKebabRight" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
                 <span class="fa fa-ellipsis-v"></span>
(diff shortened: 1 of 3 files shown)


jbpm-wb-case-mgmt/jbpm-wb-case-mgmt-client/src/main/java/org/jbpm/workbench/cm/client/list/CaseInstanceViewImpl.html
```diff
@@ -1,6 +1,6 @@
 <div class="list-group-item" data-field="list-item">
     <div class="list-view-pf-actions">
-        <button class="btn btn-default hidden" data-field="complete" data-i18n-key="Complete"></button>
+        <button class="btn btn-primary hidden" data-field="cancel" data-i18n-key="Cancel"></button>
```
cristianonicolai: @bxf12315 I don't think we want to change the button style to primary in here, start case is the main action in the case list screen.

bxf12315: @cristianonicolai  If I change to 'cancel', the button is not blue button.
 I want to the blue button ,so I change the style.

cristianonicolai: @bxf12315 a "blue"button is often used as the primary action in a page, to give the user its main focus. According to the designs we had for this screen, the primary button is located at the top of the list with the "start case" action, if you change this cancel button to be a primary ( blue ) button we would break that user experience.

bxf12315: ok go it

bxf12315:
@cristianonicolai 
Please review it.

bsremac:
The actions offered for a case and the list of offered filters by status are now aligned - that's fine, but we still have a problem with filtering the cases based on the action taken.
Namely, the cases we close are also listed under _Canceled_ cases, instead of being listed under _Closed_ cases.

cristianonicolai:
@bsremac is that when connecting to a remote kie server or only when using the mock data?

cristianonicolai:
@bsremac ok, I think we are calling the wrong service to close the case instance. Could you open a specific JIRA for that? Thanks

bsremac:
@cristianonicolai A separate issue opened [JBPM-7393](https://issues.jboss.org/browse/JBPM-7393).
I guess, when it comes to the issue being addressed under the current JIRA, the work is done.

PR #1159, opened 2018-06-26
Add profile to enable skipping of gwt compilation of showcases

jhrcek:
Hello @cristianonicolai  and @nmirasch ,

I'm introducing a new profile that will enable to skip gwt compilation of showcases in `downstream-pullrequests` jobs. In normal build, nothing is skipped. Only when you add `-Pno-showcase` to mvn command line, the showcases won't be gwt-compiled. Please let me know if you're ok with this change / if you can think of better way to achieve the skipping.


jhrcek:
Closing in favor of https://github.com/kiegroup/jbpm-wb/pull/1160

PR #1166, opened 2018-07-04
[7.7.x] RHPAM-1229: Updating task details alters Due On time

nmirasch:


nmirasch:
@cristianonicolai @bsremac I replaced the date and time picker by the DataRangePicker used in other console places. If you think this is a good idea, my proposal is to  remove that UTC components from common and review the QuickNewTaskPopup that I think is the one that remain using them.
Could you take a look please and let me know what do you think? Thanks!

bsremac:
@nmirasch Well, it looks good. By switching to this type of picker, it's more user friendly and as you mentioned it's aligned with other date/time pickers across the jBPM runtime views.
The only downside at the moment is that I have to do some Selenium tests re-factoring and I don't think I will make it on time before 7.0.2 CR1 gets out taking into account the upcoming public holidays.
If we want this fix in for 7.0.2, I'm fine, but some tests will probably be failing due to this change until I fix them.

cristianonicolai:
@bsremac sure, so we should consider this as not done for 7.0.2 :) 
@nmirasch your solution is great to me, we needed the calendar picker changed. I'm also in favour of removing the UTC components from common, I just wouldn't worry about the QuickNewTaskPopup too much as it is not in use anyway, I would be ok to simply remove it as well.
Considering the work that needs to be done to update tests, I guess we can work on the fix for master and cherry-pick once the work for 7.0.3 is available.

PR #1177, opened 2018-07-19
JBPM-7383: Unable to access completed processes with undeployed conta…

nmirasch:
…iner
Allow access to process instances details with stopped container.

### DIFF
--- a/jbpm-wb-kie-server/jbpm-wb-kie-server-backend/src/main/resources/default-query-definitions.json
+++ b/jbpm-wb-kie-server/jbpm-wb-kie-server-backend/src/main/resources/default-query-definitions.json
@@ -3,7 +3,7 @@
     "query-name": "jbpmProcessInstances",
     "query-source": "${org.kie.server.persistence.ds}",
     "query-expression": "select log.processInstanceId, log.processId, log.start_date, log.end_date, log.status, log.parentProcessInstanceId, log.outcome, log.duration, log.user_identity, log.processVersion, log.processName, log.correlationKey, log.externalId, log.processInstanceDescription, log.sla_due_date, log.slaCompliance, COALESCE(info.lastModificationDate, log.end_date) as lastModificationDate, (select COUNT(errInfo.id) from ExecutionErrorInfo errInfo where errInfo.process_inst_id=log.processInstanceId and errInfo.error_ack=0) as errorCount from ProcessInstanceLog log left join ProcessInstanceInfo info on info.InstanceId=log.processInstanceId",
-    "query-target": "FILTERED_PROCESS"
+    "query-target": "CUSTOM"
   },
   {
     "query-name": "jbpmProcessInstancesWithVariables",
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -53,11 +53,10 @@ public class RemoteProcessRuntimeDataServiceImpl extends AbstractKieServerServic
             return null;
         }
 
-        ProcessServicesClient queryServicesClient = getClient(serverTemplateId,
-                                                              ProcessServicesClient.class);
+        QueryServicesClient queryServicesClient = getClient(serverTemplateId,
+                                                            QueryServicesClient.class);
 
-        ProcessInstance processInstance = queryServicesClient.getProcessInstance(processInstanceKey.getDeploymentId(),
-                                                                                 processInstanceKey.getProcessInstanceId());
+        ProcessInstance processInstance = queryServicesClient.findProcessInstanceById(processInstanceKey.getProcessInstanceId());
 
         return build(processInstance);
     }
@@ -71,13 +70,12 @@ public class RemoteProcessRuntimeDataServiceImpl extends AbstractKieServerServic
         }
 
         List<NodeInstanceSummary> instances = new ArrayList<NodeInstanceSummary>();
-        ProcessServicesClient processServicesClient = getClient(serverTemplateId,
-                                                                ProcessServicesClient.class);
+        QueryServicesClient queryServicesClient = getClient(serverTemplateId,
+                                                            QueryServicesClient.class);
 
-        List<NodeInstance> nodeInstances = processServicesClient.findActiveNodeInstances(deploymentId,
-                                                                                         processInstanceId,
-                                                                                         0,
-                                                                                         Integer.MAX_VALUE);
+        List<NodeInstance> nodeInstances = queryServicesClient.findActiveNodeInstances(processInstanceId,
+                                                                                       0,
+                                                                                       Integer.MAX_VALUE);
 
         for (NodeInstance instance : nodeInstances) {
             NodeInstanceSummary summary = new NodeInstanceSummary(instance.getId(),
@@ -103,11 +101,9 @@ public class RemoteProcessRuntimeDataServiceImpl extends AbstractKieServerServic
             return emptyList();
         }
 
-        ProcessServicesClient processServicesClient = getClient(serverTemplateId,
-                                                                ProcessServicesClient.class);
-
-        List<NodeInstance> processInstanceHistory = processServicesClient.findNodeInstances(deploymentId,
-                                                                                            processInstanceId,
+        QueryServicesClient queryServicesClient = getClient(serverTemplateId,
+                                                            QueryServicesClient.class);
+        List<NodeInstance> processInstanceHistory = queryServicesClient.findNodeInstances(processInstanceId,
                                                                                             0,
                                                                                             Integer.MAX_VALUE);
 
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessVariablesServiceImpl.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessVariablesServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
  *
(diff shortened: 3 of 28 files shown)


nmirasch:
Removed FILTERED_PROCESS to retrieve all process instances even the container is nor started.
- Replaced use of ProcessServicesClient.getProcessInstance by QueryServicesClient.indProcessInstanceById
- Replaced use of ProcessServicesClient.ProcessServicesClient.findVariablesCurrentState by QueryServicesClient.findVariablesCurrentState(processInstanceId)
- Show unavaillable diagram message when container stopped.
- Unable to find a replacement of ProcessServicesClient.getProcessVariableDefinitions(deploymentId,                                                                                   processId). I modify the code to show in that case only the variables that have value, but it's unable to know  that process definition variables not filled.
@cristianonicolai  @mswiderski  @bsremac could you take a look? Thanks!

mswiderski:
excellent job @nmirasch it does look very good to me!

bsremac:
@nmirasch There's a problem when a started container, which still has active process instances, is attempted to be stopped via Execution Servers perspective. 
In that case, the stop request is correctly denied, but in the process instance details view the data being displayed changes and shows only data available for a stopped container - which is incorrect, since the container is still deployed and running.
The full reproducer screencast attached
[Screencast from 20-07-18 13:40:57.zip](https://github.com/kiegroup/jbpm-wb/files/2213727/Screencast.from.20-07-18.13.40.57.zip)

nmirasch:
@bsremac not sure if the request is correctly denied. As far as I know the container is stopped even an error notification is shown. The status of the container is 'stopped' and that's the check that we do to know if we can show the diagram. Maybe @mswiderski could answer better, if we have to  add extra checks there or what we have to do in that cases. Other thing I recently noticed is, in that container status (abnormally stopped), the user can start new instances from that container even the status is stopped.

bsremac:
@nmirasch You can't stop the container if there are running process instances in it, and the request for stopping the container will be denied in that case - this is the expected behavior.
If you access the container details on Kie server after the request has been denied using the URL: ```http://127.0.0.1:8080/kie-server/services/rest/server/containers/evaluation_1.0.0-SNAPSHOT```
you will see that the container status hasn't been changed
```<kie-container container-alias="evaluation" container-id="evaluation_1.0.0-SNAPSHOT" status="STARTED">```
This is why you can still start new process instances and interact with the active ones. There's nothing abnormal there.
So, the question is where are we getting the status "stopped" for a container which is obviously still running?

**Comment by cristianonicolai** [comment]
@nmirasch overall looks good to me, just a question regarding security.
Maybe as a note for future improvements, we could move most of these container handling into the service classes, which would make all of these transparent for the presenter.

@nmirasch @bsremac @mswiderski  regarding the container status, it seems like an issue with the controller, my guess is that it could be an issue in here: https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-controller/kie-server-controller-impl/src/main/java/org/kie/server/controller/impl/service/SpecManagementServiceImpl.java#L416
The status in the server template could be inconsistent with the actual status from the kie server if the later operation fails.

File: jbpm-wb-kie-server/jbpm-wb-kie-server-backend/src/main/resources/default-query-definitions.json [hunk 1]
```diff
@@ -3,7 +3,7 @@
     "query-name": "jbpmProcessInstances",
     "query-source": "${org.kie.server.persistence.ds}",
     "query-expression": "select log.processInstanceId, log.processId, log.start_date, log.end_date, log.status, log.parentProcessInstanceId, log.outcome, log.duration, log.user_identity, log.processVersion, log.processName, log.correlationKey, log.externalId, log.processInstanceDescription, log.sla_due_date, log.slaCompliance, COALESCE(info.lastModificationDate, log.end_date) as lastModificationDate, (select COUNT(errInfo.id) from ExecutionErrorInfo errInfo where errInfo.process_inst_id=log.processInstanceId and errInfo.error_ack=0) as errorCount from ProcessInstanceLog log left join ProcessInstanceInfo info on info.InstanceId=log.processInstanceId",
-    "query-target": "FILTERED_PROCESS"
```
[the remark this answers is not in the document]
mswiderski: that is correct, though the security restriction we currently have are not so commonly used and thus I think having option to show processes from removed containers sounds a better fit ... at least for now. wdyt?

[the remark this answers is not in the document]
cristianonicolai: @mswiderski I agreed,  and if I believe any remote operation on the runtime would still be protected by the user/container relationship, correct?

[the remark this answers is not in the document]
mswiderski: that's correct

nmirasch:
@cristianonicolai @bsremac updated PR with a service to access to the container status at backend. This code relays on the stored status of the container given by the specManagementService.
Could you take a look? Thanks!

cristianonicolai:
@bsremac @nmirasch I think the issue with the container status needs to be handled on its own jira and perhaps not block the progress on this one, wdyt? 

@nmirasch  as for the implementation in here, do we still need the checks for _isContainerAvailable_ in the presenter logic? I thought we would load most of the data independently from the container anyway and in some tabs ( diagram ) we would show the information that is not possible to retrieve the data without the running container. In general, we would always try to display the details without checks, not sure if I'm missing something.

bsremac:
@cristianonicolai I guess we can treat the issue of reading the false container status in case the container failed to be stopped via UI and then not displaying available data based on that status as a negative test with its own scenario.
This PR cover well the happy path, so looking at it from that perspective we could merge this one, once we are happy.

nmirasch:
@cristianonicolai @bsremac absolutely agree with the idea of taking the check to the backend. I think the change is better than rely on events firing generation. The only issue  I could think is related with the error  show to the user specifying the cause.
There is a change on the taskdashboards that previously didn't show the details of completed and exited task, now that we have a check on backend to prevent that situation, I removed that restriction. If you think that is a too strong change that could affect any of selenium test, I can let that code as it was. Just let me know what you think is the best option

The truth is this change cleans  and removes a lot of code introduced in previous version of the PR, and that is always good :) 

Could you take a look please? Thanks!

bsremac:
@jhrcek FYI
This PR is introducing a change in Task Dashboards when it comes to data availability - the restrictions over available data are now more relaxed.
Adding the relevant snippet from the comment above:
"_There is a change on the taskdashboards that previously didn't show the details of completed and exited task, now that we have a check on backend to prevent that situation, I removed that restriction. If you think that is a too strong change that could affect any of selenium test, I can let that code as it was._"

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessVariablesServiceImpl.java
```diff
@@ -80,16 +91,19 @@
         }
 
         Map<String, String> properties = new HashMap<String, String>();
+        if (isContainerAvailable(serverTemplateId,
```
cristianonicolai: @nmirasc I think you can still use the query service to get the variables definition.
Something like: 
`queryServicesClient.findProcessesById(processId).get(0).getProcessVariables()`
That way you could skip the container check?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessRuntimeDataServiceImpl.java
```diff
@@ -53,11 +53,10 @@ public ProcessInstanceSummary getProcessInstance(String serverTemplateId,
             return null;
         }
 
-        ProcessServicesClient queryServicesClient = getClient(serverTemplateId,
-                                                              ProcessServicesClient.class);
+        QueryServicesClient queryServicesClient = getClient(serverTemplateId,
```
cristianonicolai: @nmirasch there are more call in this class that use the _ProcessServicesClient_ service that I'm wondering if should also be replaced by the _QueryServicesClient_. Ex: _getProcess_, _getProcessInstanceLogs_, _getProcessInstanceActiveNodes_, _getProcessUserTasks_.

cristianonicolai:
@nmirasch do we also need to make changes to the _RemoteTaskServiceImpl_?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessVariablesServiceImpl.java
nmirasch: @cristianonicolai I have one doubt about this replacement. It returns the first process with that id but it doesn't ensure that is the one related with the process instance. It could be one with the same processId from other container.

nmirasch:
@cristianonicolai  the only thing  I did there is to manage the thrown exception when the getTask fails. Maybe, it makes sense to add there the available container check or Did you mean another kind of change?

cristianonicolai:
@nmirasch I was wondering if we had to use the same _QueryServicesClient_ although it has only a _findHumanTasksWithFilters_ method. That could be extended to retrieve a single task instance and another method to retrieve its comments.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-backend/src/main/java/org/jbpm/workbench/pr/backend/server/RemoteProcessVariablesServiceImpl.java
cristianonicolai: @nmirasch there is another method, _findProcessByContainerIdProcessId_ that might be the one to use then.

nmirasch:
@cristianonicolai I repaced all the ProcessServicesClient by QueryServicesClient I found possible. the ones that needs the containerId are the one that are managing process definition info that need the containerId, as far as I know: getUserTaskDefinitions, getProcessDefinition and getProcessVariableDefinitions. I din't find there a way to retrieve that by query services or even the info that they are trying to retrieve.
About the RemoteTaskServiceImpl I changed the call of getTask to use this queryService. In the next pr regarding display comments of completed task a would add the service to do that and also can improve this call calling to specific service and not findHumanTasksWithFilters like it is right now.

Could you take.. aooother look?  Thanks!!

nmirasch:
@mswiderski @cristianonicolai  if we want get access to completed task even the container is stopped, maybe is the moment to remove that https://github.com/kiegroup/jbpm/blob/4de984bef2bcaeb59615a4a7f395f69186dc5934/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/QueryServiceImpl.java#L184 call as we did with FILTERED_PROCESS
I don't think makes much sense prepare the uI to display task data from stopped containers and then apply this preprocessor. 
If you both agree I can generate the PR removing that preprocessos with this same jira. What do you think?

cristianonicolai:
@nmirasch I think it makes sense.

PR #1179, opened 2018-07-23
RHPAM-1043 - Use Integer type to match column type in db for error li…

cristianonicolai:
…st saved filters


bsremac:
@cristianonicolai It seems we want to try to squeeze this one into CR2, so my initial plan is not important anymore.

PR #1180, opened 2018-07-23
JBPM-7472 - Use Integer type to match column type in db for error lis…

cristianonicolai:
…t saved filters


bsremac:
@cristianonicolai I suggest merging this only on master branch for now and then wait for the next in line nightly build with this commit included. Then the whole test pipeline will be executed against it and we can be completely sure there no more loose ends left this time around before merging it to 7.7.x branch as well.

PR #1186, opened 2018-08-14
JBPM-7592 - Add resource type support for .cmmn files

cristianonicolai:
- .cmmn files are listed as Process type in the asset list
- Added upload file support to showcase


mswiderski:

bsremac:
@mswiderski Thanks for commenting on the current build issue when trying out _cmmn-itorders_ project. It puts the effort in this PR in the right perspective.
@cristianonicolai @mswiderski I see that opening forms related to the original BPMN2-based _itorders.orderhardware_ process throws an error when the process is replaced with CMMN-based one, which is expected, I suppose. Do we want to leave this as it is for now or maybe remove all the forms from _cmmn-itorders_ which rely on the original process type?

screenshot from 2018-08-14 14-22-43

mswiderski:
@bsremac I would opt at removing the forms from that project

cristianonicolai:
@bsremac based on @mswiderski last comment, are you ok for this to be merged as is?

PR #1189, opened 2018-08-23
JBPM-7427 - Display task comments when task is completed

cristianonicolai:
- Added standard alert message to show container error for task and process diagram views
- Set default sorting to be the by date desc, with latest comments first
- Set default column widths
- Set permissions around deleting comments - If not admin, only user that made the comment can remove it.

### DIFF
--- a/jbpm-wb-forms/jbpm-wb-forms-client/src/main/java/org/jbpm/workbench/forms/client/display/task/AbstractHumanTaskFormDisplayer.java
+++ b/jbpm-wb-forms/jbpm-wb-forms-client/src/main/java/org/jbpm/workbench/forms/client/display/task/AbstractHumanTaskFormDisplayer.java
@@ -42,7 +42,6 @@ import org.jbpm.workbench.forms.display.FormDisplayerConfig;
 import org.jbpm.workbench.forms.display.FormRenderingSettings;
 import org.jbpm.workbench.ht.model.TaskKey;
 import org.jbpm.workbench.ht.model.TaskSummary;
-import org.jbpm.workbench.ht.model.events.TaskCompletedEvent;
 import org.jbpm.workbench.ht.model.events.TaskRefreshedEvent;
 import org.jbpm.workbench.ht.service.TaskService;
 import org.jbpm.workbench.ht.util.TaskStatus;
@@ -70,9 +69,6 @@ public abstract class AbstractHumanTaskFormDisplayer<S extends FormRenderingSett
 
     @Inject
     protected Event<TaskRefreshedEvent> taskRefreshed;
-
-    @Inject
-    protected Event<TaskCompletedEvent> taskCompleted;
 
     @Inject
     protected User identity;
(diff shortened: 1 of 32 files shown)


bsremac:
These are the potential problems:

1. Do we need _Actions_ column in task comments after the task has been completed?

2. The task comments are not available after the process instance completes or is aborted. Is that expected?

3. Columns resizing is buggy. Have a look at the screencast attached below.
[Screencast from 27-08-18 11:41:09.zip](https://github.com/kiegroup/jbpm-wb/files/2323540/Screencast.from.27-08-18.11.41.09.zip)

4. The default columns width is too small. The full date doesn't fit. Since the date format is fixed, I think we could ensure the full visibility.
screenshot from 2018-08-27 11-40-52
5. Refresh option in task master details doesn't update comments tab upon task completion, which leaves _Delete_ action available and working even after the task was completed (in another browser window or the same or task admin user).

6. When the container is undeployed and then again re-deployed, the task comments are not available anymore.

cristianonicolai:
@bsremac let me try to answer these

1. Probably not, but I think is best to leave it as is to not have the table using different sizes for different situations. Also if you dont have permission to delete any comment you will also see an empty actions column. This UI needs to be refactored to align with the Case Comments block, then is where I think we have the opportunity to fix this.
2. Correct, that is due to the task clean up job, see: https://github.com/kiegroup/kie-docs/pull/1011
3. Again, I don't think is probably worth fixing this now as we should, in fact, refactor the UI to use a list of comments instead of a table.
4. Ok, I will increase it but strangely it works fine for me.
5. Historically, I don't think we have good support for this kind of situation, just realy relying on a refresh to see the latest updates. In this case, I think I can manage to get it working, will try to add it in the next version on this PR.
6. I think this is probably also related to item 2, need to disable the task clean up job in the Kie Server. I tested it locally and it seems fine with: starting/stopping the container as well as removing and re-deploying the container with the same name. Just a note that redeploying the container with the same name isnt a good practice and should be avoided.

nmirasch:
@cristianonicolai  @bsremac  about the 3 and 4 points.. Both worked fine for me too( tested on firefox 61.0.2 and chrome 68.0.3440.106)
About the task selection event, When I was dealing with that I thought that was interesting do some check about the task selection event, something like if (currentTaskId == 0 || currentTaskId == event.getTaskId()) just to not react to any task selection event. 
I think that refactor will have to be done and many of this table issues will be solved.
At creation point the view can know if the action column have to be added or not but it's true that the refresh can't remove that column (Pagedtable is not prepared for remove columns and like @cristianonicolai  said I don't think it's worth now having to refactor that in a near future)

cristianonicolai:
@bsremac @nmirasch I updated the PR, now also including a fix for item 5. The drawback is that the task list used to refresh the grid every time a task is completed. This same event is now global, will propagate to all users in the workbench, which means I had to store the task the user selected from the list to make sure it only refresh the grid when that same task is completed.

bsremac:
Depends on https://github.com/jboss-integration/bxms-qe-tests/pull/1719

cristianonicolai:
@bsremac @nmirasch Included some refactor to address the task selection concern. I think it's better now, at least the behavior is consolidated in a common class.

PR #1191, opened 2018-08-30
JBPM-7665: Process Instances: page with vertical scroll bar is flicke…

nmirasch:
…ring after changing window width

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridView.html
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridView.html
@@ -1,8 +1,8 @@
-<div style="padding: 20px 5px;">
-     <div class="row">
-         <div class="col-md-12 manage">
-             <div data-field="active-filters"></div>
-             <div data-field="column"></div>
-         </div>
-     </div>
+<div class="manage">
+    <div class="kie-manage-item-list">
+        <div style="padding-right: 5px;">
+            <div data-field="active-filters"></div>
+            <div data-field="column"></div>
+        </div>
+    </div>
 </div>
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/ListTable.java
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/ListTable.java
@@ -33,6 +33,8 @@ public class ListTable<T extends GenericSummary> extends ExtendedPagedTable<T> i
         this.addDataGridStyles("kie-datatable",
                                "kie-datatable");
         this.dataGrid.setStriped(false);
+        super.setPageSizesSelectorDropup(true,
+                                         false);
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


nmirasch:
@cristianonicolai @bsremac I have added the css changes with that fix for the movement of the datagrid lines.

bsremac:
@nmirasch I've noticed one issue around the number of items per page selection. Have a look at the screenshot below.

screenshot from 2018-09-01 17-29-46

cristianonicolai:
same here, the dropdown is somehow hidden now. Although the flickering isn't happening, I can see the table getting pushed to the left as it would allow space for the scroll to appear.

nmirasch:
@cristianonicolai @bsremac  thanks for review it! yes I changed that style to leave space for the dropdown.

nmirasch:
@bsremac @cristianonicolai this fix is tied to the size of page size options. After the meeting I found that maybe the list of options can change maybe have more than 4 options.. If that the case this issue you comment would be still present. Let me give this another try.

nmirasch:
@cristianonicolai @bsremac I found this is the best solution for this 4 page size options i this css. If in the future  we want to allow more options I see different solutions like only let 4 items visible or recalculate dynamically his padding using the configured options size.
Does this make sense? Any thought on this is very welcome. Thanks!

bsremac:
@nmirasch I guess the approach on the sizing of options directly depends on the implementation of https://issues.jboss.org/browse/JBPM-7696.
The currently proposed implementation in open PRs assumes the same options (5,10,15,20). Without any bigger changes in the proposed implementation, this number could potentially increase to 5 options, with the 5th one being an additional value which could be set as a global page size preference.
This is just my current view on where this could go, but I will let @cristianonicolai give the final opinion.

cristianonicolai:
@nmirasch @bsremac just thinking here, what if the pagination select drops up instead of down? That would remove the need to handle this extra padding at the bottom, it would be more friendly to visualize as it could force some scroll to happen when open and also future safe in terms of how many options are listed, wdyt?

nmirasch:
@cristianonicolai  that for sure, would fix the issue I have created this PR https://github.com/kiegroup/appformer/pull/448 with that style change but in the future be have to limit the page sizes visible options anyway. 
@bsremac @cristianonicolai I restored the previous padding value here then that appformer change is needed to merge this pr. Could you take another look?

bsremac:
@nmirasch @cristianonicolai I see no problems with that approach. Interacting with items from the list and changing the pagination are mutually exclusive operations, so I see no problem in selection consuming some of the list space.
Maybe it's worth passing this idea by Liz, just to hear her opinion, since it's a user experience topic above everything else.

bsremac:
@nmirasch Looks good to me.

cristianonicolai:
@nmirasch just to clarify, the selector should already dropup if there is no space right? I'm not sure we need to force the dropup everywhere.

nmirasch:
@cristianonicolai the problem is that drop-down uses the navigation window to know if he has space to be a drop down or drop up and not the div where it is contained. The area that contains the list of items and the pagination bar have to calculate it's own height to be able to remove the space between the list and the bottom bar. When that height is too small to contain the drop-down opened but the navigation window have space to contain that, the page size selector appears as a drop down and includes an hidden scroll that made that effect that @bsremac commented. 

Configuring the page selector always as a dropup removes the scroll problem. Other solution would be to limit the number of items of the selector and include that padding value that allow see that dropdown opened.

cristianonicolai:
ok, got it, so maybe making only the ListTable paginator to always dropup instead of all pagedtables?

nmirasch:
@cristianonicolai sure that makes more sense. I have added a method to pagedTable to configure that and the call to ListTable to only affect this tables.
His PR depends now on https://github.com/kiegroup/appformer/pull/448

cristianonicolai:
@nmirasch can you cherry pick this into 7.11.x please.

PR #1192, opened 2018-09-03
JBPM-7696 - Use default preferences for paging option in manage section

cristianonicolai:
Depends on https://github.com/kiegroup/kie-wb-common/pull/2082

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-api/pom.xml
+++ b/jbpm-wb-common/jbpm-wb-common-api/pom.xml
@@ -46,6 +46,15 @@
       <groupId>org.jboss.errai</groupId>
       <artifactId>errai-common</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.uberfire</groupId>
+      <artifactId>uberfire-preferences-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.uberfire</groupId>
+      <artifactId>uberfire-preferences-processors</artifactId>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
 
 </project>
--- a/jbpm-wb-common/jbpm-wb-common-api/src/main/java/org/jbpm/workbench/common/preferences/ItemPerPageValidator.java
+++ b/jbpm-wb-common/jbpm-wb-common-api/src/main/java/org/jbpm/workbench/common/preferences/ItemPerPageValidator.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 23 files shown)


cristianonicolai:

bsremac:
@cristianonicolai I don't know but it looks really weird from the user experience point of view to have a different number of items per page set in different Manage views via this global pagination setting just because the user has changed the number of columns in one or more views at one point but has never changed the pagination option itself.
I guess decoupling table and pagination preferences would be the only real solution for this inconsistency, which largely complicates the design and is most likely not the way to go (if it's possible altogether).

Another bad aspect of pushing some default pagination when a user has already done some local changes is the element of surprise when they find inconsistent pagination across Manage views without knowing the context. I guess, adding the word _default_ to the number which is set globally would give some sense and even serve as a recommendation to users if in one point they decide to follow the general pagination setting. Something like this:
screenshot from 2018-09-04 14-09-59

nmirasch:
@cristianonicolai @bsremac maybe that makes sense, let the user know  he is not using the default value in some way, But the user would be confuse anyway, when, no changing nothing, it shows a selected value, different from default if he has visited the view before. 
Another question is if we would have to apply this table size configuration to all the console tables like process variables or documents, I see at doc PR that configuration only affects at tables under manage items, but as user maybe I would expect they were affected by this configuration parameter.

cristianonicolai:
Assembly:
- https://github.com/kiegroup/appformer/pull/480
- https://github.com/kiegroup/kie-wb-common/pull/2082
- https://github.com/kiegroup/jbpm-wb/pull/1192

cristianonicolai:
@bsremac @nmirasch please take another look on this, the proposed solution is to simply add two more options to the pagination dropdown ( 50, 100 ) as well as allowing admins to setup a default value. I agree that changing the columns should not impact on saving the paginated option but I think that should be its own JIRA. As for showing the default label, I'm not sure is worth, as in most cases this kind of settings will be done after the console is installed, will be transparent for most users that the default value will be already in place.

bsremac:
@cristianonicolai Agree on the default label, it indeed sounds like an overkill when I think about it a bit more :)
With the addition of two more pagination options, the fairly recent choices (@nmirasch :)) have come back to haunt us :)
Process Definition perspective without or only one definition trims the top of the drop-up menu.

screenshot from 2018-09-25 13-15-34

Which bring me to the question of should we even expose pagination in case the list has < 5 items? Another own JIRA candidate? :)

cristianonicolai:
@bsremac good point, to be honest, I don't seem much value in keeping 5 and 15 options. Removing that would be enough to fix this but if not, we can probably extend the height in the definition list. Not showing the pagination option is not an easy fix and in general terms for user experience, we usually try to avoid the show/hide effect. Even if in this context is not really useful.

bsremac:
Depends on https://github.com/jboss-integration/bxms-qe-tests/pull/1843

PR #1196, opened 2018-09-10
[7.11.x] JBPM-7665: Process Instances: page with vertical scroll bar is flicke…

nmirasch:
…ring after changing window width (#1191)


nmirasch:
Depends on https://github.com/kiegroup/appformer/pull/453

PR #1197, opened 2018-09-12
[Do not merge] Fix for ECJ - GWT conflict

nmirasch:


nmirasch:
@cristianonicolai @bsremac  @manstis  could you take a look?

manstis:
@nmirasch Just to check, have you verified this fix works at _runtime_ (i.e. you've ```mvn clean process-resources gwt:run``` for both?)

nmirasch:
@manstis tried mvn clean gwt:run. Should I try something else?

manstis:
@nmirasch Nope, if both webapps run OK then that's perfect :-)

nmirasch:
@cristianonicolai  @bsremac @manstis The fix worked fine with jbpm-wb-showcase. just trying with the case jbpm-wb-case-mgmt-showcase, compilation was right but running that I got some errors, let me see if that are related with this fix. This is the error "Error: InvalidBusContentException: Invalid bus content! Expecting content type 'application/json', but got 'text/html;charset=UTF-8!"

nmirasch:
@cristianonicolai @bsremac @manstis false alarm, It was related with my environment. It works fine for both showcases.

nmirasch:
the *real* fix is merged in errai now

PR #1203, opened 2018-09-19
JBPM-7765 - Add offset to row index

cristianonicolai:


bsremac:
@cristianonicolai Looks good, this is definitely something Selenium tests are not covering.
Just to double check the behavior we have - select all should cover only the current page, not all pages, right?

cristianonicolai:
@bsremac correct.

PR #1210, opened 2018-10-01
JBPM-7787: Implement new process instance log timeline view

nmirasch:

PR #1214, opened 2018-10-09
JBPM-7788: Implement expand information for nodes logs in process instance logs

nmirasch:


nmirasch:
@cristianonicolai  @bsremac Just waiting for the previous https://github.com/kiegroup/jbpm-wb/pull/1210 getting merged, I want to share the human task details expanded code. This is not a standard PR because it contains two commits but the goal is know your opinion, Once the other would be merge I'll rebase this one. Could you take a look? Thanks!

bsremac:
@nmirasch Task details look good. Have a just one minor enhancement suggestion which now somehow looks more obvious when exposing it in the logs as well.
Namely, **UserID** parameter is empty when the task is not reserved. We have the same behavior in Taks _Details_ tab. Maybe replacing the empty string with a message along the lines of: "Task not yet claimed" or even simpler "Not claimed" would be more informative when looking at the log details.
screenshot from 2018-10-10 09-08-32

nmirasch:
@bsremac added 'Not claimed' when there is not task owner

cristianonicolai:
@nmirasch I'm looking at the _TaskInstance_ object returned in the Kie Server call, I think there is some usefull information that we could add in here like: potential owners, business admins, excluded owners and createdby. wdyt including it in the UI? Also, instead of timestamps, perhaps label it _Created On_?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/ProcessInstanceLogPresenter.java
```diff
@@ -153,81 +141,43 @@ public boolean onError(ClientRuntimeError error) {
...
-        view.setActiveLogOrderButton(LogOrder.ASC);
-        view.setActiveLogTypeButton(LogType.BUSINESS);
+    public void loadTaskDetails(Long workItemId,
+                                ProcessInstanceLogHumanTaskView humanTaskView) {
+        taskService.call(
+                (TaskSummary task) -> {
+                    humanTaskView.setDetailsData(task);
+                }).getTask(serverTemplateId,
+                           containerId,
+                           workItemId);
```
cristianonicolai: @nmirasch I think you probably need to call another service here, one that finds the task based on the _workItemId_,  check if _UserTaskServicesClient.findTaskByWorkItemId_ can help with that.

bsremac:
@nmirasch I have just noticed another potential problem in the details content.
The timestamp doesn't change. It always shows the time when the task was entered. The other thing which is uncharacteristic for logged data is that the already logged data changes as the corresponding node transitions from one state to the other. In this way, inspecting details loses the value as all the history is lost. Have a look at the screenshot:
screenshot from 2018-10-11 13-06-00

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/ProcessInstanceLogPresenter.java
nmirasch: @cristianonicolai about showing extra TaskInstance data like, potential owners, business admins, excluded owners and createdby. All this info is only filled by UserTaskServicesClient.getTaskInstance. We can use this call instead UserTaskServicesClient.findTaskByWorkItemId (it don't fill that info) you suggested if we want to show that information and add an extra field TaskAssignmentSummary to TaskSummary to make available that info. Make it sense?

nmirasch:
@bsremac yes, the date displayed is when it was Created, I just update the label from timestamp to CreatedOn. Right now I'm adding to the details more information, maybe makes sense add lastModificationDate there?

bsremac:
@nmirasch When it comes to adding any dates to details, it depends on how accurate we want to be with the logs, to begin with. We have the time the event was logged already shown in a broad form i.e: _3 hours ago, a few seconds ago_ etc. Since we are logging only when the task was entered (which is task creation date) and when it was completed (which is the last modification date), we already have this information as part of the new timeline. It only makes sense to add these two dates to details if we want to provide the exact time with to the second accuracy.

The other thing I was trying to point out with the previous comment and the screenshot is that most of the task details will differ between the time the task was entered and the time it was completed, but in the logs we are showing the same information for both events, being always the current task details. 
With this in mind, the whole exposure of the task details looses a lot of its value as it deosn't provide any insight in what the details were when the task was entered and hence reduces the potenital to draw any conlusions from ispecting the logs, such as what was the task status when created, did it have an owner or not, did the owner who was initaly assigned also complted the task or was it somebody else etc. Basically all this information is lost from the logs when the current data is only data that is available in the logs.

nmirasch:
@bsremac @cristianonicolai its true, the data retrieved for the same workitem id is the same for both. In the current implementation I just added the log date as lastUpdate, at least human task log nodes that info will show the time (not relative like in the timeline) when it was created and when it was completed.
About the extra information like potential owners, business admins, excluded owners I have to use that getTaskInstance() instead of findTaskByWorkItemId but it doesn't work for completed instances. Maybe include there an extra call to retrieve that extra information in case the instance is in the proper status?

cristianonicolai:
@nmirasch I think is fine as it is, lets get some feedback first before having the extra information there.
@bsremac  I agree that in terms of a timeline I would expect that the detailed information is relative to happened on that point in time. The information we've tough is very limited, for tasks for instance, we don't keep a historical version of the data. So to me, it is really designed to bring some more context to the user about a node, ex avoiding to jump to another screen, and that is the current details. Maybe we need to make it more clear in the UI that this is the updated information. The way it works now it's really to link the nodes with their contextual data.

nmirasch:
jenkins please retest this

cristianonicolai:
@bsremac good point, I just wondering how would that play when we introduce filtering, lets say you filter out by entered events, then all details for completed ones would be lost. We could check the last appearance in the UI only but I have a feeling that it would only complicate the user experience and the coding. I would probably suggest get this initial version as is and get feedback for an enhancement of top of this, based on what we discussed here. @nmirasch wdyt?

cristianonicolai:
jenkins please retest this

cristianonicolai:

bsremac:
@cristianonicolai You are right. With filtering introduced it makes sense to have details attached to the human task nodes no matter their "node state". I can't think of an easy logical-wise solution which would amend both problems at the same time.
+1 to keep it then as it is.

bsremac:
Related Selenium testing is tracked with https://github.com/jboss-integration/bxms-qe-tests/pull/1885

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/ProcessInstanceLogHumanTaskView.html
```diff
@@ -0,0 +1,18 @@
...
+        <dd><span data-field="createdOn"/></dd>
+        <dt><span data-i18n-key="CreatedBy"/></dt>
+        <dd><span data-field="createdBy"/></dd>
+        <dt><span data-i18n-key="Id"/></dt>
+        <dd><span data-field="taskId"/></dd>
+        <dt><span data-i18n-key="State"/></dt>
+        <dd><span data-field="taskStatus"/></dd>
+        <dt><span data-i18n-key="Owner"/></dt>
+        <dd><span data-field="actualOwner"/></dd>
+        <dt><span data-i18n-key="Content"/></dt>
```
cristianonicolai: @nmirasch @bsremac wdyt about using _Description_ instead in here?

PR #1229, opened 2018-11-01
JBPM-7858: Implement expand information for Work Item nodes logs in p…

nmirasch:
…rocess instance logs


jbpm-wb-process-runtime/jbpm-wb-process-runtime-api/src/main/java/org/jbpm/workbench/pr/model/WorkItemParameterSummary.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.io.Serializable;
+
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.jboss.errai.databinding.client.api.Bindable;
+import org.jbpm.workbench.common.model.GenericSummary;
+
+@Portable
+@Bindable
+public class WorkItemParameterSummary extends GenericSummary implements Serializable {
```
cristianonicolai: Please add _toString_ as well.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-api/src/main/java/org/jbpm/workbench/pr/model/WorkItemSummary.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.jboss.errai.common.client.api.annotations.NonPortable;
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.jbpm.workbench.common.model.GenericSummary;
+
+@Portable
+public class WorkItemSummary extends GenericSummary<Long> {
+
+    private Integer state = 0;
+    List<WorkItemParameterSummary> parameters;
```
cristianonicolai: private

jbpm-wb-process-runtime/jbpm-wb-process-runtime-api/src/main/java/org/jbpm/workbench/pr/model/WorkItemSummary.java
```diff
@@ -0,0 +1,85 @@
...
+package org.jbpm.workbench.pr.model;
+
+import java.util.List;
+
+import org.jboss.errai.common.client.api.annotations.NonPortable;
+import org.jboss.errai.common.client.api.annotations.Portable;
+import org.jbpm.workbench.common.model.GenericSummary;
+
+@Portable
+public class WorkItemSummary extends GenericSummary<Long> {
```
cristianonicolai: please add _toString_

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/ProcessInstanceLogItemDetailsView.java
```diff
@@ -0,0 +1,163 @@
...
+        if (model.getCreatedBy().isEmpty()) {
+            createdByLabel.setHidden(true);
+            createdBy.setHidden(true);
+        } else {
+            createdBy.setTextContent(model.getCreatedBy());
+        }
+        updatedOn.setTextContent(DateUtils.getDateTimeStr(logDate));
+        taskId.setTextContent(model.getId().toString());
+        taskStatus.setTextContent(new TaskStatusConverter().toWidgetValue(model.getTaskStatus()));
+        if (model.getActualOwner().isEmpty()) {
```
cristianonicolai: Should this be _model.getActualOwner() == null_ ?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/ProcessInstanceLogItemDetailsView.java
```diff
@@ -0,0 +1,163 @@
...
+        return nodeDetailsContainer;
+    }
+
+    public void setTaskDetailsData(final TaskSummary model,
+                                   Date logDate) {
+        removeCSSClass(humanTaskContainer,
+                       "hidden");
+        taskId.setTextContent(model.getId().toString());
+        createdOn.setTextContent(DateUtils.getDateTimeStr(model.getCreatedOn()));
+        if (model.getCreatedBy().isEmpty()) {
```
cristianonicolai: Same here, should it be null?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/log/ProcessInstanceLogItemDetailsView.java
```diff
@@ -0,0 +1,163 @@
...
+            createdByLabel.setHidden(true);
+            createdBy.setHidden(true);
+        } else {
+            createdBy.setTextContent(model.getCreatedBy());
+        }
+        updatedOn.setTextContent(DateUtils.getDateTimeStr(logDate));
+        taskId.setTextContent(model.getId().toString());
```
cristianonicolai: Seems to be duplicated _taskId_ in here.

nmirasch:
@cristianonicolai Thanks, updated!

jsoltes:
jenkins please execute full downstream build

jsoltes:
@nmirasch looks good to me, let's see what full downstream build produces. :)

jsoltes:
@nmirasch overall looks fine, just two things I noticed when I was trying it out in the UI. 

1. Right after the start of the process, there is ugly html code in the details of the task. I am not sure, what is the purpose of the NotCompletedNotify and NotStartedNotify properties, but it looks like some kind of system message. Can this be omitted from the Work item data display? Or can it at least be formatted to some more readable format e.g. applying the html properties, or omitting the tags?
htmlinnotifyparts
2. Sometimes the work item data are present and sometimes not. Is this on purpose?
missing_work_item_data

nmirasch:
@jsoltes yes, that's expected, only adds the workitem details when the workitem related is available at the engine, it means the task haven't been finished (completed,..).

nmirasch:
@jsoltes for the first point, that ugly html depends on the process definition that defines the what are the workitems parameters. In this example that html is the content of the different notifications managed at that task

mswiderski:
@nmirasch @jsoltes I would actually filter out some of the task variables and the Notification and Reassignment are good examples of it. Similar would be skippable or ActorId ... these are added by the engine and not user so is of very little value in this context I'd say. wdyt?

nmirasch:
@mswiderski Thanks, I think that is a very good idea, I only worried about how to identify those parameters. Currently, the UI display all the work item parameters (just doing toString over the value object). I'm not sure how we could identify what parameters display or not, Any suggestion is very welcome (as always)

jsoltes:
@mswiderski yeah, I agree. We have to have in mind these details serve for people to have some overview so these don't have a value. @nmirasch for me it would be enough to exclude the ugly Notifies, but you can also exclude the ActorId as Maciej suggested. I dunno, is it possible to exclude it by the name of the parameter?

nmirasch:
@jsoltes The problem is that parameters names depend on the process definition, we don have any control from the UI about what could be there it's just a list of pairs (name, value)

jsoltes:
@nmirasch well, if those engine related parameters have always the same name for each process you could go through the list and filter out the NotCompletedNotify and NotStartedNotify parameters with their values... Just guessing, I don't really see much into the code.

mswiderski:
I would say we should have a predefined list of list variables (task inputs) that should be filtered out, to start with I'd go for

- TaskName
- ActorId
- GroupId
- Skippable
- NotStartedReassign
- NotStartedNotify
- NotCompletedReassign
- NotCompletedNotify
- BusinessAdministratorId
- BusinessAdministratorGroupId
- TaskStakeholderId
- ExcludedOwnerId
- RecipientId

you can put this either directly in code or have some sort of property file that can be easily updated if needed.

jsoltes:
@mswiderski +1 for the property file.

nmirasch:
@mswiderski Currently we are retrieving the workItem from the context, not doing the query directly over that table for the parameters.
Here https://github.com/kiegroup/jbpm/blob/563a11c39253b02bacda8e6065eea41d154775b3/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java#L609 is the service call. 
I'm not sure if your incoming changes can affect the parameters loaded at the context related with the work item.
@cristianonicolai, @mswiderski  My question is looking at the proposed list of parameters to filter, Does make sense to show other workitem parameter for a human task? What do you think?

cristianonicolai:
@nmirasch I guess the only value in there would be custom task variables as everything else would be hidden by the engine or UI. Maybe we could start without wi details for human tasks for now.

mswiderski:
completely agree with @cristianonicolai that it does not make sense to show work item data for human tasks. Instead you should rely on task data (both input and outputs)

nmirasch:
@cristianonicolai @mswiderski @jsoltes  Updated the PR, Now, when the instance node have a workitem id and it's not a HumanTaskNode displays the workitem details. Could you take a look? Thanks!

cristianonicolai:

PR #1247, opened 2018-11-28
JBPM-7807 - Add support for triggering nodes in process instance

cristianonicolai:
Depends on https://github.com/kiegroup/appformer/pull/552

### DIFF
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-api/src/main/java/org/jbpm/workbench/pr/model/ProcessNodeSummary.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-api/src/main/java/org/jbpm/workbench/pr/model/ProcessNodeSummary.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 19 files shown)


cristianonicolai:
@nmirasch @jsoltes initial implementation, it allows selecting a node and triggering it but still no interaction in the actual process diagram. I believe is reasonable enough to be merged already. Next step is to add node instances and timers information. The final step will be the actual selection from the diagram, similar to what is done when you manually select a node from the select box.

cristianonicolai:

jsoltes:
@cristianonicolai codewise looks good, lets see if full downstream build produces anything.

cristianonicolai:

jsoltes:
@cristianonicolai looks like there is a compilation error on webapp in full-downstream build...

cristianonicolai:

cristianonicolai:
@jsoltes let me try to address some of these.
1 - I'm appending the node type and id to nodes that do not have an explicit name. The reason why, is that I only use the node type, ex: split, the process might have multiple of these and the users wouldn't know which one to match. I'm not using a sequence number for the specific node type is any better. At least this way, when you expand the node details, you can clearly see that this is the actual id for that node.
2- Goal is that the selection will be interactive, by selecting a process node from the diagram. We could change the order or items in the select, but I find having an alphabetical can be more consistent and easier for users to find the node they need. A process might have many different branches, so deciding on what comes first is not really straightforward.
Will take a look on how to handle items 3, 4 and 5.

jsoltes:
@cristianonicolai thank you. :) 
I have an idea, that would solve the 1. and 2. and isn't that hard to implement. Let's order the selector according to the engine id. And for consistency, could you add it to the human tasks as well? Also, maybe it would look better if the engine id was at the beggining. So instead of 
> bla-4
blabla-2
blablabl-3
bl-1

It would be
> 1-bl
2-blabla
3-blablabl
4-bla

what do you think?

cristianonicolai:
Closing PR as changes are incorporated into https://github.com/kiegroup/jbpm-wb/pull/1250

PR #1250, opened 2018-12-10
JBPM-7807 - Process instance admin operations 

cristianonicolai:


cristianonicolai:
@nmirasch @jsoltes please note that I found an issue regarding the re-scheduling a timer instance that needs fixing in jbpm core. Will have a separated PR for that.

cristianonicolai:
Related PR to updating timer instance https://github.com/kiegroup/jbpm/pull/1392

jsoltes:
@cristianonicolai can you please fix the compilation issues?

cristianonicolai:
jenkins please retest this

cristianonicolai:

jsoltes:
@cristianonicolai Seems that all the bugs I mentioned previously are fixed. I have some new questions, though...
1. I noticed that now it is possible to use trigger button only once per process instance, after that it doesn't appear on any of the nodes. The same goes for the retrigger. Is this intended?
2. When you go through the selector instances, just selecting one-by-one, the Trigger button is there for first three, but then when you return to them, it disappears.
3. Why is there an extra dropdown for Timer instances (could the timers be part of the Node instances dropdown)? Could it be hidden, when no Timer instances are present in the process?
4. When you try to reschedule the timer instance, you get an error: `Unable to complete your request. The following exception occurred: "Unexpected error during processing: null".` (I tried to change the period in a timer with setup Fire once after druation: P10D).

cristianonicolai:
@jsoltes 

1- Fixed the trigger button. Retrigger seems ok to me, it only shows for node instance that are in progress.
2 -Should be fixed now.
3- They are really different concepts IMHO. Maybe it can be combined, but there is no data to correlate it now. Something to be discussed with the team perhaps.
4 - Should be related to https://github.com/kiegroup/jbpm/pull/1392

jsoltes:

jsoltes:
@cristianonicolai 
1. The trigger seems ok now. :+1: 
2. The retrigger produces two new instances - one with the same id and one with the id raised by 1. I think it is bug (I have mentioned this one already on the closed pr).
3. Ok. Can you please hide the Timer Instances dropdown, when there are no timer instances in the process?
4. Sure, I forgot about that.
5. (a new one) Now that is possible to abort the process by triggering the endNode, the business central can get into inconsistent state (which produces unexpected behavior when clicking on Abort for example). I suggest as after action refreshing the diagram view (it should look the same as on completed instances - no abort and signal button present).

cristianonicolai:
@jsoltes 

2 - Not likely this is something caused by the UI, as we only get the list back from the Kie Server. I would prefer to investigate this separately once this is merged, does't seem like a blocker and changes are getting quite large in this PR.
3 - Done
5 - Done

cristianonicolai:
Jenkins execute compile downstream build

PR #1253, opened 2018-12-11
JBPM-8046 - Select inputs should be initialized before model is set

cristianonicolai:
Depends on https://github.com/kiegroup/appformer/pull/567


cristianonicolai:
jenkins please retest this

PR #1270, opened 2019-01-07
JBPM-8110: Incorrect log order for events occurring within 1s. Use th…

nmirasch:
…e id gererated by the DB sequence to sort instead the timestamp
- Changed the taskEvents sort field form logTime -> id generated by seq.
- Replaced the nodeinstanceId used at sorting info from process instance log by the id generated by seq.


nmirasch:
This code try to avoid issues when the database stores logs with the same timestamps because not have enough precision to differentiate between them. In the proposed implementation the ids generated by the db sequence are used to resolve that conflicts at process instance logs or to be specified as the sort fields at tasks logs 
@cristianonicolai @jsoltes could you take a look, please? Thanks!

PR #1284, opened 2019-02-06
[do not merge] JBPM-8206 - Load process and task variables by default

cristianonicolai:


cristianonicolai:

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/list/ProcessInstanceListPresenter.java
```diff
@@ -164,8 +164,7 @@ protected DataSetReadyCallback getDataSetReadyCallback(final Integer startRange,
                             lastPage = true;
                         }
 
-                        final String filterValue = isFilteredByProcessId(tableSettings.getDataSetLookup().getOperationList());
-                        if (filterValue != null) {
+                        if (preferences.getLoadProcessVariablesByDefault() || isFilteredByProcessId(tableSettings.getDataSetLookup().getOperationList()) != null) {
```
nmirasch: @cristianonicolai In this case:
- The preference is set to loadProcessVariablesByDefault.
- The user selects one of the domain columns and the table shows the right value
- The preference is changed to Not loadProcessVariablesByDefault.
- When the user access to the processInstances again, This domain column is still present but this condition makes that the content is not loaded. It can be seen a bit inconsistent regarding it don't show the value because of the preferences, maybe we have to check if the displayed columns are available in the applied dataset. What do you think?

cristianonicolai: @nmirasch yeah, I saw that, but not sure how to get around, because in this case we're mixing variables from different tasks and processes, so how to differentiate that these are empty or the preference has been turned of?

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskslist/AbstractTaskListPresenterTest.java
```diff
@@ -283,80 +286,66 @@ public void isFilteredByTaskNameInvalidTest() {
...
+
+        testDomainSpecificData();
+    }
+
+    protected void testDomainSpecificData() {
         when(dataSetMock.getRowCount()).thenReturn(1);//1 task
         //Task summary creation
-        when(dataSetMock.getValueAt(0,
-                                    COLUMN_TASK_ID)).thenReturn(Long.valueOf(1));
+        when(dataSetMock.getValueAt(0, COLUMN_TASK_ID)).thenReturn(Long.valueOf(1));
```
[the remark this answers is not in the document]
cristianonicolai: @jstastny-cz yes, it's my fault :) we agreed to change the code styling, only breaking lines when it actually makes sense. So now I started to clean up a bit in the files I'm changing, it removes quite a few unnecessary lines.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/test/java/org/jbpm/workbench/ht/client/editors/taskslist/AbstractTaskListPresenterTest.java
```diff
@@ -16,12 +16,9 @@
 
 package org.jbpm.workbench.ht.client.editors.taskslist;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
cristianonicolai: I usually, and this is how my IDE is setup, do a wildcard import for whenever there are 4 or more imports. Its seems reasonable to me.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/list/ProcessInstanceListPresenter.java
nmirasch: @cristianonicolai maybe create a new method at ProcessInstanceListViewImpl and remove that extra columns if that condition at the presenter don't retrieve the dataset that could fill the content..something like this:
    public void removeDomainSpecifColumns(ExtendedPagedTable<ProcessInstanceSummary> extendedPagedTable) {
        for (ColumnMeta<ProcessInstanceSummary> cm : extendedPagedTable.getColumnMetaList()) {
            if (cm.isExtraColumn()) {
                extendedPagedTable.removeColumnMeta(cm);
            }
        }
        extendedPagedTable.storeColumnToPreferences();
    }
It's true that the extra column selection is lost, but at least it prevents empty content when it's not empty. It's just an idea but we have to think if to add this call makes sense regarding the possible situation it tries to prevent.

PR #1286, opened 2019-02-08
JBPM-8215 - Display delay time in seconds

cristianonicolai:


nmirasch:
@cristianonicolai I think there is something wrong at seconds conversion. The delay requested is converted to milliseconds, that as far as I know is right, but the action translates it to seconds delay

selection_002

selection_003

maybe I'm doing something wrong..

cristianonicolai:
@nmirasch good catch thanks, hopefully fixed now.

cristianonicolai:

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/diagram/MillisecondsToSecondsConverter.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.workbench.pr.client.editors.instance.diagram;
+
+import org.jboss.errai.databinding.client.api.Converter;
+
+public class MillisecondsToSecondsConverter implements Converter<Long, String> {
```
jstastny-cz: Shouldn't be the converter methods toModelValue and toWidgetValue inverse functions?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/diagram/MillisecondsToSecondsConverter.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public Long toModelValue(final String time) {
+        if (time == null || time.trim().equals("")) {
+            return 0l;
```
jhrcek: I'd recommend using upperace `L` to denote long literals. `l` looks too similar to `1` and which can lead to confusion.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/diagram/MillisecondsToSecondsConverter.java
cristianonicolai: the model brings value in milis but the update method takes seconds, so its just one way. Once its been converted to seconds it remains like that.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/diagram/MillisecondsToSecondsConverter.java
cristianonicolai: Done

cristianonicolai:
@jhrcek @jstastny-cz you guys happy for this to get merged?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/diagram/MillisecondsToSecondsConverter.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class MillisecondsToSecondsConverter implements Converter<Long, String> {
+
+    @Override
+    public Long toModelValue(final String time) {
+        if (time == null || time.trim().equals("")) {
+            return 0L;
+        }
+
+        return Long.valueOf(time);
```
jhrcek: Can it ever happen that the string contains non-numeric non-whitespace characters? E.g. "abc"?

```java
Long.valueOf("abc")
|  Exception java.lang.NumberFormatException: For input string: "abc"
```

Either way it would be nice to have a test which either expects exception or modify the method to use 0 as default in these cases..

cristianonicolai: Sure, included handling for this.

PR #1291, opened 2019-02-13
JBPM-8239:Process Instance log 'Sub Process' filter not matching expe…

nmirasch:
…cted entries
Fixed typo error at SubProcessNode


nmirasch:
@cristianonicolai @jstastny-cz could you take a look? thanks!

PR #1294, opened 2019-02-18
Fix springboot on JDK 11 (jax-ws-api)

MarianMacik:
depends on kiegroup/jbpm#1443


MarianMacik:
Test failures not related (they have been failing for 6 builds)

PR #1299, opened 2019-02-20
JBPM-8268 - Fix node selection when diagram is generated by jbpm-desi…

cristianonicolai:
…gner


jstastny-cz:
@cristianonicolai  the issue appears only for legacy business processes, that's why the xmlns:oryx check and change in path, right?

cristianonicolai:
@jstastny-cz exactly.

PR #1300, opened 2019-02-21
AF-1663: Fix tests to adapt to D3 changes

jesuino:
depends on https://github.com/kiegroup/appformer/pull/646


cristianonicolai:
@jesuino perhaps this class could be replaced with plain mocks, something like:
`mock( Foo.class,  withSettings().extraInterfaces( Runnable.class )` ?

jesuino:
@cristianonicolai thanks for the suggestion! I added a new commit, please let me know if it looks better for you.

PR #1309, opened 2019-02-26
DROOLS-3425 : Unexpected behavior of docks when multiple editors are opened

Rikkola:
https://issues.jboss.org/browse/DROOLS-3425

https://github.com/kiegroup/kie-wb-common/pull/2503
https://github.com/kiegroup/drools-wb/pull/1082
https://github.com/kiegroup/jbpm-designer/pull/837
https://github.com/kiegroup/jbpm-wb/pull/1309
https://github.com/kiegroup/optaplanner-wb/pull/327

### DIFF
--- a/jbpm-wb-integration/jbpm-wb-integration-client/src/main/java/org/jbpm/workbench/wi/client/casemgmt/CMMNEditorPresenter.java
+++ b/jbpm-wb-integration/jbpm-wb-integration-client/src/main/java/org/jbpm/workbench/wi/client/casemgmt/CMMNEditorPresenter.java
@@ -34,8 +34,10 @@ import org.uberfire.mvp.PlaceRequest;
 import org.uberfire.workbench.model.menu.Menus;
 
 @Dependent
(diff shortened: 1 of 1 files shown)


cristianonicolai:
jenkins please retest this

kkufova:
I agree with @manstis, this is the only repo that's different from the others. @Rikkola, WDYT?

Rikkola:
@manstis @kkufova No it works from the inherited class that is `KieEditor`, but if any location just does the `super.onClose()` call in `onClose()` then that can be removed.

manstis:
@Rikkola OK, no problem.

There are _lots_ of [places](https://github.com/kiegroup/drools-wb/blob/893258715c5883843e34fe7419386f5fd6302e37/drools-wb-screens/drools-wb-drl-text-editor/drools-wb-drl-text-editor-client/src/main/java/org/drools/workbench/screens/drltext/client/editor/DRLEditorPresenter.java#L205) where the call to `super` is used; but they are benign and don't cause any problems.. so I won't demand the PRs be updated (to change code that already existed)!!

Rikkola:
@manstis if the method overwrites onClose() then it needs to call super.onClose(). What I mean is if the super.onClose() is the only thing in the method, then that method could be removed.

PR #1318, opened 2019-03-15
JBPM-8116: It is not possible to see SLA violations in business centr…

nmirasch:
…al UI
- Changed task queries to retrieve slacompliance and sladuedate
- Updated dataset to include the new info
- added the new columns to taks lists
- added the filter by sla compliance to task basic filters
- Updated task details to show sla compliance
- Added new service to tasks services to retrieve the sla compliance and sla duedate to refresh tasks details
- Moved SLAStatusConverter and sla status constants to jbpm-commons to be reused from process and taks
- Simplified SLAComplianceCell to show the common constants instead to use passed descriptions, regarding the constans are the same for process and tasks


nmirasch:
@cristianonicolai @jstastny-cz first approach of adding SLA to task. I tried to unify as much as possible between process and task, letting just the name of the field in each module (even right now they are the same :S ).
I would ask you to take a closer look on the queries changes done to include the sla info: a "left join NodeInstanceLog nil on nil.workItemId = t.workItemId" was added regarding the related info is in NodeInstanceLog table. I have tried that on my local environment at is working fine, but if you have a better approach to include that is more than welcome.
Please take a look and let me know what do you think. Thanks!

nmirasch:

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskdetails/TaskDetailsPresenter.java
```diff
@@ -92,12 +92,21 @@ protected void setTaskDetails(String status,
...
-                                  String processId) {
+                                  String processId,
+                                  Long workItemId,
+                                  Integer slaCompliance) {
         view.setTaskDescription(description);
         view.setSelectedDate(expirationTime);
         view.setUser(actualOwner);
         view.setTaskStatus(status);
         view.setTaskPriority(priority);
+        //if(workItemId!=null) {
```
cristianonicolai: please remove comment out code.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskdetails/TaskDetailsPresenter.java
```diff
@@ -141,11 +152,11 @@ public void onTaskRefreshedEvent(@Observes final TaskRefreshedEvent event) {
...
                                            task.getProcessInstanceId(),
-                                           task.getProcessId());
+                                           task.getProcessId(),
+                                           task.getWorkItemId(),
+                                           task.getSlaCompliance());
                         }
-                    }).getTask(getServerTemplateId(),
-                               getContainerId(),
-                               getTaskId());
+                    }).getTaskWithSLA(getServerTemplateId(), getContainerId(), getTaskId(), view.getWorkItemId());
```
cristianonicolai: workitemid should be kept in the presenter to hold logic, not in the view.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/RemoteTaskServiceImpl.java
```diff
@@ -77,6 +80,33 @@ public TaskSummary getTask(String serverTemplateId,
         }
     }
 
+    @Override
+    public TaskSummary getTaskWithSLA(String serverTemplateId, String containerId, Long taskId, Long workItemId) {
```
cristianonicolai: getTask already returns the _workItemId_ it doesnt need to be received as parameter in here.

cristianonicolai: I wonder if we should enhance _getTask_ instead, to make the data consistent in all calls. wdyt?

nmirasch: @cristianonicolai I added this new getTaskWithSLA because the getTask didn't fill the workItemId. The idea was to reuse the info (workitem) retrieved by de dataset query (maintained wrongly at view) to be able to refresh the task data and  sla info retrieving the nodeinstance related. Maybe we can simplify this, updating the service find bytaskId to fill this info. does this makes sense?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/list/ProcessInstanceListBasicFiltersPresenter.java
```diff
@@ -122,7 +124,7 @@ public void loadFilters() {
                                                                            values);
                                       addSearchFilterList(f,
                                                           columnFilter);
-        });
+                                  });
```
jstastny-cz: I am not sure this is an improvement, though not harmful.

jbpm-wb-human-tasks/jbpm-wb-human-tasks-api/src/main/java/org/jbpm/workbench/ht/model/TaskDataSetConstants.java
```diff
@@ -43,6 +43,8 @@
     public static final String COLUMN_PROCESS_INSTANCE_CORRELATION_KEY = "correlationKey";
     public static final String COLUMN_PROCESS_INSTANCE_DESCRIPTION = "processInstanceDescription";
     public static final String COLUMN_EXCLUDED_OWNER = "entity_id";
+    public static final String COLUMN_SLA_DUE_DATE = "sla_due_date";
```
jstastny-cz: Is it us, who define the column names? Please keep it consistent at least in the newly added columns. (Majority is camelCase so I'd stick to that) ...

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/RemoteTaskServiceImpl.java
cristianonicolai: @nmirasch I think so, from what I can see, _workItemId_ is available in the _getTask_ method just not set in the _TaskSummary_ object. I guess we just need to ensure that _getTask_ still work for when the container is not available, with this new info about the node instance.

nmirasch:
Depends on:
https://github.com/kiegroup/jbpm/pull/1464
https://github.com/kiegroup/droolsjbpm-integration/pull/1781

jbpm-wb-human-tasks/jbpm-wb-human-tasks-api/src/main/java/org/jbpm/workbench/ht/model/TaskDataSetConstants.java
nmirasch: @jstastny-cz  this constants have to be consistent with the database column names to let the data set find the data at query. Then the answer is the value is defined by the database.

nmirasch:
@cristianonicolai @jstastny-cz I have changed the service that findTaskById to be able to retrieve the task sla compliance data without the need to maintain the workitem in any place, just finding the data at TaskInstance retrieved by the service. Please could you take a look? thanks!

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/util/SLAComplianceCell.java
```diff
@@ -21,63 +21,59 @@
...
+import org.jbpm.workbench.common.client.resources.i18n.Constants;
 import org.kie.api.runtime.process.ProcessInstance;
 
 import com.google.gwt.cell.client.AbstractCell;
 import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
 
 public class SLAComplianceCell extends AbstractCell<Integer> {
-    
+
+    private final Constants constants = Constants.INSTANCE;
```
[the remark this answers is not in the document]
nmirasch: I think there's no need to use that constants field  if there's no SLAComplianceCell instance, so making that final would be enough. But there's no problem to change it if you thing it worth.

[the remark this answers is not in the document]
jstastny-cz: The cell is being instantiated once per the table? If that's so, I understand the solution.

cristianonicolai:

nmirasch:

nmirasch:

nmirasch:
rebased

cristianonicolai:

PR #1323, opened 2019-03-18
JBPM-8306: Process diagram images are hardcoded to large sizes

nmirasch:
work with svg viewBox attribute instead with and height

### DIFF
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.java
@@ -81,8 +81,12 @@ public class ProcessDiagramWidgetViewImpl extends Composite implements ProcessDi
 
         final D3 svg = d3.select("#processDiagramDiv svg");
 
-        double svgWidth = Double.parseDouble(svg.attr("width").toString());
-        double svgHeight = Double.parseDouble(svg.attr("height").toString());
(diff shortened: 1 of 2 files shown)


nmirasch:
depends on https://github.com/kiegroup/droolsjbpm-integration/pull/1767

nmirasch:
@cristianonicolai @jstastny-cz Updated the way to retrieve the svg  width and height from viewbox attribute. Doing it we allow to see the process diagram in the same scale that was created. Let me know what you think. Thanks!

cristianonicolai:

jstastny-cz:
@nmirasch why exactly do we need to set the width and height attributes while we have the viewBox?

nmirasch:
@jstastny-cz it is done to keep working the zooming on the diagram that use this values to move the svg to the top- left corner. I thought this was the faster way to avoid redo the zooming regarding the new svg attributes

cristianonicolai:
@jstastny-cz are you happy with these changes?

jstastny-cz:
According to https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox
> The value of the viewBox attribute is a list of four numbers min-x, min-y, width and height, separated by whitespace and/or a comma,

The attribute value with commas should be probably considered valid and should be handled accordingly. @nmirasch what do you think?

nmirasch:
@jstastny-cz regarding this viewbox attribute is added at rest service and processed here, I don't see the possible case that the viewbox comes separated by comma
@cristianonicolai do you think that is useful to consider that case here?

cristianonicolai:
@nmirasch agree with you, we're consuming the image only from the specific endpoint that we manage, knowing specifically how it would be formatted/set values.

nmirasch:
@cristianonicolai Ok, Thanks,  and thank you @jstastny-cz for bring this up

nmirasch:
rebased

PR #1331, opened 2019-04-12
JBPM-8420 - Only executed lookups when related data set changed.

cristianonicolai:


cristianonicolai:

PR #1342, opened 2019-04-30
JBPM-8456: When the default filter is removed, an error is shown and …

nmirasch:
…the screen is not more accessible
Set the first saved filter when the default filter is removed.
Update the saved filters list with the new default filter in case the default is removed or the filters are restored


nmirasch:
@jstastny-cz could you take a look?

nmirasch:
@cristianonicolai could you take a look?

PR #1363, opened 2019-06-11
RHPAM-2120 Change the default strategy from Singleton to PerProcessIn…

bxf12315:
…stance
@cristianonicolai @nmirasch 
For the issue, I find two way can fix it, (another PR in [here](https://github.com/kiegroup/droolsjbpm-knowledge/pull/388) ), I not sure which is better, just for my side this PR is better, because of it's code change less than another PR.


cristianonicolai:
@bxf12315 I believe this might not be needed, seems that you covered it in https://github.com/kiegroup/droolsjbpm-knowledge/pull/388

PR #1373, opened 2019-07-18
Fix missing uberfire preferences dependency

cristianonicolai:


nmirasch:
@cristianonicolai this PR fixes the compilation and allow to build master, but when I try to start the case management showcase it raise the following message :
Invalid bus content! Expecting content type 'application/json', but got 'text/html;charset=UTF-8!.
is it working in your environment?

cristianonicolai:

cristianonicolai:
@nmirasch I could start it using gwt dev mode just fine, have you checked the final zip?

PR #1385, opened 2019-09-02
JBPM-8763: Tries to load KeyCloak if a flag is true

jesuino:
Business Central, as controller, tries to load keycloak classes. If Keycloak classes are not available, then it loads the default credentials provider to communicate with Kie Server registered servers.

After `AF-2086`, keycloak JARs are available by default in Business Central, which means that BC will tries to use Keycloak, even if user did not setup keycloak, leading to errors and a complete failure of BC components that requires Kie Server access (dashboards, servers management...)

This PR aims to create a new system property which allow users to disable Keycloak classes loading. Only if the system property `org.jbpm.workbench.kie_server.keycloak` is set to true BC will try to load keycloak classes.


jesuino:
@cristianonicolai would you please review this?  THanks!

jesuino:

PR #1387, opened 2019-09-03
[7.26.x] JBPM-8763 - Check UserPrincipal class type for using Keycloak

cristianonicolai:


cristianonicolai:

cristianonicolai:
@jesuino @tomasdavidorg executed local tests using Keycloak v7 server in docker image, plus setting up both _keycloak-wildfly-adapter-dist-6.0.1_ and _keycloak-wildfly-adapter-dist-7.0.0_, in both cases it seems to work fine. Here is the setup I used on Wildfly: https://gist.github.com/cristianonicolai/0a007e9eb9a20e06575de0dec2ee70f1
Could you please also verify this with the proper/supported versions of EAP and SSO? Thanks

cristianonicolai:
@mswiderski yeah, jars have been included now as part of the work in https://issues.jboss.org/browse/AF-2086, that's why the authentication was broken.

cristianonicolai:

cristianonicolai:

jesuino:
I tested it with Keycloak 4.8.3 JARs because that's the version that will be used on 7.26.0.Final. KIE Server is *not* running on the same EAP as BC.

The tests I made:

* Create users and roles on BC reflected on RH SSO;
* Same for user and roles created on RH SSO, they are reflected on BC;
* Installed Evaluation sample, started a process for an user Created on Keyloack (user "employee" with roles: admin, kie-server,rest-all)
   - I noticed that if I start the process with other use than "employee" then the user employee does not have access to the process' tasks - but everything works if we start the process using user employee;

No changes were made directly in the BC WAR (no *.properties or web.xml was edited).

@mswiderski  @cristianonicolai  you can read more about why the JARs were added to BC in this discussion -> https://issues.jboss.org/browse/KIECLOUD-246?focusedCommentId=13743626&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13743626

krisv:
@cristianonicolai given the ongoing verification discussion and that the current solution (with system property) seems to work well (and necessary changes were already done and verified in the cloud templates too), my recommendation would be to NOT cherry-pick this to 7.26.x anymore (to minimize risk due to further changes there)

cristianonicolai:
@kris, np, will close this PR and work only on the master branch change.

PR #1389, opened 2019-09-12
JBPM-8774: Task List Bulk Reassign error when a task has inappropriat…

nmirasch:
…e status
Synchronize the checks to allow or deny delegate or forward operation on tasks, following now the same criteria as the jbpm engine

### DIFF
--- a/jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/TaskAssignmentSummaryMapper.java
+++ b/jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/TaskAssignmentSummaryMapper.java
@@ -17,10 +17,10 @@
 package org.jbpm.workbench.ht.backend.server;
 
 import java.util.Collections;
-import java.util.List;
 import java.util.function.BiFunction;
 
 import org.jbpm.workbench.ht.model.TaskAssignmentSummary;
+import org.jbpm.workbench.ht.util.TaskStatus;
 import org.kie.internal.identity.IdentityProvider;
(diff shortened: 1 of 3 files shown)


mswiderski:
Tasks are not really created by users but by system even if given user was starting process that led to creating a task. So I would say we should not use createdBy as any privileged role

cristianonicolai:

jbpm-wb-human-tasks/jbpm-wb-human-tasks-backend/src/main/java/org/jbpm/workbench/ht/backend/server/TaskAssignmentSummaryMapper.java
```diff
@@ -51,62 +51,61 @@ public TaskAssignmentSummary apply(final TaskInstance task,
 
     protected Boolean isForwardAllowed(final TaskInstance task,
                                        final IdentityProvider identityProvider) {
-        if ("Completed".equals(task.getStatus())) {
-            return false;
+        if (TaskStatus.TASK_STATUS_READY.equals(task.getStatus())) {
```
jstastny-cz: the implementation of isForwardAllowed and  isDelegationAllowed seem identical. If the criteria are supposed to be the same for both, we should implement them in a single method, so that they don't diverge in time.

Consider introducing private method that would these two call (to keep the distinction between forwarding and delegating a task)

nmirasch: @jstastny-cz, this duplication was related with maintenance purposes as they are reflecting different operations checks, but regarding currently they are checking the same conditions I unified them in your proposed private method. Could you take a look? Thanks!

PR #1390, opened 2019-09-13
JBPM-8775: Bulk task reassignment and bulk PI signaling reset checkboxes before finishing... 

nmirasch:
…the operation
Added BulkActionPerformedEvent that is fired when a bulk operation that requires a data introduction popup is performed, to allow unselect the checkboxes after the operation is done

### DIFF
--- a/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractListPresenter.java
+++ b/jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractListPresenter.java
@@ -17,10 +17,13 @@ package org.jbpm.workbench.common.client.list;
 
 import java.util.List;
 
+import javax.enterprise.event.Observes;
+
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.view.client.AsyncDataProvider;
 import com.google.gwt.view.client.HasData;
 import com.google.gwt.view.client.Range;
+import org.jbpm.workbench.common.client.list.event.DeselectAllItemsEvent;
 import org.jbpm.workbench.common.client.menu.RefreshMenuBuilder;
 import org.jbpm.workbench.common.client.resources.i18n.Constants;
 import org.jbpm.workbench.common.model.QueryFilter;
@@ -77,7 +80,7 @@ public abstract class AbstractListPresenter<T> implements RefreshMenuBuilder.Sup
     }
 
     public void addDataDisplay(final HasData<T> display) {
-        if(dataProvider.getDataDisplays().size() == 1){
+        if (dataProvider.getDataDisplays().size() == 1) {
             dataProvider.removeDataDisplay(dataProvider.getDataDisplays().iterator().next());
         }
         dataProvider.addDataDisplay(display);
@@ -100,6 +103,16 @@ public abstract class AbstractListPresenter<T> implements RefreshMenuBuilder.Sup
         if (getListView().getListGrid() != null) {
             getListView().getListGrid().setVisibleRangeAndClearData(getListView().getListGrid().getVisibleRange(),
                                                                     true);
+        }
+    }
+
+    public void onDeselectAllItemsEvent(@Observes DeselectAllItemsEvent event) {
+        deselectAllItems();
+    }
+
+    public void deselectAllItems() {
+        if (getListView().getListGrid() != null) {
+            getListView().getListGrid().deselectAllItems();
         }
     }
 }
--- a/jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/popup/TasksReassignmentPresenter.java
+++ b/jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/popup/TasksReassignmentPresenter.java
@@ -26,6 +26,7 @@ import javax.inject.Inject;
 import com.google.gwt.core.client.GWT;
 import org.jboss.errai.bus.client.api.messaging.Message;
 import org.jboss.errai.common.client.api.Caller;
+import org.jbpm.workbench.common.client.list.event.DeselectAllItemsEvent;
 import org.jbpm.workbench.ht.client.resources.i18n.Constants;
 import org.jbpm.workbench.ht.model.TaskAssignmentSummary;
 import org.jbpm.workbench.ht.model.TaskKey;
@@ -55,6 +56,8 @@ public class TasksReassignmentPresenter {
 
     @Inject
     private PlaceManager placeManager;
+
+    private Event<DeselectAllItemsEvent> deselectAllItemsEvent;
 
     private Event<TaskRefreshedEvent> taskRefreshed;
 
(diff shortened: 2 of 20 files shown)


nmirasch:
@cristianonicolai  @jstastny-cz  could you take a look? Thanks!

cristianonicolai:

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/popup/TasksReassignmentPresenter.java
```diff
@@ -118,6 +122,7 @@ public void reassignTasksToUser(final String userId) {
                                      return;
                                  }
                              });
+                             bulkActionPerformedEvent.fire(new BulkActionPerformedEvent());
```
jstastny-cz: where is the event being fired also for other than the reassign operations?

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/signal/ProcessInstanceSignalPresenter.java
```diff
@@ -94,6 +99,7 @@ public void signalProcessInstances(List<Long> processInstanceIds) {
         }
         processService.call((Void v) -> {
                                 processInstancesUpdatedEvent.fire(new ProcessInstancesUpdateEvent(0L));
+            bulkActionPerformedEvent.fire(new BulkActionPerformedEvent());
```
nmirasch: @jstastny-cz here!

jstastny-cz: ... I didn't consider this affecting task bulk actions due to the class name ... so the ProcessInstanceSignalPresenter is also taking part in task bulk actions?

nmirasch: Nop, It's performing a multiple signaling on process instances that requires introduction of data: name and data. It had the same issue with the checkboxes you found at task. That's the reason to create this generic 'BulkActionPerformedEvent' to be able to deselect when the bulk action is performed

nmirasch: @jstastny-cz maybe makes sense rename the JBPM jira and the commit to not to be linked with task bulk operation, and just talk about bulk actions and checkboxes?

jstastny-cz: that's what I assumed, but then I could not see a place where for tasks the event is being fired other than during reassign operation.

nmirasch: yes, this event is only used when the operation requires  introduction of data (signal data in case of PI and user id in case of task reassignment). The other, have bulk action confirmation that includes the unselection once it's confirmed

jstastny-cz: Wouldn't be unified solution (firing the event everywhere) better? Having two different behaviors for that same sort of actions is puzzling (example: me).

nmirasch:
@cristianonicolai @jstastny-cz  changed the approach to avoid to use a new event when a bulk action is performed, adding instead a new refreshGrid that allow specify if the unselect operation have to be done on refreshing. Another change introduced here is fire TaskRefreshedEvent only one time. Please, could you take a look? Thanks again!

jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/popup/TasksReassignmentPresenter.java
```diff
@@ -112,12 +112,7 @@ public void reassignTasksToUser(final String userId) {
                                      displayUnsuccessfulNotification(constants.ReassignmentNotAllowedOn(String.valueOf(taskAssignmentSummary.getTaskId()), taskAssignmentSummary.getTaskName(), userId));
                                  }
                              });
-                             reassignments.forEach(taskAssignmentSummary -> {
-                                 if (taskAssignmentSummary.isDelegationAllowed()) {
-                                     taskRefreshed.fire(new TaskRefreshedEvent());
-                                     return;
-                                 }
-                             });
+                             taskRefreshed.fire(new TaskRefreshedEvent());
```
cristianonicolai: @nmirasch if you fire this event, dont you need to set the taskId, etc?

cristianonicolai:
@nmirasch @jstastny-cz correct the issue is that we have a popup that is called in its own presenter and view. Perhaps we can unify this by creating a specific event (DeselectAllItemsEvent?) that can be fired in different places but handled in a single place. 
For this issue in particular, I would suggest that we only fire that event in here: https://github.com/kiegroup/jbpm-wb/blob/master/jbpm-wb-human-tasks/jbpm-wb-human-tasks-client/src/main/java/org/jbpm/workbench/ht/client/editors/taskslist/popup/TasksReassignmentPresenter.java#L121

nmirasch:
@cristianonicolai @jstastny-cz  Tried to follow the approach of firing That new DeselectAllItemsEvent. I changed the jira description to be able to fire that from  Process instance bulk signal too that had the same issue. Let me know what you think.

cristianonicolai:
@nmirasch new approach looks good to me

cristianonicolai:

jstastny-cz:
@nmirasch @cristianonicolai 
to summarize: in the current state, we have deselect action implemented
* at presenter level for task reassignment action (using the event)
* at view level for other actions (using the callback in getSuspendCommand and similar)

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractListPresenter.java
```diff
@@ -102,4 +105,11 @@ public void refreshGrid() {
                                                                     true);
         }
     }
+
+    public void onDeselectAllItemsEvent(@Observes DeselectAllItemsEvent event) {
+        if (getListView().getListGrid() != null) {
+            getListView().getListGrid().deselectAllItems();
+        }
+        refreshGrid();
```
cristianonicolai: just not sure if we always want to refresh the grid when deselectall is executed.

cristianonicolai:
@jstastny-cz correct, perhaps with the new event approach that all could be consolidated.

nmirasch:
@jstastny-cz @cristianonicolai Moved the unselectCheckboxes calls to presenter across the whole bulk actions. At bulk reassignment, restored the TaskRefresedEvent firing (filled now with the first task reassigment data) to refresh the grid to avoid link the unselect and refreshGrid. The  DeselectAllItemsEvent is used just where is required (bulk actions that requires popup to introduce data) the other actions just call to presenter deselectAllItems() directly.

nmirasch:

nmirasch:

nmirasch:

jstastny-cz:
@nmirasch please rebase the PR, the failures might be solved by that.

nmirasch:

jstastny-cz:

nmirasch:

nmirasch:

PR #1404, opened 2019-11-12
JBPM-8894 Saved Filter rewritten when adding column with sort order

bxf12315:


nmirasch:
@bxf12315 maybe add something like verify(listTable,never()).saveGridPreferences(); at AbstractMultiGridViewTest.testAddColumnSortHandler() test would cover the change

bxf12315:
> @bxf12315 maybe add something like verify(listTable,never()).saveGridPreferences(); at AbstractMultiGridViewTest.testAddColumnSortHandler() test would cover the change

yes you are right

bxf12315:
@nmirasch done

jbpm-wb-common/jbpm-wb-common-client/src/main/java/org/jbpm/workbench/common/client/list/AbstractMultiGridView.java
```diff
@@ -214,7 +214,6 @@ protected void addColumnSortHandler(ExtendedPagedTable listTable) {
                     gridPreferencesStore.setGridSortedColumnPreference(gridSortedColumnPreference);
                 }
             }
-            listTable.saveGridPreferences();
```
cristianonicolai: @bxf12315 sounds like we would persist the changes in the sorting at all with this change. I believe that might still be needed, @nmirasch wdyt?

bxf12315: @cristianonicolai yes, sort column need persist, but it saves it in to current GridPreferencesStore, if create a new filter, it will save again. So If we delete it, do not impact save sort column into new filter.

cristianonicolai:

PR #1411, opened 2020-01-28
JBPM-8953 Load saved filter after Save filter launches an error

bxf12315:


nmirasch:

cristianonicolai:
@jstastny-cz the logic on the iterator looks fine to me, hasNext will include the first item, unless I missing something here.

jstastny-cz:
Correct, excuse my morning dementia.

jstastny-cz:
Ready to be merged, even though full downstream reports failure, the business central war is available and the fix has been verified.

PR #1420, opened 2020-02-21
JBPM-9036 Disable 'Double-clicking' on  ProcessDiagram

bxf12315:


bxf12315:
> I just was thinking if the double click on the process definition need to have the same behavior instead the zoom. But this looks good to me and works fine

What's the default behavior on process definition? 
I merged the code into PR:https://github.com/kiegroup/jbpm-wb/pull/1424

PR #1433, opened 2020-05-11
Use containsOnly insead of containsExactly in ProcessInstanceDiagramPresenterTest.java

cpugputpu:
The test in `org.jbpm.workbench.pr.client.editors.instance.diagram.ProcessInstanceDiagramPresenterTest#testOnProcessInstanceSelectionEvent` can fail due to a different iteration order of `Map`. The failure is presented as follows:
`java.lang.AssertionError:`
`Actual and expected have the same elements but not in the same order, at index 0 actual element was:`
 `<MapEntry[key="_2", value=1L]>`
`whereas expected element was:`
  `<_1=1>`

The fix is to use `containsOnly` instead of `containsExactly` so that the iteration order of `Map` does not make the assertion fail. In this way, the test will be more robust, and the failure above will be removed.


mareknovotny:
ok to test

PR #1443, opened 2020-07-03
JBPM-9214 The zoom does not work when start a new process from Proces…

bxf12315:
…s Definition

### DIFF
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.java
@@ -23,6 +23,7 @@ import javax.inject.Inject;
 import javax.inject.Named;
 
 import com.google.gwt.event.dom.client.ClickEvent;
+import com.google.gwt.user.client.DOM;
 import com.google.gwt.user.client.Timer;
 import com.google.gwt.user.client.ui.Composite;
 import com.google.gwt.user.client.ui.RequiresResize;
@@ -78,7 +79,7 @@ public class ProcessDiagramWidgetViewImpl extends Composite implements ProcessDi
 
     private boolean isExpand = false;
 
-    boolean isDoubleClick;
+    protected boolean isDoubleClick;
 
     public void setD3Component(D3 d3) {
         this.d3 = d3;
@@ -89,6 +90,8 @@ public class ProcessDiagramWidgetViewImpl extends Composite implements ProcessDi
         d3 = D3.Builder.get();
         expandAnchor.setIcon(IconType.EXPAND);
         this.isDoubleClick = false;
+
+        processDiagramDiv.id = DOM.createUniqueId();
     }
 
     @Override
@@ -105,7 +108,7 @@ public class ProcessDiagramWidgetViewImpl extends Composite implements ProcessDi
     public void displayImage(final String svgContent) {
         processDiagramDiv.innerHTML = svgContent;
 
-        final D3 svg = d3.select("#processDiagramDiv svg");
+        final D3 svg = d3.select("#" + processDiagramDiv.id + " svg");
 
         String[] viewBoxValues = svg.attr("viewBox").toString().split(" ");
 
@@ -269,5 +272,9 @@ public class ProcessDiagramWidgetViewImpl extends Composite implements ProcessDi
                 nodeSelectionCallback.callback((String) node.attr("bpmn2nodeid"));
             }
         }
+    }
+
+    public String getProcessDiagramDivId() {
+        return processDiagramDiv.id;
     }
 }
--- a/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/diagram/ProcessInstanceDiagramViewImpl.java
+++ b/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/instance/diagram/ProcessInstanceDiagramViewImpl.java
@@ -282,7 +282,8 @@ public class ProcessInstanceDiagramViewImpl extends Composite implements Process
 
     protected void showHideBadges(){
         final D3 d3 = D3.Builder.get();
-        final D3 nodes = d3.selectAll("#processDiagramDiv svg [jbpm-node-badge]");
+        final String processDiagramDivId = diagram.getProcessDiagramDivId();
+        final D3 nodes = d3.selectAll("#" + processDiagramDivId + " svg [jbpm-node-badge]");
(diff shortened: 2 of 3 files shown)


cristianonicolai:
@bxf12315 have you checked if the issue could be caused by the wrong element being selected on https://github.com/kiegroup/jbpm-wb/blob/master/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.java#L108 
Iḿ wonering that it could be two dics with same id ( processDiagramDiv ) at the point. Maybe you just need to have different ids?

bxf12315:
> @bxf12315 have you checked if the issue could be caused by the wrong element being selected on [/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.java@`master`#L108](https://github.com/kiegroup/jbpm-wb/blob/master/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.java?rgh-link-date=2020-07-07T13%3A25%3A25Z#L108)
> Iḿ wonering that it could be two dics with same id ( processDiagramDiv ) at the point. Maybe you just need to have different ids?

Yes you are right, the reason is two same id div in on page, Proces Definiton detail have ProcessDiagramPresenter, Process instance detail also have the component. if we use different id, we need create new ProcessDefinitionDiragram for it and it total same to ProcessDiagramPresenter. I think ProcessDiagramPresenter is reuseable for process defintions and process instance .

cristianonicolai:
@bxf12315 just create a new unique id every time `ProcessDiagramWidgetViewImpl` is created? then use that as selector.

bxf12315:
xswa

> @bxf12315 just create a new unique id every time `ProcessDiagramWidgetViewImpl` is created? then use that as selector.

we inject it in here https://github.com/kiegroup/jbpm-wb/blob/7e60a5274b01b790d2ef313961b061ca89672de1/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.java#L48.
https://github.com/kiegroup/jbpm-wb/blob/523f3596e47896f24ec8e822b04d16a8c1412d33/jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramWidgetViewImpl.html#L9

I'm not sure how to refactor them. The PR just want to keep mini change.

bxf12315:
@cristianonicolai Follow your comment, I use change id way and it's work.
But in the solution, one page has two same functions components, we just change one div id of them, there also have many same id's div(same functions). I'm not sure what problems this will cause， but it's work and does not have an issue.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramPresenter.java
```diff
@@ -46,6 +46,7 @@ public void setProcessImageService(final Caller<ProcessImageService> processImag
     public void onProcessSelectionEvent(@Observes final ProcessDefSelectionEvent event) {
         view.showBusyIndicator(constants.Loading());
         view.disableExpandAnchor();
+        view.setProcessDiagramDivId("ProcessDefDiagram");
```
cristianonicolai, line 49: the presenter should not be responsible for defining an id in the view. Simply generated a ramdom one on the view side?
`DOM.createUniqueId();`

bxf12315: if use the way, one page two "<div class="col-md-12" data-field="diagramContainerDiv" data-i18n-prefix="ProcessDiagramWidgetViewImpl.">", is it reasonable?

bxf12315: updated, Please review again.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/definition/list/ProcessDefinitionListPresenter.java
```diff
@@ -57,12 +57,14 @@
...
 import org.uberfire.workbench.model.menu.Menus;
 
+import static org.jbpm.workbench.common.client.PerspectiveIds.PROCESS_DEFINITION_DETAILS_SCREEN;
 import static org.jbpm.workbench.common.client.PerspectiveIds.SEARCH_PARAMETER_PROCESS_DEFINITION_ID;
 import static org.jbpm.workbench.common.client.util.DataSetUtils.getColumnStringValue;
 import static org.jbpm.workbench.pr.model.ProcessDefinitionDataSetConstants.*;
 import static org.kie.workbench.common.services.shared.resources.PerspectiveIds.PROCESS_INSTANCES;
 @Dependent
 @WorkbenchScreen(identifier = PerspectiveIds.PROCESS_DEFINITION_LIST_SCREEN)
+@SuppressWarnings("PMD.TooManyStaticImports")
```
cristianonicolai, line 67: I dont think any of these changes are needed to fix the issue. could you revert it please.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/diagram/ProcessDiagramPresenter.java
cristianonicolai: sorry but I didnt get whats the issue here.

jbpm-wb-process-runtime/jbpm-wb-process-runtime-client/src/main/java/org/jbpm/workbench/pr/client/editors/definition/list/ProcessDefinitionListPresenter.java
bxf12315: done

jstastny-cz:
Why haven't been me or @prashuts asked for review here? This has broken our tests.

bxf12315:
@jstastny-cz sorry I forget it, which test broken?

prashuts:
@bxf12315  - Tests related to ProcessInstanceDiagram where script mainly involves Process Diagram panels. Scripts failed due to change in Locator attriubute id value (older - processDiagramDiv)

bxf12315:
Yes According to the @cristianonicolai's comments, use createUniqueId to generate a id of the div.
So Can I do some help for fix the issue?

PR #1448, opened 2020-08-05
JBPM-9271 Change Service Tasks labels to Custom Tasks

bxf12315:


jstastny-cz:
jenkins execute full downstream build please

jbpm-wb-integration/jbpm-wb-integration-client/src/main/resources/org/jbpm/workbench/wi/client/i18n/Constants.properties
```diff
@@ -68,39 +68,39 @@ DDValidationNotValidResolver=Not valid resolver selected for {0}
...
+MavenInstallHelp=Instructs if enabled custom tasks should be installed into Maven repository
+InstallPomDeps=Install custom task dependencies into project
+InstallPomDepsHelp=Instructs that custom task dependencies are added as project dependencies upon installation
+UseVersionRange=Use version range when installing custom task into a project
+UseVersionRangeHelp=Instructs that a version range will be used when installing custom task in projects
 ServiceTaskList=Custom Tasks Administration
-NoServiceTasksFound=No service tasks found
+NoServiceTasksFound=No custom tasks found
 BaseSettings=Settings
-ServiceTasks=Service Tasks
```
romartin, line 81: Just a really minor comment, not just in this line, but in general we're changing only the literal "Service" to "Custom" in the "values" for the properties file, at some point would also make sense to change the "keys", so once just coding on top of this, the developer will not be confused about service vs custom. 
Anyway, from the user perspective, it works, so we can go ahead.

prashuts:
jenkins do fdb

PR #1473, opened 2020-12-09
[JBPM-9099] Long running query at kie-server startup

fjtirado:
Modifying query as per JIRA description.

**JIRA**: 

[https://issues.redhat.com/browse/JBPM-9099](https://issues.redhat.com/browse/JBPM-9099)

** referenced prs**
https://github.com/kiegroup/kie-soup/pull/168

### DIFF
--- a/jbpm-wb-kie-server/jbpm-wb-kie-server-backend/src/main/resources/default-query-definitions.json
+++ b/jbpm-wb-kie-server/jbpm-wb-kie-server-backend/src/main/resources/default-query-definitions.json
@@ -7,8 +7,8 @@
   },
   {
     "query-name": "jbpmProcessInstancesWithVariables",
-    "query-source": "${org.kie.server.persistence.ds}",
-    "query-expression": "select vil.processInstanceId, vil.processId, vil.id, vil.variableId, vil.value from VariableInstanceLog vil where vil.id in (select MAX(v.id) from VariableInstanceLog v group by v.variableId, v.processInstanceId)",
+	"query-source": "${org.kie.server.persistence.ds}",    
+    "query-expression": "select vil.processInstanceId, vil.processId, vil.id, vil.variableId, vil.value from VariableInstanceLog vil left join VariableInstanceLog vil2 on vil.processInstanceId = vil2.processInstanceId and vil.variableId = vil2.variableId and vil.id < vil2.id where vil2.id is null",
     "query-target": "CUSTOM"
   },
   {


exabrial:
Thanks for submitting this. 

I have two things for you, first, we noticed these indexes helped some of the other queries significantly:

```
CREATE INDEX `idx. ProcessInstanceInfo.lastModificationDate` ON `drools`.`ProcessInstanceInfo` (`lastModificationDate`) USING BTREE;
CREATE INDEX `idx.NodeInstanceLog.workItemId` ON `drools`.`NodeInstanceLog` (`workItemId`) USING BTREE;
```

Next, is there support to paginate these queries? Our variable instance log table is about 54 million rows. There's not really a reason for the UI to pull of that data back at once. We rewrote a lot of these queries internally to ad limits of about 10k rows returned (this solution has a lot of issues, mainly we can't look at more than a days worth of history in the UI. I think we'd be willing to put up the hours to implement this if we could get some direction from the project maintainers so the PR doesn't get rejected :)

elguardian:
@exabrial Thanks for your points. 
Keep in mind that during bootstrap this queries are limited by a pagination statement (LIMIT 1). These queries should get really fust now that depend on index.
Do you mind to put your findings in the jira. We can verify them. (though the use of BTREE indexes are not supported by every database.

I am going to merge this one as soon as CI gets green.

elguardian:
Jenkins retest this

fjtirado:
@exabrial Please take a look to https://github.com/kiegroup/kie-soup/pull/168, there we will prevent UI to retrieve any row at startup by using limit 0 
when this query is really executed by UI, it is automatically paginated by server code already.

elguardian:
Jenkins retest this

cristianonicolai:
@fjtirado @elguardian I believe this can be an improvement but it doesn't fix the original issue reported in the JIRA. Perhaps this should be handled in a different JIRA, as an improvement/enhancement.

cristianonicolai:
Jenkins run fdb

elguardian:
Hi @cristianonicolai both belongs to the same issue about performance. It is true that the effect during bootstrap si limited as this alone would not be able to fix the issue. If it concerns you we can split the jira in two but I don't think is worth it.

jbpm-wb-kie-server/jbpm-wb-kie-server-backend/src/main/resources/default-query-definitions.json
```diff
@@ -7,8 +7,8 @@
   },
   {
     "query-name": "jbpmProcessInstancesWithVariables",
-    "query-source": "${org.kie.server.persistence.ds}",
-    "query-expression": "select vil.processInstanceId, vil.processId, vil.id, vil.variableId, vil.value from VariableInstanceLog vil where vil.id in (select MAX(v.id) from VariableInstanceLog v group by v.variableId, v.processInstanceId)",
+	"query-source": "${org.kie.server.persistence.ds}",
```
cristianonicolai, line 10: perhaps this line change can be avoided.

cristianonicolai (approved):
tested locally seems to work fine, just a minor comment.

PR #1601, opened 2023-07-25
RHPAM-4664: Date filter is not working as expected

jesuino:
[JBPM-10167](https://issues.redhat.com/browse/JBPM-10167)

*Date filter is not working as expected*

In remote datasets the date filter is generate as an `COLUMN = {value selected by user}`. It does not work for dates. With this change we generate a BETWEEN operator for dates only and a suitable interval of dates so the correct values are generated.


ederign:
@domhanak who is the best person to test this?

tomasdavidorg (approved):
@jesuino Thanks for the steps. I was able to reproduce the issue with the older version. With the Business Central built from this PR, it works as expected.

mareknovotny:
jenkins do fdb

mareknovotny:
all good and green, merging

